PR #38, opened 2015-01-05
Job for deleting images from cloudinary

swatijadhav:


steveyken:
Thanks, @swati - good to see this. Just a couple of things I noticed from reading about ActiveJob and Sidekiq recently.

* Sidekiq author suggests not creating too many queues. Perhaps having a "high", "default", "low" queues would be best... I put airbrake in it's own queue, so it doesn't hold up other jobs if one part of the site is haemorrhaging (see https://github.com/mperham/sidekiq/wiki/Advanced-Options "I don't recommend having more than a handful of queues. Sidekiq is not designed to work well with dozens of queues.") The delete job would definitely be a low priority job.
* Also, best not to provide an object (`Image` in this case) as a parameter to the job... use an id instead and then look it up. See https://github.com/mperham/sidekiq/wiki/Best-Practices#1-make-your-jobs-input-small-and-simple and https://github.com/mperham/sidekiq/wiki/The-Basics "This means the arguments to your worker must be simple JSON datatypes (numbers, strings, boolean, array, hash)"

swatijadhav:
@steveyken Thanks for reviewing this PR and also for the suggestions with references. :)
Sure, I will add these changes now.

PR #40, opened 2015-01-10
Rails 4 2

shivanibhanwal:
Migrated app from Rails 4.1 to 4.2 and Ruby 2.1.5 to Ruby 2.2.
- Used ActiveJobs of rails 4.2 for the flowdock notifications.
- Fixed issue with the notification between Donor/Reviewer/Supervisor
-  Changes made in the token and auth code
- Postgres_ext_serializer gem changes to support the latest changes


app/jobs/flowdock_notification_job.rb
```diff
@@ -0,0 +1,8 @@
+class FlowdockNotificationJob  < ActiveJob::Base
+  queue_as :email
+
+  def perform(options)
+    @token = options
+    FlowdockNotification.otp(@token).deliver_later
+  end
+end
```
steveyken: @shivani I don't think we need to define a job for this. The ```deliver_later``` is creating the job for us in action_mailer. See http://api.rubyonrails.org/classes/ActionMailer/Base.html - it will do it for us so the following should just work

    FlowdockNotification.otp(@token).deliver_later

Right now, this is creating 2 jobs! One job to create another job that will send an email! :)

app/services/email_flowdock_service.rb
```diff
@@ -8,7 +8,7 @@ def initialize(user)
 
   def send_otp
     token = @user.most_recent_token
-    FlowdockNotification.otp(token).deliver # deliver_later once on Rails 4.2
+    FlowdockNotificationJob.perform_later(token)
   end
 
 end
```
steveyken: @shivani just revert this commit to say .deliver_later and remove the previous active_job

config/initializers/active_job.rb
```diff
@@ -1,9 +1,7 @@
 require "active_job"
```
steveyken: can remove this line now  too :)

shivanibhanwal: Right :) thanks ! missed it.

app/services/email_flowdock_service.rb
shivanibhanwal: It worked just tested in Production mode.
Thanks Steve. Removing the job

PR #47, opened 2015-01-27
mark offer as closed by reviewer or supervisor

swatijadhav:


app/controllers/api/v1/offers_controller.rb
```diff
@@ -98,6 +98,15 @@ def complete_review
...
 
+    api :PUT, '/v1/offers/1/close_offer', "Mark Offer as closed."
+    param :offer, Hash, required: true do
+      param :state_event, String, "State transition event ex: 'close'", required: true
+    end
+    def close_offer
+      @offer.update_attributes(review_offer_params)
+      render json: @offer, serializer: serializer
+    end
+
```
steveyken: @swatijadhav ```complete_review``` and ```close_offer``` are exactly the same method. Could you DRY the up a bit? Thanks

steveyken: Actually, I like the fact that there are two methods but perhaps do something like:

    def close_offer
      @offer.update_attributes(review_offer_params.merge(state_event: 'close'))
      render json: @offer...
   end

That way the state is explicitly mixed in and the APP and android app don't need to know the actual state_event name. I think really state_machine should be a hidden API where possible.

PR #55, opened 2015-03-16
Gcw 123 inconsistent tz formatting in js

swatijadhav:


steveyken:
Thanks @swatijadhav this looks good. It's useful to see what the sql query should be. Can I suggest making ```OFFER_TIME_ATTRIBUTES``` more generic, perhaps ```TIME_ATTRIBUTES``` and put the ```define_method``` on ```ApplicationSerializer```. That way we can easily define ```TIME_ATTRIBUTES``` for other classes, not just ```OfferSerializer```

Thanks.

PR #57, opened 2015-03-18
Gcw 157 when offer is submitted for review, send thanks you message to donor

swatijadhav:


config/initializers/constants.rb
```diff
@@ -1,3 +1,4 @@
 NEXT_AVAILABLE_DAYS_COUNT = 10
 START_DAYS_COUNT = 1
 CROSSROADS_TRUCK_COST = 600
+OFFER_THANK_YOU_MESSAGE = "Thank you for your offer, our team will start reviewing it soon. The reviewer will message you if they have questions about any of the items. We apologise if we cannot accept some items. Storage is limited and needs in the community change daily. Once the review is complete you will be able to arrange transport for any accepted items. If you have questions feel free to reply to this message."
```
steveyken: @swatijadhav this won't be able to be translated... can we use a view template? perhaps create something inside of ```views/messages/offer_thank_you.en.erb``` and then load the ERB template. You can use:

    file = "#{Rails.root}/app/views/offers/thank_you_message.#{I18n.locale}.erb"
    message = ERB.new(File.read(file))

swatijadhav: Just added translations here  https://github.com/crossroads/api.goodcity/commit/8a826fc95450b360b1f307a161e89718e9aff3a8

PR #63, opened 2015-04-07
Gcw 233 handle transport booking modification & cancellation edge cases

swatijadhav:


app/controllers/api/v1/deliveries_controller.rb
```diff
@@ -74,5 +74,10 @@ def delivery_params
         :contact_id, :schedule_id, :delivery_type, :gogovan_order_id)
     end
 
+    def delete_existing_delivery_record
+      delivery = Delivery.where(offer_id: params[:delivery][:offer_id]).last
+      delivery.try(:destroy)
+    end
```
steveyken: could do 

    if (delivery = Delivery.where(offer_id: params[:delivery][:offer_id]).last).present?
       delivery.destroy
    end

PR #67, opened 2015-04-20
Gcw 230 provide gogovan driver with a de

sunil-sharma:


app/controllers/api/v1/offers_controller.rb
```diff
@@ -125,6 +125,7 @@ def close_offer
     end
 
     def ggv_order_offer
+      @offer = Offer.find_by_ggv_uuid(params[:id])
       render json: @offer, serializer: serializer, exclude_messages: params[:exclude] == "messages"
     end
```
steveyken: @sunil-sharma  I think this method should move to the GoGoVanOrder controller as a 'driver_details' action. It should still return the serialized offer though.

app/models/offer.rb
```diff
@@ -119,6 +119,15 @@ def scheduled?
...
   end
 
+  def self.find_by_ggv_uuid(ggv_uuid)
+    begin
+      offer = GogovanOrder.find_by(ggv_uuid: ggv_uuid).delivery.offer
+    rescue Exception => e
+      raise ActiveRecord::RecordNotFound
+    end
+    with_eager_load.find(offer.id)
+  end
```
steveyken: @sunil-sharma  I think this method belongs on the GogovanOrder class as that is the primary class involved here. Also , as houndci specifies, it's not great to catch generic exceptions like this as we can't see what's really going wrong. It's fine for the API to return a 500 or a 404 - you can let it decide. So I would just remove the rescue entirely. If the lookup goes wrong, errbit will tell us and the user will get a 500 error - that's what we want.

sunil-sharma: Sure, I'll move it into the GogovanOrder class.
In order to handle scenario where driver bookmarked the link(or then offer has been cancelled) then we have to show 404 popup to driver, as @mattgow mentioned in the story GCW-230.

Thanks

app/controllers/api/v1/offers_controller.rb
sunil-sharma: Thanks, I will move this.

app/models/offer.rb
steveyken: Good point re 404. Also, we only want to show offers when the GGV order is in a valid collection state, either pending or active. Rather than catch the exception I think it's better to use ```try``` here as it is more specific. 

    offer = GogovanOrder.find_by_ggv_uuid(params[:ggv_uuid]).try(:delivery).try(:offer)
    raise ActiveRecord::RecordNotFound unless offer

PR #70, opened 2015-04-27
added received message for donor

sunil-sharma:


steveyken:
@sunil-sharma good work on this! nice 'n simple

PR #77, opened 2015-05-08
GCW-232: set ggv_poll_job wait time based on user's online status

swatijadhav:


steveyken:
Good job @swatijadhav !

PR #78, opened 2015-05-12
Added ggv cancel order messahe in both language

sunil-sharma:


app/models/offer.rb
```diff
@@ -215,6 +214,13 @@ def send_notification(text)
 
   private
 
+  def cancel_message
+    time = delivery.try(:schedule).try(:formatted_date_and_slot)
+    text = I18n.t("offer.ggv_cancel_message", time: time, locale: "en")
+    text += "<br/>"
+    text += I18n.t("offer.ggv_cancel_message", time: time, locale: "zh-tw")
```
steveyken: @sunil-sharma :+1 I didn't realise you could pass locale in here. Nice one!

sunil-sharma: Thanks @steveyken

PR #79, opened 2015-05-13
fixed creation of multiple deliveries for single offer and push-updates for delivery-deletion

swatijadhav:


app/controllers/api/v1/deliveries_controller.rb
```diff
@@ -124,6 +125,10 @@ def serializer
       Api::V1::DeliverySerializer
     end
 
+    def delete_existing_delivery
+      Delivery.where(offer_id: params[:delivery][:offer_id]).map(&:destroy)
```
steveyken: @swatijadhav If ```params[:delivery][:offer_id]``` evaluates to ```nil``` then we might have unexpected consequences here. Can you add a check to ensure it's not nil? Thanks.

app/models/delivery.rb
```diff
@@ -12,6 +12,7 @@ class Delivery < ActiveRecord::Base
   before_save :update_offer_state
   before_destroy :push_back_offer_state
   after_save :send_updates, if: "offer.scheduled?" && :process_completed?
+  after_destroy {send_updates :delete unless Rails.env.test? }
```
steveyken: Can we make tests pass without needing Rails.env.test? I'm a little wary of code that isn't run during tests. Thanks

PR #80, opened 2015-05-14
added column van_allowed to crossroads_transport

sunil-sharma:
schema.rb

### DIFF
--- a/app/serializers/api/v1/crossroads_transport_serializer.rb
+++ b/app/serializers/api/v1/crossroads_transport_serializer.rb
@@ -1,7 +1,7 @@
 module Api::V1
 
   class CrossroadsTransportSerializer < ApplicationSerializer
-    attributes :id, :name, :cost
+    attributes :id, :name, :cost, :is_van_allowed
 
     def name__sql
       "name_#{current_language}"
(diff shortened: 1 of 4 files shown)


steveyken:
@swatijadhav @sunil-sharma thinking a bit more about that... what if the column was called 'enabled' - then we'd have a way to enable the 1/8 truck but not the 2/8 truck. Then the logic for Truck disabled would be that all entries were set to enabled=false. What do you think?

swatijadhav:
Hi @steveyken,

Currently, the use-case behind adding the "is_van_allowed" column, is just to identify the 'Disable' option. Also at a time, only one crossroads-transport option is going to be available for single offer.

May be we are not getting the use-cases you are thinking of here.. 
We display all those records from 'crossroads_transports' table, do we want here to use 'enable' so that we can display only selected values? 
Correct us if we misunderstood it. Thanks.

steveyken:
ok that's fine for now then. can always re-evaluate if the requirements change.

sunil-sharma:
Thanks @Steve.

swatijadhav:
Thanks @steveyken :+1:

PR #81, opened 2015-05-19
Gcw 356 Replace ItemType with PackageType

swatijadhav:


mcm-ham:
We could simplify the code a bit if we simply sent SubPackageType over the API instead.

app/models/subpackage_type.rb
```diff
@@ -0,0 +1,7 @@
+class SubpackageType < ActiveRecord::Base
+  belongs_to :package_type
+  belongs_to :child_package_type, class_name: :PackageType, foreign_key: :subpackage_type_id
+
+  belongs_to :default_child_package_type, class_name: :PackageType, foreign_key: :subpackage_type_id
+  belongs_to :other_child_package_type, class_name: :PackageType, foreign_key: :subpackage_type_id
```
shivanibhanwal: What is the difference in code for default_child_package_type and other_child_package_type? How we distinguish between both?

PR #83, opened 2015-05-22
GCW-362 in push-updates, send required details of sender and avoid associated details

swatijadhav:
Hi @mcm-ham 

Can you please review this changes? 

Thanks.


mcm-ham:
The only issue I see is that address should be included for data sent to reviewers otherwise it'll be missing from the donor details page. My preferred approach is to keep all the logic about what to include or not in one location in the serializer so we don't have to remember to update a second location if we make changes in the future. So I suggest adding a property to UserSerializer that can override include_attribute and then send data to donor and staff separately in push_updates.

PR #101, opened 2015-06-23
GCW- 391 Logs for item edit/update

sunil-sharma:


app/models/version.rb
```diff
@@ -26,4 +34,22 @@ def to_s
...
+  end
+
+  def item_log?
+    self.item_type == "Item" &&
+    (self.object_changes.keys & %w(donor_description donor_condition_id state)).present?
+  end
+
+  def item_package_log?
+    self.item_type == "Package" && %w(received missing).include?(self.object_changes["state"][1])
+  end
```
steveyken: @sunil-sharma I wonder if it is possible to use the following papertrail setup to exclude some events. https://github.com/airblade/paper_trail#choosing-attributes-to-monitor

E.g. if we only care about changes to donor_description, donor_condition_id and state then we could use :skip or :only

For item_package_log, I think it would be better to send all Version changes to Ember (via PushUpdates) and filter on the client side. This is because we will probably want to Versions for another feature one day and we will run into problems if we are only sending certain updates from the API.

steveyken:
@sunil-sharma can you also add documentation for the new data - for apipie. Thanks.

app/models/version.rb
sunil-sharma: Thanks @steveyken, Sure, I will change it.
But what will happen with existing data and what if we need some other data in future?

PR #106, opened 2015-07-06
GCW 399 outbound calls to donors are routing via twilio

swatijadhav:


app/controllers/api/v1/twilio_controller.rb
```diff
@@ -3,13 +3,47 @@
...
-    after_filter :set_header, except: [:assignment, :hold_music]
+    after_filter :set_header, except: [:assignment, :hold_music, :twilio_generate_call_token]
     skip_authorization_check
     skip_before_action :validate_token
     skip_before_action :verify_authenticity_token
 
+    # OUTBOUND CALL : START
+    def connect_outbound_call
+      caller_id, offer_id, mobile = params["phone_number"].split("#")
+      redis.hmset("twilio_outbound_#{mobile}",
```
steveyken: @swatijadhav  I think it would be good to encapsulate the redis hmset/hmget calls into a service class. This would remove the underlying access and key implementation (twilio_outbound_#{mobile}) from the controller and mean that it would be easy to swap out with a different scheme later (say if we no longer wanted Redis or wanted to adjust the namespacing slightly)

app/controllers/api/v1/twilio_controller.rb
```diff
@@ -3,13 +3,47 @@
...
+          r.Say "Couldn't reach #{user(child_call.to).full_name} try again soon. Goodbye."
+        end
+        r.Hangup
+      end
+      render_twiml response
+    end
+
+    def outbound_call_status
+      offer_id = redis.hmget("twilio_outbound_#{child_call.to}", :offer_id)
+      user_id  = redis.hmget("twilio_outbound_#{child_call.to}", :caller_id)
```
steveyken: @swatijadhav As per comment above, it would be good to extract hmget into a service class. In this particular case it would mean you could make one redis call and extract both params rather than having to call redis twice here.

    tcm = TwilioCallManager.new(to: child_call.to)
    tcm.offer_id
    tcm.user_id

and

    class TwilioCallManager
    
      OutBoundCallPrefix = "twilio_outbound_"
      
      def initialize(options={})
        @call_to = options[:to]
      end
      
      def offer_id
        data[:offer_id]
      end
      
      def user_id
        data[:user_id]
      end
      
      private
      
      def data
        @data ||= redis.hmget("#{OutBoundCallPrefix}_#{@to}")
      end
      
      def redis
        @redis ||= Goodcity::Redis.new
      end
      
    end

app/jobs/send_outbound_call_status_job.rb
```diff
@@ -0,0 +1,23 @@
...
+    user  = User.find_by(id: user_id)
+    offer = Offer.find_by(id: offer_id)
+
+    if(user && offer)
+      text = if status == "completed"
+          "Called #{offer.created_by.full_name}"
+        else
+          "Call attempt failed: #{status.titleize}"
+        end
+
```
steveyken: @swatijadhav Think I commented somewhere already but I don't think this needs to be an active job as it is only creating one DB entry. i.e. it isn't calling an external service that has significant latency. Just a service class that handles the details should be fine. e.g. app/services/send_outbound_call_status.rb

app/controllers/api/v1/twilio_controller.rb
```diff
@@ -3,13 +3,47 @@
...
 
-    after_filter :set_header, except: [:assignment, :hold_music]
+    after_filter :set_header, except: [:assignment, :hold_music, :twilio_generate_call_token]
     skip_authorization_check
     skip_before_action :validate_token
     skip_before_action :verify_authenticity_token
 
+    # OUTBOUND CALL : START
+    def connect_outbound_call
+      caller_id, offer_id, mobile = params["phone_number"].split("#")
```
steveyken: We are passing in ```caller_id```, ```offer_id``` and ```mobile``` here but do we really need all this information? Would it be better simply to send ```offer_id``` and work out the ```caller_id``` as the authenticated donor and ```mobile``` from ```offer.created_by.mobile```

This would mitigate any param injection attacks and avoid duplication.

PR #111, opened 2015-07-14
Gcw 442 browse goods top level category 

swatijadhav:


app/models/package_sub_category.rb
```diff
@@ -0,0 +1,6 @@
+class PackageSubCategory < ActiveRecord::Base
+  belongs_to :package_type
+  belongs_to :package_category
+
+  validates :package_type_id, uniqueness: { scope: :package_category_id }
+end
```
steveyken: @swatijadhav my previous comment about whether we should rename to ```package_categories_package_types``` should have gone here. cc @mcm-ham

db/migrate/20150709152138_create_package_categories.rb
```diff
@@ -0,0 +1,11 @@
...
+  def change
+    create_table :package_categories do |t|
+      t.string :name_en
+      t.string :name_zh_tw
+      t.integer :parent_id
+
+      t.timestamps null: false
+    end
+  end
+end
```
steveyken: @swatijadhav be good to have an index on here too - for ```parent_id```

db/migrate/20150710135052_create_package_sub_categories.rb
```diff
@@ -0,0 +1,10 @@
+class CreatePackageSubCategories < ActiveRecord::Migration
+  def change
+    create_table :package_sub_categories do |t|
+      t.integer :package_type_id
+      t.integer :package_category_id
+
+      t.timestamps null: false
+    end
+  end
+end
```
steveyken: @swatijadhav  be good to have indexes on here as well - to avoid whole table scans.

lib/tasks/package_categories.rake
```diff
@@ -0,0 +1,52 @@
...
+      if package && package_category
+        PackageSubCategory.create(
+          package_type_id:     package.id,
+          package_category_id: package_category.id
+        )
+      end
+    end
+
+  end
+end
```
steveyken: @swatijadhav I think it would be great if this were a part of ```rake db:seeds``` too. Perhaps extract the method bodies into ```lib/import/package_category_importer.rb``` and call the relevant ```PackageCategoryImporter.import!``` from the rake tasks above and the ```rake db:seeds``` script. cc @mcm-ham

lib/tasks/package_categories.rake
```diff
@@ -0,0 +1,52 @@
...
+
+      "Household" => ["Appliances", "Baby", "Textiles / clothes",
+        "Kitchen, bedroom, bathroom, dining",  "Other household"],
+
+      "Small goods & bulk items" => ["Books", "Food", "Glasses", "Stationery",
+        "Tools", "Toys", "Medical", "Clothing & accessories",
+        "Mixed / miscellaneous", "Household textiles"],
+
+      "Recreation" => ["Music", "Other", "Playground", "Sport", "Bicycles"]
+    }
```
steveyken: @swatijadhav This would be great in a yaml file so it's easier to use in ```factory_girl``` tests. See the other taxonomies for the same approach. e.g. https://github.com/crossroads/api.goodcity/blob/master/spec/factories/sequences.rb#L21

spec/factories/package_categories.rb
```diff
@@ -0,0 +1,7 @@
+FactoryGirl.define do
+  factory :package_category do
+    name_en    "MyString"
+    name_zh_tw "MyString"
```
steveyken: @swatijadhav re comment above, we can wire this up to the real data using yaml files and factory_girl's ```generate``` method. e.g. https://github.com/crossroads/api.goodcity/blob/master/spec/factories/territories.rb#L5

steveyken:
thanks swati, good work. just a few comments and then good to merge.

PR #112, opened 2015-07-15
GCW-442-browse-goods-top-level-category

swatijadhav:
- send package-types and items details to browse app
- fixed package_type ability specs
- rspec and apipie-doc for package-category controller


app/controllers/api/v1/items_controller.rb
```diff
@@ -83,6 +84,10 @@ def update
       end
     end
 
+    def browse
+      render json: @items.accepted, each_serializer: serializer
+    end
+
```
steveyken: @swatijadhav cc @mcm-ham I think it might be good to have a separate ```BrowseController``` here to separate out the logic and future work. It should still return ```@items.accepted``` but I like the idea of separating the controller actions for the new app. That way we can be relatively sure that change in the browse.goodcity spec will have less of an impact on app.goodcity etc etc

spec/models/package_category_spec.rb
```diff
@@ -1,4 +1,15 @@
 require 'rails_helper'
 
 RSpec.describe PackageCategory, type: :model do
+  describe 'Association' do
+    it { is_expected.to have_many :package_sub_categories }
```
steveyken: @swatijadhav cc @mcm-ham am I right in thinking ```package_sub_categories``` is a join table? If so, should it be called ```package_categories_package_types``` ?

steveyken:
:+1: thanks @swatijadhav good work. just a couple of comments otherwise all good so far.

app/controllers/api/v1/items_controller.rb
swatijadhav: Sure @steveyken 
we will add separate `BrowseController` for browse app. Thanks.

PR #113, opened 2015-07-16
twilio-call-manager class for inbound-call details

steveyken:


app/services/twilio_inbound_call_manager.rb
```diff
@@ -0,0 +1,115 @@
...
+    "#{NotifyPrefix}_#{@user_id}"
+  end
+
+  def donor_id_key
+    "#{DonorPrefix}_#{@user_id}"
+  end
+
+  def store(key, value)
+    redis.set(key, value)
+    redis.expireat(key, Time.now.to_i + 60)
```
steveyken: @swatijadhav Looks like you can send the expireat key when you set the value. See https://github.com/redis/redis-rb/blob/98e3e7a516fc9b4609bc8ab482605f835a4de621/lib/redis.rb#L657

steveyken:
@swatijadhav Thanks! Good job :)

app/services/twilio_inbound_call_manager.rb
swatijadhav: Sure @steveyken 
Thanks for the reference :+1:

PR #120, opened 2015-08-13
Refactor Twilio Call Manager

steveyken:
for review @swatijadhav 

I've tried to clean up the TCM class a bit. Have converted class methods to instance methods and also refactored ```from_admin?``` to be ```caller_is_admin?``` This should make things a bit easier to read.


swatijadhav:
@steveyken :+1:

PR #125, opened 2015-10-07
GCW 629 receiving state for offer

swatijadhav:


app/models/offer.rb
```diff
@@ -133,9 +144,13 @@ def set_initial_state
...
-      if offer.try(:delivery).try(:gogovan_order).try(:status) != 'cancelled'
-        offer.try(:delivery).try(:gogovan_order).try(:cancel_order)
-      end
+      ggv_order = offer.try(:gogovan_order)
+      ggv_order.try(:cancel_order) if ggv_order.try(:status) != 'cancelled'
+    end
+
+    after_transition :on => :start_receiving do |offer, transition|
+      ggv_order = offer.try(:gogovan_order)
+      ggv_order.try(:cancel_order) if ggv_order.try(:pending?)
```
steveyken: Good refactor here. I also wonder if there is a way to remove the knowledge of 'cancelled' and 'pending' state for the GGV Order. Would it be sufficient to call `ggv_order.cancel` and for that to work out whether it needs to cancel? Is there any case in which we don't want to cancel an existing ggv_order when we are closing, re_reviewing or it's in pending?

PR #126, opened 2015-10-17
Delivery abilities

steveyken:
The following CanCanCan definition for `Delivery` was too open. It allows any logged in user to perform these actions on *any* delivery object, not just their own.

    can [:create, :show, :update, :destroy, :confirm_delivery], Delivery

This has been tightened to only allow Donors to perform actions on deliveries that belong to their own offers.

    can [:create], Delivery
    if reviewer || supervisor
      can [:index, :show, :update, :destroy, :confirm_delivery], Delivery
    else
      can [:show, :update, :destroy, :confirm_delivery], Delivery, offer_id: user_offer_ids
    end

Specs have been added to confirm the permissions are locked down.

The `DeliveriesController#delete_existing_delivery` method has also been tightened to check permissions before deleting existing deliveries. As a general rule, user param data should be considered highly dangerous and not be allowed to directly influence actions such as delete without proper checking.

    Delivery.where(offer_id: params[:delivery][:offer_id]).map(&:destroy)

becomes

    offer_id = params[:delivery][:offer_id]
    Delivery.where(offer_id: offer_id).each do |delivery|
      authorize!(:destroy, delivery)
      delivery.destroy
    end

cc @swatijadhav

### DIFF
--- a/app/controllers/api/v1/deliveries_controller.rb
+++ b/app/controllers/api/v1/deliveries_controller.rb
@@ -128,7 +128,11 @@ module Api::V1
     end
 
     def delete_existing_delivery
-      Delivery.where(offer_id: params[:delivery][:offer_id]).map(&:destroy)
+      offer_id = params[:delivery][:offer_id]
+      Delivery.where(offer_id: offer_id).each do |delivery|
+        authorize!(:destroy, delivery)
+        delivery.destroy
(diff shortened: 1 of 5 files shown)


swatijadhav:
@steveyken  :+1:

PR #130, opened 2015-11-23
GCW-750: send native-notification for newly submitted-offer

swatijadhav:


app/models/offer.rb
```diff
@@ -215,7 +215,7 @@ def assign_reviewer(reviewer)
   end
 
   def send_new_offer_notification
-    PushService.new.send_notification Channel.reviewer, true, {
```
shivanibhanwal: @swatijadhav so earlier this code was sending message to whom? because new offer does not have any reviewer assigned? Just trying to understand.

swatijadhav: Earlier, it was sending it over `reviewer` channel. We send native notifications to only channels with `user_` suffix and not to `reviewer` or `supervisor` channels.
Hence updated it to send over the all reviewers channels.

shivanibhanwal: ok got it thanks! :+1:

PR #133, opened 2015-11-27
send versions using to_json

swatijadhav:


steveyken:
:+1: what are the speed improvements for this?

steveyken:
feel free to merge now if you're finished.

swatijadhav:
@steveyken Here we have used `to_json` and user's full_name is evaluated in the sql query itself, but it was not taken by the AMS, it was still sending the extra db query to find `user`
hence we are using the `to_json`. The overall tame taken to fetch versions is reduced to half.

PR #135, opened 2015-11-30
Refactored model, controller and services code.

swatijadhav:


app/controllers/api/v1/deliveries_controller.rb
```diff
@@ -104,14 +104,22 @@ def confirm_delivery
...
-        render json: @delivery.errors.to_json, status: 422
+        render_errors
       end
     end
 
     private
 
+    def render_delivery_json
+      render json: @delivery, serializer: serializer
+    end
```
steveyken: @swatijadhav I'm all for refactoring code particularly where it's re-used a few times but I think one-line methods that don't introduce any new logic aren't necessary. We're just adding an extra function call in the stack.

So for example, this method doesn't add any value. It just passes parameters onto the next function.

    def render_delivery_json
      render json: @delivery, serializer: serializer
    end

But this one does add value because it transforms the errors into a json format.

    def render_errors
      render json: @delivery.errors.to_json, status: 422
    end

Does that make sense?

steveyken:
Great work @swatijadhav  :100:

app/controllers/api/v1/deliveries_controller.rb
swatijadhav: yes @steveyken agreed.
Just added it, as it was repeated at two places, in `update` and `show` methods.
I will revert it back. Thanks for reviewing :)

swatijadhav: Changes done https://github.com/crossroads/api.goodcity/commit/36f727
cc: @steveyken

PR #137, opened 2015-12-07
Refactoring

swatijadhav:
https://codeclimate.com/github/crossroads/api.goodcity/compare/refactoring#ratings

### DIFF
--- a/app/controllers/api/v1/deliveries_controller.rb
+++ b/app/controllers/api/v1/deliveries_controller.rb
@@ -100,7 +100,7 @@ module Api::V1
     end
     def confirm_delivery
       @delivery = Delivery.find_by(id: params["delivery"]["id"])
-      delete_old_associations
+      @delivery.delete_old_associations
       @delivery.gogovan_order = GogovanOrder.book_order(current_user,
         order_params) if params["gogovanOrder"]
       if @delivery && @delivery.update(get_delivery_details)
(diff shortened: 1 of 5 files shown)


app/controllers/api/v1/deliveries_controller.rb
```diff
@@ -139,22 +139,21 @@ def delete_existing_delivery
       end
     end
 
-    def delete_old_associations
```
steveyken: :+1:

app/models/ability.rb
```diff
@@ -138,4 +122,49 @@ def initialize(user)
...
+    @reviewer || @supervisor
+  end
+
+  def action_ability(actions, class_name, records)
+    can actions, class_name, records do |record|
+      record.item.offer.created_by_id == @user_id
+    end
+  end
 end
+
```
steveyken: @swatijadhav Overall, I like what you've done here as it separates various sections into their own function. I think `csui` and `action_ability` makes things a bit harder to read by hiding implementation. If you could revert just those bits, I think this will be ok. When doing a security audit, I really want to see the logic all together so it's clear exactly what each permission covers.

steveyken:
@swatijadhav on the whole this is great, thanks Swati. Just a couple of comments on the `ability.rb` file and then I think we're good to merge.

swatijadhav:
Code climate score for this branch is now **2.87 GPA** (+0.07) :smile:

steveyken:
:)

PR #139, opened 2015-12-16
rake task to schedule GGV jobs

swatijadhav:


lib/tasks/queue_ggv_polling.rake
```diff
@@ -0,0 +1,9 @@
+namespace :goodcity do
+
+  # rake goodcity:queue_ggv_polling
+  desc 'Queue GGV polling job for GGV-orders'
+  task queue_ggv_polling: :environment do
+    orders = GogovanOrder.where("status IN (?)", ["active", "pending"])
+    orders.each{ |order| PollGogovanOrderStatusJob.perform_later(order.id) }
```
steveyken: @swatijadhav we need to check if the job is already in the queue. If not, we'll double up.

swatijadhav:
@steveyken Can you please verify it now? Thanks.

PR #142, opened 2016-01-07
Gcw 808 location changes

swatijadhav:
Run following rake task on staging and production, after merging changes:

        rake goodcity:add_stockit_locations


app/models/ability.rb
```diff
@@ -122,6 +122,9 @@ def public_ability
     can :index, Timeslot
     can :index, GogovanTransport
     can :index, CrossroadsTransport
+
+    # Required in stockit
+    can [:index, :create], Location
```
steveyken: @swatijadhav can we make this staff only

steveyken:
Pending the comments above, this is good to go. Thanks!

PR #143, opened 2016-01-07
GCW-809 designation_label column for package

swatijadhav:


steveyken:
@swatijadhav Sorry, can we change `designation_label` to `designation_name` - it will match stockit. Sorry I misled you before. Afterwards, this is good to merge.

swatijadhav:
Sure @steveyken will rename it. Thanks

PR #144, opened 2016-01-11
Handle stockit updates and update attribute-mapping

swatijadhav:


app/controllers/api/v1/packages_controller.rb
```diff
@@ -42,12 +43,13 @@ def show
     api :POST, "/v1/packages", "Create a package"
     param_group :package
     def create
-      @package = Package.new(package_params)
-      @package.offer_id = offer_id
-      if @package.save
-        render json: @package, serializer: serializer, status: 201
-      else
-        render json: @package.errors.to_json, status: 422
+      if package_record
```
steveyken: @swatijadhav if `package_record` is false (i.e. is `existing_package` is false), then we won't render a response?

swatijadhav: @steveyken yes we are not passing any response, by default it sent `200` response.
Shall we pass any response here also if `existing_package` is false?

steveyken: Yes, I think it would be good to send an explicit response, otherwise we might miss a scenario one day.

swatijadhav: Sure @steveyken

app/controllers/api/v1/packages_controller.rb
```diff
@@ -14,7 +16,7 @@ class PackagesController < Api::V1::ApiController
 
     def_param_group :package do
       param :package, Hash, required: true do
-        param :quantity, Integer, desc: "Package quantity", allow_nil: true
+        param :quantity, lambda { |val| [String, Fixnum].include? val.class }, desc: "Package quantity", allow_nil: true
```
steveyken: @swatijadhav how are we getting a Fixnum here? params should always be strings. Was it a test?

app/controllers/api/v1/packages_controller.rb
```diff
@@ -88,17 +94,54 @@ def add_item_to_stockit
...
     end
 
     def offer_id
-      Item.where(id: @package.item_id).pluck(:offer_id).first
+      @package.item.offer_id
     end
 
+    def package_record
+      if package_params[:inventory_number]
+        if existing_package
```
steveyken: can you make this `Package.find_by(inventory_number: package_params[:inventory_number])` and remove `existing_package` as it makes the logic easier to follow. I wasn't sure what `existing_package` means, otherwise we can rename it from `existing_package` to `find_package_by_inventory_number` but I think it only saves us one line.

app/controllers/api/v1/packages_controller.rb
```diff
@@ -88,17 +94,54 @@ def add_item_to_stockit
...
+    end
+
+    def item_params
+      params["package"].require(:item).permit(:donor_condition_id)
+    end
+
+    def save_item_details
+      params["package"]["item"] &&
+      (attributes = item_attributes) &&
+      @package.item.update_attributes(attributes)
```
steveyken: Do we need local variable `attributes` assignment here? What about

     params["package"]["item"] && item_attributes && @package.item.update_attributes(item_attributes)

app/controllers/api/v1/packages_controller.rb
```diff
@@ -88,17 +94,54 @@ def add_item_to_stockit
...
+      end
+    end
+
+    def location_id
+      Location.find_by(stockit_id: package_params[:location_id]).try(:id)
+    end
+
+    def existing_package
+      @package = Package.find_by(inventory_number: package_params[:inventory_number])
+    end
```
steveyken: I found I don't really understand what this method does from the name `existing_package`. I don't mind too much if it's separate from the `existing_package` method but it's hard to understand the logic flow of `existing_package` without having to actually read what this method does. A better name would be `find_package_by_inventory_number` and then it would make sense in the logic flow of `existing_package`

However, since this is a one-line method, I would be tempted to remove it anyway and save the call stack. I find one-line methods can make the code harder to read and understand. They can be justified if they are called in several places but my brain only has a limited stack!!

app/controllers/concerns/goodcity_sync.rb
```diff
@@ -0,0 +1,23 @@
+module GoodcitySync
+
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :set_request_from_stockit, only: [:update, :destroy, :create]
```
steveyken: @swatijadhav we can't really infer from these actions that the request actually comes in from Stockit. What if we start editting packages through Ember next week/month? However, once we do the authentication ticket, we will know that the request has come in from an API user, so no need to change this until that happens.

app/jobs/stockit_update_job.rb
```diff
@@ -8,7 +8,7 @@ def perform(package_id)
       response = Stockit::Browse.new(package).update_item
 
       if response && (errors = response["errors"] || response[:errors])
-        log_text = "Inventory: #{inventory_number} Package: #{package.id}"
+        log_text = "Inventory: #{package.inventory_number} Package: #{package.id}"
```
steveyken: :+1: good catch

app/models/ability.rb
```diff
@@ -125,6 +125,7 @@ def public_ability
 
     # Required in stockit
     can [:index, :create], Location
+    can :create, Package
```
steveyken: Can you annotate this with #TODO to make a note to tighten this up when we add authentication... this is very open right now!

app/controllers/api/v1/packages_controller.rb
```diff
@@ -1,6 +1,8 @@
 module Api::V1
   class PackagesController < Api::V1::ApiController
+    include GoodcitySync
 
+    skip_before_action :validate_token, only: :create
```
steveyken: let's stick a #TODO on this as we must remember to tighten it up

app/controllers/api/v1/packages_controller.rb
swatijadhav: @steveyken purpose to add `attributes` local variable is just to avoid extra method call `item_attributes`
As in `item_attributes`, we look for donor-condition, which might cause extra `find_by` query again.
Please let me know your thoughts, Thanks.

app/controllers/api/v1/packages_controller.rb
swatijadhav: @steveyken 

when we receive request from gc-admin 

    Started POST "/api/v1/packages" for 122.169.90.136 at 2016-01-14 11:52:09 +0530
    Processing by Api::V1::PackagesController#create as JSON
    Parameters: {"package"=>{"quantity"=>1, "length"=>6, "width"=>54, "height"=>89, "notes"=>"Test item", "state"=>"expecting", "state_event"=>nil, "received_at"=>nil, "rejected_at"=>nil, "created_at"=>nil, "updated_at"=>nil, "image_id"=>nil, "offer_id"=>519, "inventory_number"=>nil, "item_id"=>"811", "package_type_id"=>"99", "location_id"=>nil}}

When we recieve request from stockit

    Started POST "/api/v1/packages.json" for 127.0.0.1 at 2016-01-14 11:54:02 +0530
    Processing by Api::V1::PackagesController#create as JSON
    Parameters: {"package"=>{"designation_name"=>"Lapsap - China - Hong Kong (Special Administrative Region)", "item"=>{"donor_condition_id"=>"Lightly Used"}, "notes"=>"kk kkk kkkk", "inventory_number"=>"E55668", "location_id"=>"214", "quantity"=>"4"}}

here, the quantity field value is in string and fixnum format.

app/controllers/api/v1/packages_controller.rb
```diff
@@ -42,12 +44,16 @@ def show
...
+        @package.offer_id = offer_id
+        if @package.save
+          save_item_details
+          render json: @package, serializer: serializer, status: 201
+        else
+          render json: @package.errors.to_json, status: 422
+        end
       else
-        render json: @package.errors.to_json, status: 422
+        render nothing: true, status: 204
```
steveyken: I think this case is where we should create the package normally - as if it had come in from a non-stockit request. i.e. 

      @package = Package.new(package_params)
      @package.offer_id = offer_id
      if @package.save
        render json: @package, serializer: serializer, status: 201
      else
        render json: @package.errors.to_json, status: 422
      end

app/controllers/api/v1/packages_controller.rb
swatijadhav: Sure @steveyken, I will update it. Thanks.

app/controllers/api/v1/packages_controller.rb
steveyken: Interesting... I see '.json' is missing from packages... perhaps that affects how the values are parameterized. I knew Rails used to do it but I thought they stopped as things like symbols (which live forever) could cause denial of service issues when you run out of memory.

app/controllers/api/v1/packages_controller.rb
steveyken: Oops! Yes, that's fine then. Leave it as it is. I thought `item_attributes` was a variable not a method!

PR #154, opened 2016-01-26
Stockit service refactoring

steveyken:
I've refactored the Stockit API classes to have their own namespace and cleaned up the calls. Stockit::Location and Stockit::Item now have their own classes and share a common Stockit::Base mixin module.

Added rspec tests for StockitDeleteJob and StockitUpdateJob and fixed case where response["errors"] || response[:errors] code was needed. This was happening because we had the following code:

    {
      "errors": { ... }
    }

Using `"errors": xyz` turns the "errors" string into a symbol. Just set it to always be a string.

@swatijadhav  can you review?


app/jobs/stockit_delete_job.rb
```diff
@@ -1,13 +1,17 @@
 class StockitDeleteJob < ActiveJob::Base
   queue_as :stockit_updates
 
-  def perform(inventory_number)
-    response = Stockit::Browse.new(inventory_number).remove_item
+  def perform(package_id)
+    package = Package.find_by(id: package_id)
```
swatijadhav: @steveyken Previously, we were using `inventory_number` because we are calling this job while deleting package.
Sidekiq is fast enough that it will perform this job before the deletion of package.
But in case, if this job queued and tries to execute after the package is deleted, then this might not delete `item` from stockit.

Please let me know your thoughts. Thanks :)

app/services/stockit/item.rb
```diff
@@ -0,0 +1,96 @@
...
+      end
+
+      def update(package)
+        new(package).update
+      end
+
+      def delete(package)
+        new(package).delete
+      end
+    end
```
swatijadhav: :+1:

swatijadhav:
Nicely Done @steveyken :+1: 
Thanks :)

app/jobs/stockit_delete_job.rb
steveyken: ah good catch! I had replaced inventory_number with the package object in Stockit::Item so that everything was consistent but I understand the problem you mean. Will have a think and update the PR.

swatijadhav: Sure @steveyken

steveyken: 891badc fixes this

PR #157, opened 2016-02-02
Package state refactor

steveyken:
The thinking here is that we can encapsulate the required model state within a workflow action. In Ember, we can call the correct `state_event` and the API ensures the model is in a consistent state. We also move the `inventory_number=nil` mutation to the `receive` action to fix an edge case where item is marked missing but already has a barcode label on it and is then subsequently added back into inventory. It will keep the same label.

### DIFF
--- a/app/models/package.rb
+++ b/app/models/package.rb
@@ -32,14 +32,22 @@ class Package < ActiveRecord::Base
     state :expecting, :missing, :received
 
     event :mark_received do
-      transition [:expecting, :missing, :received] => :received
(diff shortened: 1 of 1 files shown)


steveyken:
@swatijadhav what do you think? We would need to update the respective actions in admin.goodcity and I will fix tests. Just wanted to get your take first.

swatijadhav:
@steveyken Yes this would be good, and also better to understand the use-case. :+1:

steveyken:
Closing this PR. Feature wasn't completed or merged.

PR #158, opened 2016-02-04
Gcw 862 add barcode printing

mcm-ham:


app/controllers/api/v1/packages_controller.rb
```diff
@@ -75,6 +75,26 @@ def destroy
       render json: {}
     end
 
+    api :POST, "/v1/packages/print_barcode", "Print barcode"
+    def print_barcode
+      begin
+        package = Package.find params[:package_id]
+      rescue ActiveRecord::RecordNotFound
+        return render json: {errors:"Package not found with supplied package_id"}, status: 400
+      end
```
swatijadhav: Any specific reason to use begin-rescue block here? It can be handled here https://github.com/crossroads/api.goodcity/blob/master/app/controllers/api/v1/api_controller.rb#L6 right?
cc: @mcm-ham @steveyken

swatijadhav:
:+1:

app/controllers/api/v1/packages_controller.rb
mcm-ham: I'm used to 404 meaning Uri does not exist e.g. /api/package/1, but since the Uri does exist in this case /api/package/print_barcode I'm returning a 400 to indicate invalid request.

PR #176, opened 2016-04-26
Gcw 904 add stockit orders

swatijadhav:
- Configure database.yml file to access Stockit DB 
- [X] development 
- [X] testing 
- [ ] staging 
- [ ] production

- Implement Stockit models and serializers.
- Access orders
- [X] first set of matching 30 results
- [X] order: latest first
- [X] search based on order.code , organisations.name, local_orders.client_name, contacts.first_name , contacts.last_name, contacts.mobile_phone_number, contacts.phone_number


app/serializers/api/v1/stockit/designation_serializer.rb
```diff
@@ -0,0 +1,19 @@
...
+    has_one :contact, serializer: ContactSerializer
+    has_one :organisation, serializer: OrganisationSerializer
+    has_one :local_order, serializer: LocalOrderSerializer
+
+    def local_order_id
+      object.local_order_id
+    end
+
+    def local_order_id__sql
+      object.local_order_id.blank? ? "null" : "'#{object.local_order_id}'"
```
steveyken: Would be good to sanitize this by default so that a rogue local_order_id can't inject

PR #182, opened 2016-05-06
GCW-903 update auth for stock-app

mcm-ham:


app/models/ability.rb
```diff
@@ -52,6 +52,10 @@ def delivery_abilities
     end
   end
 
+  def designation_abilities
+    can [:index], Stockit::Designation if staff?
+  end
```
steveyken: @mcm-ham Would be good to add some tests for this in spec/models/abilities/designation_abilities.rb

PR #218, opened 2016-08-03
GCW-1133 package can be removed from set

swatijadhav:


app/controllers/api/v1/packages_controller.rb
```diff
@@ -146,6 +146,11 @@ def move_stockit_item
       send_stock_item_response
     end
 
+    def remove_from_set
+      @package.update(set_item_id: 0)
```
steveyken: @swatijadhav would nil be better here?

PR #225, opened 2016-08-22
GCW-1208 sync location-deletion

swatijadhav:


app/controllers/api/v1/locations_controller.rb
```diff
@@ -43,6 +43,11 @@ def create
       end
     end
 
+    def destroy
+      Location.find_by(stockit_id: params[:id]).try(:destroy)
+      render json: {}
+    end
```
steveyken: @swatijadhav this works for incoming stockit DELETE requests but not from the standard ember REST (Imagine if we wanted to add the ability to delete locations in the GoodCity app.) Can we alter this slightly to cater for both cases so we are future-proof?

    location = 
      if (id = params[:stockit_id]).present?
        Location.find_by(stockit_id: id)
      else
        Location.find(params[:id])
      end
    location.try(:destroy)

PR #229, opened 2016-08-26
GCW-1204 sync countries and all types of designations

swatijadhav:


app/controllers/api/v1/countries_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+        render json: {}, status: 201
+      else
+        render json: @country.errors.to_json, status: 422
+      end
+    end
+
+    private
+
+    def country_record
+      @country = Country.where(stockit_id: country_params[:stockit_id]).first_or_initialize
```
steveyken: @swatijadhav similar to the other day, I'd like these controllers to continue to follow the conventions of assuming params[:id] first. That way it will be easy to add a country controller to GoodCity ember project and expect it to behave as any normal controller.

Can we do something like 

    def country_record
      if params[:id].present?
        @country = Country.where(id: params[:id]).first_or_initialize
      elsif country_params[:stockit_id].present?
        @country = Country.where(stockit_id: country_params[:stocki_id]).first_or_initialize
      else
        raise ActiveRecord::NotFound
      end
    end

PR #242, opened 2016-09-16
rename stockit_designation to order

swatijadhav:


app/controllers/api/v1/packages_controller.rb
```diff
@@ -155,13 +155,13 @@ def move_stockit_item
     def remove_from_set
       @package.remove_from_set
       render json: @package, serializer: stock_serializer, root: "item",
-        include_stockit_designation: true
+        include_order: true
```
swatijadhav: @steveyken `include_order` is an extra attribute passed to serilizer. It means when it serializes package record, it should include associated order details when  `include_order` is `true`

used here https://github.com/crossroads/api.goodcity/pull/242/files#diff-6084b0e0adc025a09fa6a89c04460521R24

Thanks.

swatijadhav:
Thanks for reviewing changes @steveyken. Changes done as per the suggestion. 

Henceforth, whenever we need to access records directly using model-name, we will access it as:

    ModelName.accessible_by(current_ability)

For more details: refer https://github.com/ryanb/cancan/wiki/authorizing-controller-actions#index-action

PR #243, opened 2016-09-21
Live updates for images on Browse

Darshan-Chauhan:
Hi,

This PR has some changes for live image updates on browse app.
Take a look and let me know if any further changes are required.

Thanks.

### DIFF
--- a/app/models/concerns/push_updates.rb
+++ b/app/models/concerns/push_updates.rb
@@ -33,8 +33,8 @@ module PushUpdates
 
   def browse_updates(operation)
     operation = is_browse? ? operation : "delete"
-    json = Api::V1::BrowsePackageSerializer.new(self).as_json[:browse_package]
-    data = { item: { package: json }, operation: operation }
+    json = Api::V1::BrowsePackageSerializer.new(self).as_json
+    data = { item: { package: json[:browse_package], images: json[:images]  }, operation: operation }
     service.send_update_store(Channel.browse, false, data)
   end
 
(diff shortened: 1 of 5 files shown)


app/models/package.rb
```diff
@@ -26,6 +26,7 @@ class Package < ActiveRecord::Base
   before_save :save_inventory_number, if: :inventory_number_changed?
   before_save :update_set_relation, if: :stockit_sent_on_changed?
   after_commit :update_set_item_id, on: :destroy
+  after_touch {update_client_store :update unless Rails.env.test? }
```
steveyken: I'd prefer to remove the `unless Rails.env.test?` and stub out push updates in the test, globally.

app/models/concerns/push_updates.rb
```diff
@@ -33,8 +33,8 @@ def update_client_store(operation)
 
   def browse_updates(operation)
     operation = is_browse? ? operation : "delete"
-    json = Api::V1::BrowsePackageSerializer.new(self).as_json[:browse_package]
-    data = { item: { package: json }, operation: operation }
+    json = Api::V1::BrowsePackageSerializer.new(self).as_json
+    data = { item: { package: json[:browse_package], images: json[:images]  }, operation: operation }
```
steveyken: Should the `images` detail go inside the serializer? or perhaps a BrowseItemSerializer that includes `package` and `images` nodes? This would make it easier to extend in the future. What do you think?

Darshan-Chauhan: @steveyken , BrowseItemSerializer is just used to pass the parent-item details(package-type, donor-condition). We only send package-details to browse app using  BrowsePackageSerializer and item-details are send via BrowsePackageSerializer.
Here while sending updates we only need updates of package and images hence using BrowsePackageSerializer

PR #256, opened 2016-11-07
Nov 2016 release

steveyken:


app/controllers/api/v1/orders_controller.rb
```diff
@@ -56,20 +69,33 @@ def recent_designations
...
+        @order.created_by = current_user
+        @order.detail_type = "GoodCity"
+      end
       @order
     end
 
     def order_params
-      params.require(:order).permit(:stockit_id, :code, :status, :created_at, :stockit_contact_id, :detail_id, :detail_type, :stockit_organisation_id, :description, :stockit_activity_id)
+      params.require(:order).permit(:stockit_id, :code, :status, :created_at,
+        :stockit_contact_id, :detail_id, :detail_type, :description, :state, :state_event, :stockit_organisation_id, :stockit_activity_id, :purpose_description,
```
steveyken: Technically, it would be possible for a user to change their own order status here. We might want to restrict that further down the line.

PR #259, opened 2016-12-03
GCW-1354 updated redis gem version to 3.3.2

arthi-j:
The updated redis gem has TLS/SSL support. The REDIS_URL environment variable on the server where api.goodcity is deployed has to be updates to point to the redis ssl port on Azure.

The format of the url is -
rediss://[azure redis host]:[ssl port]


steveyken:
I've merged this in with master manually to resolve the conflict.

PR #262, opened 2016-12-08
Orders_packages and packages related changes

Darshan-Chauhan:


Darshan-Chauhan:
**Note**: we need to merge changes of StockIt first, because this changes affects StockIt also.

app/models/orders_package.rb
```diff
@@ -20,4 +22,95 @@ def set_initial_state
       transition :requested => :designated
     end
   end
+
+  def self.update_partially_designated_item(package)
+    orders_package = OrdersPackage.find(package[:orders_package_id].to_i)
```
namrataukirde: Instead of `find` it is better to to use find_by_id or where with first. Find will break if record not foud.

app/models/orders_package.rb
```diff
@@ -20,4 +22,95 @@ def set_initial_state
...
+      updated_by: User.current_user,
+      state: "designated"
+      )
+  end
+
+  def self.find_packages(order_id, package_id)
+    orders_packages = OrdersPackage.where("order_id = ? and package_id = ?", order_id, package_id)
+  end
+
+  def self.find_records(order_id)
```
namrataukirde: can we rename method name to `get_records_associated_with_order_id` or something like this ? That will help us to understand purpose of method with name itself.

app/models/orders_package.rb
```diff
@@ -20,4 +22,95 @@ def set_initial_state
...
+    OrdersPackage.where(order_id: order_id)
+  end
+
+  def self.get_total_dispatched_qty(package_id)
+    packages = OrdersPackage.where("package_id = ? and state = ?", package_id, "dispatched")
```
namrataukirde: Can we add scope for this as it is redundant ?

app/models/orders_package.rb
```diff
@@ -20,4 +22,95 @@ def set_initial_state
...
+    orders_package = OrdersPackage.find(orders_package_id)
+    orders_package.update(sent_on: Date.today, state: "dispatched")
+  end
+
+  private
+  def recalculte_quantity
+    total_quantity = 0
+    orders_packages = OrdersPackage.get_designated_and_dispatched_packages(package_id)
+    designate_orders_packages = OrdersPackage.get_designated_packages(package_id)
+    if(designate_orders_packages.length == 1)
```
namrataukirde: can we move it to another method?

app/models/orders_package.rb
```diff
@@ -20,4 +22,95 @@ def set_initial_state
...
+      Package.update_designation(designate_orders_packages.first.package_id, designate_orders_packages.first.order_id)
+    elsif(designate_orders_packages.length == 0)
+      Package.remove_designation(package_id)
+    end
+    orders_packages.each do |orders_package|
+      total_quantity += orders_package.quantity
+    end
+    Package.update_in_stock_quantity(package_id, total_quantity)
```
namrataukirde: can we add method that will return us `total qty` and use it here ?

app/controllers/api/v1/orders_packages_controller.rb
```diff
@@ -0,0 +1,43 @@
+module Api::V1
+  class OrdersPackagesController < Api::V1::ApiController
+    skip_authorization_check
```
steveyken: @Darshan-Chauhan what should be the user access policy here?

app/controllers/api/v1/orders_packages_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+    end
+
+    api :GET, '/v1/orders_packages', "List all orders_packages"
+    def index
+      return search if params['search_by_order_id'].present?
+    end
+
+    def search
+      @orders_packages = OrdersPackage.find_records(params["search_by_order_id"])
```
steveyken: Can we use the standard CanCanCan methods here to ensure that only records that the user is allowed to view are shown. `@orders_packages` should be instantiated if the authorization checks are run.

app/models/item.rb
```diff
@@ -118,15 +118,35 @@ def update_stockit_item
...
   end
 
-  def designate_set_to_stockit_order(order_id)
+  def update_designation(params)
     inventory_packages.set_items.each do |package|
-      package.designate_to_stockit_order(order_id)
+      orders_package_with_package_id = OrdersPackage.find_packages(package.order_id, package.id)
+      orders_package_with_params_id = OrdersPackage.find_packages(params[:order_id], package.id)
+      OrdersPackage.update_designation(orders_package_with_package_id, params[:order_id])
+      OrdersPackage.delete_unwanted_cancelled_packages(orders_package_with_params_id, params[:order_id])
```
steveyken: @Darshan-Chauhan there are a lot of class methods here... I wonder if they would be better off as instance methods on `OrdersPackage`

app/models/orders_package.rb
```diff
@@ -1,16 +1,18 @@
 class OrdersPackage < ActiveRecord::Base
   belongs_to :order
   belongs_to :package
-  belongs_to :reviewed_by, class_name: 'User'
+  belongs_to :updated_by, class_name: 'User'
 
   after_initialize :set_initial_state
+  after_create :recalculte_quantity
```
steveyken: typo: `recalculate_quantity`

app/models/orders_package.rb
steveyken: Agree re renaming but I also think this method could be refactored into a `scope` via has_many relationship instead?

    scope :by_order_id(order_id), -> { where(order_id: order_id) }

app/models/orders_package.rb
```diff
@@ -20,4 +22,95 @@ def set_initial_state
...
+  def self.get_total_dispatched_qty(package_id)
+    packages = OrdersPackage.where("package_id = ? and state = ?", package_id, "dispatched")
+    total_dispatched_qty = 0
+    packages.each do |orders_package|
+      total_dispatched_qty += orders_package.quantity
+    end
+    total_dispatched_qty
+  end
+
+  def self.update_designation(orders_package, order_id)
```
steveyken: I wonder if these `self` methods can be refactored into instance methods. In fact, for some, you can iterate over objects and use existing active_record update_columns.

So instead of

    OrdersPackage.update_designation(orders_package, order_id)

Can have something more like

    class OrdersPackage
      def update_designation(order_id)
        update(order_id, order_id)
      end
    end

    orders_package.update_designation(order_id)

In general, I think class methods need to be avoided where possible.

app/models/orders_package.rb
steveyken: Crazy idea but what if you extracted a lot of these package designation methods out of OrderPackages and into their own domain service class. So leave the OrderPackage class to deal only with the ActiveRecord database stuff and put all the 'logic' behind adding/updating/removing designations into a OrderPackageDesignation class?

app/models/orders_package.rb
Darshan-Chauhan: @steveyken  has_many relationship do not accept any arguments, we can try with do..end block with has_many and writing scope in it but that seems ugly code. So is it okay If i just write scope(the one that you've suggested before)?

steveyken:
Good job on the refactoring Darshan. Looking good to merge. Can you fix the one failing test and when new rspecs are done and passing it's ready to merge to master.

Darshan-Chauhan:
Sure @steveyken , I'll fix that test when I start working on new rspec tests.

app/services/stockit/orders_package_sync.rb
```diff
@@ -0,0 +1,105 @@
...
+        package: package_params
+      }
+    end
+
+    def item_params
+      {
+        quantity: @orders_package.quantity,
+        code_id: package.package_type.try(:stockit_id),
+        inventory_number: add_stockit_prefix(package.inventory_number),
+        case_number: package.case_number.blank? ? nil : package.case_number,
```
namrataukirde: We can use `presence` method here.

app/services/stockit/orders_package_sync.rb
```diff
@@ -0,0 +1,105 @@
+require 'stockit/base'
+
+module Stockit
+  class OrdersPackageSync
+
+    include Stockit::Base
+
+    attr_accessor :package, :errors, :offset, :per_page
+
+    def initialize(package = nil, orders_package = nil, offset = nil, per_page = nil)
```
namrataukirde: Do we need nil for all arguments here (i.e all args are optional)?

app/services/stockit/orders_package_sync.rb
Darshan-Chauhan: I think blank? is more convenient  for empty string then presence

app/services/stockit/orders_package_sync.rb
Darshan-Chauhan: No, we don't. I'll update it

app/services/stockit/orders_package_sync.rb
namrataukirde: But if its blank then you are assigning nil that can be done as `case_number: package.case_number.presence`

Darshan-Chauhan: 👍

app/jobs/stockit_sync_orders_package_job.rb
```diff
@@ -0,0 +1,22 @@
+class StockitSyncOrdersPackageJob < ActiveJob::Base
+  queue_as :stockit_sync_orders_package_updates
+
+  def perform(package_id, orders_package, operation)
+    package = Package.find_by(id: package_id)
+
+    if package
+      if(operation == "create")
```
steveyken: A case / switch statement might be nice here to tidy the code up. (non-essential)

app/jobs/stockit_sync_orders_package_job.rb
```diff
@@ -0,0 +1,22 @@
+class StockitSyncOrdersPackageJob < ActiveJob::Base
+  queue_as :stockit_sync_orders_package_updates
+
+  def perform(package_id, orders_package, operation)
```
steveyken: is `orders_package` a simple ruby object? i.e. not an active record obj? If so, could you rename it to something like `orders_package_id` just so it's clear. Job params should only be simple ruby objects as they need to be serialized in ActiveJob.

app/models/orders_package.rb
```diff
@@ -76,10 +77,11 @@ def self.add_partially_designated_item(package)
   end
 
   private
-  def recalculate_quantity
+  def recalculate_quantity(operation)
     update_designation_of_package
     package = Package.find_by_id(package_id)
     package.update_in_stock_quantity(get_total_quantity)
+    StockitSyncOrdersPackageJob.perform_now(package.id, self, operation)
```
steveyken: Can't pass self here - must be self.id and the corresponding job must look up the OrderPackage from the database as this action is asynchronous and will happen sometime in the future.

app/models/orders_package.rb
```diff
@@ -104,4 +106,8 @@ def get_total_quantity
     end
     total_quantity
   end
+
+  def destroy_stockit_record(operation)
+    StockitSyncOrdersPackageJob.perform_now(package.id, self, operation)
```
steveyken: same comment re passing `self`

app/services/stockit/orders_package_sync.rb
```diff
@@ -0,0 +1,105 @@
...
+
+    def delete
+      url = url_for("/api/v1/items/destroy")
+      put(url, { gc_orders_package_id: @orders_package.id })
+    end
+
+    private
+
+    def add_stockit_prefix(inventory_number)
+      return "X#{inventory_number}" if !!((inventory_number || "")[0..0] =~ /[0-9]/)
```
steveyken: Could we define a global constant for stockit prefix "X" so we can refer to it elsewhere too?

app/services/stockit/orders_package_sync.rb
steveyken: I'd not come across `presence` before - this is really nice. I can see myself using it a lot in future. http://apidock.com/rails/Object/presence

namrataukirde: Yes @steveyken , Its very useful in many of cases  :)

config/application.rb
```diff
@@ -35,11 +35,11 @@ class Application < Rails::Application
...
-    config.lograge.enabled = true
-    config.lograge.enabled = Lograge::Formatters::Json.new
-    config.log_formatter = proc do |severity, datetime, progname, msg|
-      "time=\"#{datetime.iso8601(3)}\" level=\"#{severity}\" pid=\"#{Process.pid}\" #{msg.to_s.gsub('"', "'")}\n"
-    end
+    # config.lograge.enabled = true
+    # config.lograge.enabled = Lograge::Formatters::Json.new
+    # config.log_formatter = proc do |severity, datetime, progname, msg|
+    #   "time=\"#{datetime.iso8601(3)}\" level=\"#{severity}\" pid=\"#{Process.pid}\" #{msg.to_s.gsub('"', "'")}\n"
+    # end
```
steveyken: @Darshan-Chauhan causing a problem or just hard to read in dev mode? Feel free to move this into production and staging environments, if so.

Darshan-Chauhan: Sorry @steveyken , pushed wrong file. But yeah it's a hard read. But we can work with comments in dev mode. Do we want to push it to staging and prod also?
If so then I'll keep this otherwise I'll revert this commit.

Let me know

steveyken: I'd like to keep the log formatting in staging and production using the formatter above as it feeds into our log reporting software. I don't mind if you override it for development.

Darshan-Chauhan: Sure @steveyken , Will revert it.

PR #263, opened 2017-01-24
Gcw 1405 create a demo data suite

trojanh:


lib/tasks/demo.rake
```diff
@@ -0,0 +1,143 @@
...
+      create_contacts count_number
+      create_organizations count_number
+
+
+    end
+
+    def create_users count_number
+      puts "User:\t\tCreating #{count_number} user accounts (donors, reviewers, supervisors) "
+      #Create 10 user accounts (donors, reviewers, supervisors)
+      count=count_number
```
namrataukirde: This is not needed. `count_number ` can be directly used.

spec/factories/users.rb
```diff
@@ -13,6 +13,10 @@
 
     association :image
 
+    trait :mobile_number do
+       mobile { "+8525"+Random.rand(1000000).to_s.rjust(7,'0') }
+    end
```
steveyken: You can just use `generate(:mobile)` instead. See spec/factories/sequences.rb

spec/factories/packages.rb
```diff
@@ -36,7 +37,7 @@
     trait :received do
       state "received"
       received_at { Time.now }
-      inventory_number "123456"
+      inventory_number {rand(1000000).to_s.rjust(6,'0')}
```
steveyken: Can you turn this into a sequence? - see spec/factories/sequences.rb Then it can be re-used above.

spec/factories/orders.rb
```diff
@@ -1,11 +1,27 @@
...
-    code "MyString"
-    detail_type "MyString"
-    detail_id 1
-    stockit_contact_id 1
-    stockit_organisation_id 1
-    stockit_id 1
+    status   ["draft", "submitted", "processing", "closed", "cancelled"].sample
+    state   ["draft", "submitted", "processing", "closed", "cancelled"].sample
+
+    code            "#{(65 + Random.rand(26)).chr}#{Random.rand(5)}"
```
steveyken: Can you turn this into a sequence? - see spec/factories/sequences.rb

spec/factories/orders.rb
```diff
@@ -1,11 +1,27 @@
...
+    detail_type     ["CarryOut", "Shipment", "StockitLocalOrder", "Goodcity"].sample
+
+    description     FFaker::Lorem.sentence
+    purpose_description FFaker::Lorem.sentence
+    stockit_id      1
```
steveyken: Perhaps this can be null rather than 1 as 1 indicates there is a stockit item corresponding to it but really there isn't.

spec/factories/orders.rb
```diff
@@ -1,11 +1,27 @@
...
   factory :order do
-    status "MyString"
-    code "MyString"
-    detail_type "MyString"
-    detail_id 1
-    stockit_contact_id 1
-    stockit_organisation_id 1
-    stockit_id 1
+    status   ["draft", "submitted", "processing", "closed", "cancelled"].sample
+    state   ["draft", "submitted", "processing", "closed", "cancelled"].sample
```
steveyken: Why do we have two status/state values? @Darshan-Chauhan @namrataukirde

spec/factories/offers.rb
```diff
@@ -71,6 +71,19 @@
...
     end
 
+    trait :with_messages_body do
+      message=["Thank you for this", "What an excellent thing.", "Thanks for your reply", "We thank you for choosing to donate.", "The item is in good condition"].sample
+      transient do
+        messages_count 1
+      end
+
+      after(:create) do |offer, evaluator|
+        create_list(:message, evaluator.messages_count, body: message ,sender: offer.created_by, offer: offer)
```
steveyken: Good work @trojanh

lib/tasks/demo.rake
```diff
@@ -0,0 +1,143 @@
...
+namespace :demo do
+  unless ENV['LIVE'] == "true"
+    task load: :environment do
+
+      ARGV.each { |a| task a.to_sym do ; end }
+
+      #specify number of test cases to produce
+      count_number= Integer(ARGV[1]) rescue 0 >0 ? ARGV[1].to_i : 10
+      puts "This will generate #{count_number} record of Users, Offers, Packages, OrdersPackages, Orders, Contacts & Organisations"
+      create_users count_number
```
steveyken: Can you use () here for function calls - ruby 2 prefers things this way :)

spec/factories/offers.rb
trojanh: Thanks @steveyken , I have made the above changes. Please review it .

PR #264, opened 2017-01-31
Undispatch from gc

namrataukirde:


spec/models/package_spec.rb
```diff
@@ -169,20 +169,269 @@
...
+      sibling_package = create :package, :with_set_item, :package_with_locations, item: package.item
       package.dispatch_stockit_item
       package.save
       expect(package.set_item_id).to be_nil
       expect(sibling_package.reload.set_item_id).to be_nil
     end
   end
+
+  describe '#add_location' do
+    it 'adds location to package if not added' do
```
steveyken: use `let` syntax here to clean things up a bit and make it simpler to add new specs in the future...

        describe "#add_location" do
        
        let(:location) { create(:location) }
        let(:package) { create(:package) }

        describe "adds location to package if not added"
            it do
            package.add_location(location.id)
            expect(package.locations).to include(location)
            end
        end

        describe 'unless location already exists' do
            let(:package) { create(:package, :package_with_locations) }
            let(:location) { package.packages_locations.first.location }
            it do
            expect(package.add_location(location.id)).to be_nil
            expect(package.locations).to include(location)
            end
        end
        end

spec/models/package_spec.rb
```diff
@@ -169,20 +169,269 @@
...
+    end
+  end
+
+  describe '#update_or_create_qty_moved_to_location' do
+    let!(:package) { create :package }
+    let!(:location) { create :location }
+
+    it 'creates associated packages_location record if we do not have packages_location record with provided location_id' do
+      expect{
+        package.update_or_create_qty_moved_to_location(location.id, 10)
```
steveyken: Will this fail if package doesn't have at least quantity 10? Perhaps it could be `package.quantity` or similar instead?

spec/models/package_spec.rb
```diff
@@ -169,20 +169,269 @@
...
+
+    it 'updates existing packages_location quantity to with new quantity which is addition of qty to move and packages_location quantity' do
+      packages_location = create :packages_location, quantity: 10, location: location, package: package
+      package.update_or_create_qty_moved_to_location(location.id, 10)
+      expect(packages_location.reload.quantity).to eq 20
+    end
+  end
+
+  describe '#update_existing_package_location_qty' do
+    let!(:package) { create :package, received_quantity: 140, quantity: 140 }
```
steveyken: Do you need `let!` here... what about just `let`?

spec/models/package_spec.rb
```diff
@@ -169,20 +169,269 @@
...
+  describe '#move_partial_quantity' do
+    let!(:package) { create :package }
+    let!(:location) { create :location }
+    let!(:location_1) { create :location }
+    let!(:packages_location) { create :packages_location, quantity: 12, package: package, location: location_1 }
+
+    context 'moving some qty to location for which associated packages_location do not exist' do
+      it 'subtract quantity to move from packages_location record(current location)' do
+        quantity_to_move = 5
+        package_qty_changes = [{packages_location_id: packages_location.id, package_id: package.id, new_qty: quantity_to_move}].to_json
```
steveyken: see previous note about moving JSON.parse... you won't need to do this step then

PR #265, opened 2017-02-02
Gcw 1373 order states

trojanh:


app/models/order.rb
```diff
@@ -29,6 +30,24 @@ class Order < ActiveRecord::Base
 
   scope :active_orders, -> { where('status NOT IN (?)', INACTIVE_STATUS) }
 
+  def update_packages
+    if(detail_type == "GoodCity")
+      orders_packages.each do |orders_package|
+        orders_package.package.update_allow_web_publish
+        orders_package.state = 'designated'
```
Darshan-Chauhan: Can we move this to instance method as orders_package record is available is?

app/models/order.rb
```diff
@@ -29,6 +30,24 @@ class Order < ActiveRecord::Base
...
+        orders_package.save
+      end
+    end
+  end
+
+  def update_packages_quantity
+    if(state == "draft" && detail_type == "GoodCity")
+      orders_packages.each do |orders_package|
+        orders_package.update(quantity: orders_package.package.quantity)
```
Darshan-Chauhan: ^

Darshan-Chauhan:
👍

namrataukirde:
👍

steveyken:
Was about to say I would really love to have some tests for this, then I saw your other branch. Are you able to merge that soon too?

trojanh:
I started writing specs yesterday but then switched to rebasing my PRs. Will work on it today and will probably finish it.

app/models/order.rb
```diff
@@ -29,6 +30,22 @@ class Order < ActiveRecord::Base
 
   scope :active_orders, -> { where('status NOT IN (?)', INACTIVE_STATUS) }
 
+  def update_packages
```
steveyken: It would be good if this function name reflected more about what it is doing. e.g.  `designate_orders_packages`

app/models/order.rb
```diff
@@ -47,6 +64,10 @@ def set_initial_state
     before_transition on: :submit do |order|
       order.add_to_stockit
     end
+
+    after_transition on: :submit do |order|
+      order.update_packages
```
steveyken: I would put the guard on this line, so

    order.update_packages if (detail_type == "GoodCity")

That way, the function `update_packages` (renamed to `designate_orders_packages`) would do exactly that in every case it is called. It would be surprising to use the function later on another detail_type and find that it doesn't designate so this would fix that case

app/models/package.rb
```diff
@@ -111,6 +111,10 @@ def add_location(location_id)
     end
   end
 
+  def update_allow_web_publish
```
steveyken: I'm a bit confused here... function says `update_allow_web_publish` but we are setting the attribute to false?

app/models/orders_package.rb
```diff
@@ -46,6 +47,24 @@ def undispatch_orders_package
...
+  def update_state_to_designated
+    package.update_allow_web_publish
+    update(state: 'designated')
+  end
+
+  def update_quantity
+    update(quantity: package.quantity)
+  end
+
+  def update_state_to_designated
```
steveyken: Duplicated function above

app/models/order.rb
```diff
@@ -29,6 +30,22 @@ class Order < ActiveRecord::Base
...
+  def update_packages
+    if(detail_type == "GoodCity")
+      orders_packages.each do |orders_package|
+        orders_package.update_state_to_designated
+      end
+    end
+  end
+
+  def update_packages_quantity
+    if(state == "draft" && detail_type == "GoodCity")
```
steveyken: I would suggest moving this guard to the `after_create` hook that calls this function. That way, things are clearer and we are free to call `update_packages_quantity` at other times too.

app/models/orders_package.rb
trojanh: It happened because of rebase. 
Will sort it out. Thanks for pointing.

app/models/package.rb
trojanh: actually the attribute name is allow_web_publish and i was following rails convention. I had thought of such ambiguity, anyway i'll rename it to make more sense.

app/models/order.rb
trojanh: Sounds great. Will do that.

PR #270, opened 2017-02-17
Gcw 1373 order states

trojanh:


app/models/order.rb
```diff
@@ -15,7 +15,7 @@ class Order < ActiveRecord::Base
   has_and_belongs_to_many :cart_packages, class_name: 'Package'
   has_one :order_transport
 
-  after_create :update_packages_quantity
+  after_create :update_orders_packages_quantity, if: :if_state_is_draft_and_detail_type_is_quantity
```
Darshan-Chauhan: can we rename if condition to **state_is_draft_and_detail_type_is_quantity?**

lib/tasks/demo.rake
```diff
@@ -100,12 +96,15 @@ namespace :demo do
...
         @order=create_single_order
         @package=FactoryGirl.create(:package, :with_item)
-        FactoryGirl.create(:orders_package,
+        @orders_package=FactoryGirl.build(:orders_package,
                               package: @package,
                               order: @order,
-                              quantity: Random.rand(15),
+                              quantity: @package.quantity,
                               updated_by: @updated_by
                             )
```
namrataukirde: check indentation here ?

spec/models/orders_package_spec.rb
```diff
@@ -13,4 +13,20 @@
...
+      @orders_package = create :orders_package, :with_state_requested
+      @orders_package.update_state_to_designated
+      expect(@orders_package.state).to match("designated")
+    end
+  end
+
+  describe "update_quantity" do
+    it "Updates orders_packages quantity" do
+      @orders_package = create :orders_package, :with_state_requested
+      @orders_package.update_quantity
```
namrataukirde: nor expectatiion is checked here..

app/models/order.rb
namrataukirde: state_is_draft_and_detail_type_is_goodcity?

PR #272, opened 2017-03-09
Gcw 1463 populate organisation table from json data

steveyken:
@trojanh I have some time to review so doing this now


lib/tasks/populate_organisation.rake
```diff
@@ -0,0 +1,44 @@
...
+    organisation_type = OrganisationType.find_by_name_en(organisation_type_name) || OrganisationType.create(name_en: organisation_type_name ,name_zh_tw: organisation_type_name ,category_en: organisation_type_name  , category_zh_tw: organisation_type_name)
+
+    #set country
+    country_name = "China - Hong Kong (Special Administrative Region)"
+    country_code = "HKG"
+    country_region = "Asia"
+
+    #for job
+    stockit_id = StockitCountryAddJob.perform_now(country_name, country_name, country_code, country_region)
+    country = Country.find_by(stockit_id: stockit_id)|| Country.create(name_en: country_name ,name_zh_tw: country_name)
```
steveyken: Can you remove StockitCountryAddJob and replace with

    COUNTRY_NAME_EN = "China - Hong Kong (Special Administrative Region)"
    country = Country.find_by_name_en(COUNTRY_NAME_EN)

lib/tasks/populate_organisation.rake
```diff
@@ -0,0 +1,44 @@
...
+namespace :populate_organisation do
+  task organisation:  :environment do
+    organisation_fields_mapping = {
+      "name_en" => "name_en",
+      "name_zh_tw" => "name_zh",
+      "website" => "url",
+    }
+    #set organisation_type
+    organisation_type_name = "NGO"
+    organisation_type = OrganisationType.find_by_name_en(organisation_type_name) || OrganisationType.create(name_en: organisation_type_name ,name_zh_tw: organisation_type_name ,category_en: organisation_type_name  , category_zh_tw: organisation_type_name)
```
steveyken: When creating OrganisationType, you can leave category_en/zh_tw and name_zh blank. There is no validation for them.

lib/tasks/populate_organisation.rake
```diff
@@ -0,0 +1,44 @@
...
+    country_name = "China - Hong Kong (Special Administrative Region)"
+    country_code = "HKG"
+    country_region = "Asia"
+
+    #for job
+    stockit_id = StockitCountryAddJob.perform_now(country_name, country_name, country_code, country_region)
+    country = Country.find_by(stockit_id: stockit_id)|| Country.create(name_en: country_name ,name_zh_tw: country_name)
+
+    begin
+      file = Nestful.get("https://goodcitystorage.blob.core.windows.net/public/s88-orgs.json").response.body
```
steveyken: Can you extract this URL out into a constant and put at the top of the task along with COUNTRY_NAME_EN as well?

lib/tasks/populate_organisation.rake
```diff
@@ -0,0 +1,44 @@
...
+    rescue Exception => e
+      Airbrake.notify(e, error_class: "populate_organisation", error_message: "Organisation File Error")
```
steveyken: Good!

lib/tasks/populate_organisation.rake
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    # file = File.read('app/assets/organisation.json')
+    if (file.present?)
+      JSON.parse(file).each do |data|
+        organisation_fields_mapping =  organisation_fields_mapping.keep_if { |k, v| data.key? v }
+        organisation = Organisation.find_by(registration: data['org_id']) ||
+          Organisation.new(registration: data['org_id'], organisation_type: organisation_type, country: country)
+          organisation_fields_mapping.each do |organisation_column, data_key|
```
steveyken: This is good. Can you just fix the indentation slightly so it's easier to read? Thanks

app/jobs/stockit_country_add_job.rb
```diff
@@ -0,0 +1,14 @@
+class StockitCountryAddJob < ActiveJob::Base
```
steveyken: as discussed, we can delete this job

app/services/stockit/country_sync.rb
```diff
@@ -5,14 +5,35 @@ class CountrySync
 
     include Stockit::Base
```
steveyken: Can we revert these changes as well?

steveyken:
Thanks @trojanh . As we discussed, we'll take a slightly different approach so a few comments here to set that up. Could you also look into writing some tests for the rake task once the changes have been made .. it might be easier to extract the code out into a lib/ class so it's easier to test... we can talk more about this when you're done with the other changes. Thanks. Overall, good job!

trojanh:
Thanks @steveyken for reviewing the PR. I'll make the required changes.

lib/tasks/populate_organisation.rake
namrataukirde: I feel adding constant to Constant.rb will be better. Let me know what you think @steveyken

steveyken: yes that would make sense too but perhaps if we extract all of this out into a class then the constants can be class constants.

namrataukirde: Yes Right 👍

lib/goodcity/organisation_populator.rb
```diff
@@ -0,0 +1,57 @@
+module Goodcity
```
steveyken: @trojanh I'd like to retire the lib/classes folder and switch to using the lib/goodcity namespace so I've moved this class here.

lib/tasks/organisations.rake
```diff
@@ -1,5 +1,12 @@
+require "goodcity/organisation_populator"
```
steveyken: @trojanh There was already an organisations.rake file so it made sense to put the rake task in there.

lib/goodcity/organisation_populator.rb
```diff
@@ -22,7 +22,7 @@ def initialize
     end
 
 
-    def populate_organisation_db
+    def run
```
steveyken: @trojanh I think the class name made it clear what the job is so no need to repeat in the name

lib/goodcity/organisation_populator.rb
trojanh: Yeah sure, @steveyken . Thanks.

lib/goodcity/organisation_populator.rb
```diff
@@ -12,7 +12,6 @@ class OrganisationPopulator
 
     def initialize
       @organisation_type = get_organisation_type
-      @country           = Country.find_by_name_en(COUNTRY_NAME_EN)
```
steveyken: @trojanh In general, the initialize function shouldn't be used to do any 'work' e.g. database calls, http lookups, file loads etc. It should be just enough code to return a new class that hasn't done anything yet but is in a ready and consistent state to be called. E.g. imagine if we added another top level method to count the number of organisations that exist, we wouldn't need to find the country or load the URL just to do this so it would be extra unnecessary work when calling initialize.

lib/goodcity/organisation_populator.rb
```diff
@@ -38,8 +37,9 @@ def run
 
 
     private
-    def get_organisation_type
-      OrganisationType.find_by_name_en(ORGANISATION_TYPE_NAME) || OrganisationType.create(name_en: ORGANISATION_TYPE_NAME)
+
+    def organisation_type
+      @organisation_type ||= OrganisationType.find_or_create_by(name_en: ORGANISATION_TYPE_NAME)
```
steveyken: @trojanh It looks like you were looking up the organisation_type in the initializer in order to only do it once and not repeat for each loop iteration. A useful trick here is to use the `||=` operator to memoize the lookup. The first time the function is called, it will call find_or_create_by, but the subsequent times, it will just return the previous (non-nil) value and is therefore effectively cached. This has the advantage of extracting another DB call out of the initializer.

lib/goodcity/organisation_populator.rb
```diff
@@ -10,31 +10,29 @@ class OrganisationPopulator
       "website" => "url",
     }
 
-    def initialize
-      begin
-        @file = Nestful.get(URL).response.body
```
steveyken: @trojanh now that we've extracted the URL download out of the initializer there is nothing left to do so it can be removed. Since no initializer is required, we can create a simple `self.run` convenience method so that it's shorter function call i.e. `Goodcity::OrganisationPopulator.run` instead of `Goodcity::OrganisationPopulator.new.run`

Also, the airbrake exceptions should be caught inside the rake task already. If not, then we should add it at the global rake level (for every task) not just this one.

spec/lib/goodcity/organisation_populator_spec.rb
```diff
@@ -4,10 +4,12 @@
 describe Goodcity::OrganisationPopulator do
   let(:organisation_populator) { Goodcity::OrganisationPopulator.new }
   let(:file)                   { File.read("#{Rails.root}/spec/fixtures/organisation.json")}
+  let!(:country)               { FactoryGirl.create(:country, name_en: "China - Hong Kong (Special Administrative Region)") }
```
steveyken: @trojanh in general, it's better to use factories to create data and also `let` syntax. (although let! vs before is debatable!)

lib/goodcity/organisation_populator.rb
trojanh: will keep that in mind henceforth, nice explanation @steveyken

spec/lib/goodcity/organisation_populator_spec.rb
```diff
@@ -29,22 +29,23 @@
   end
 
   context "update organisation" do
+    let(:registration_id_one) { "91/09657"}
```
steveyken: @trojanh Use `let` syntax instead of `@instance` variables. `let` has delayed evaluation so it's not run until it's called.

trojanh: ok, sure

spec/lib/goodcity/organisation_populator_spec.rb
```diff
@@ -44,12 +44,13 @@
       end
 
       it ":update the existing records" do
+        organisation_populator.run
```
steveyken: @trojanh before this change, we were effectively testing that nothing changes when nothing is called! Now we're testing that updates are applied when the populator is run.

steveyken:
@trojanh I've commented on all the commits I've changed so you can get a feel for the sorts of things that would be useful to pick up. Have a read and follow up with me if there are any questions. Otherwise, this is now ready to merge.

trojanh:
I read all your comments and got many things to learn. Thanks for reviewing @steveyken .

lib/goodcity/organisation_populator.rb
trojanh: I was wondering what name to use, so ended up with that. Anyway run sounds better.

PR #273, opened 2017-03-21
Renamed reference_to_orders_package to orders_package_id

steveyken:
Strange to have a reference called reference... using the standard rails naming convention so that the belongs_to and hasbtm methods work normally.

Any reason why this is a bad idea?


namrataukirde:
We dont have any relationship between packages_location and orders_packages. Thus we gave such name. If it had relationship then we are fine with rails convention.

steveyken:
Going to close this as it's old and we've moved on. Don't merge.

PR #276, opened 2017-03-30
GCW-1505 mask quantity packages

Darshan-Chauhan:


app/controllers/api/v1/packages_controller.rb
```diff
@@ -124,7 +124,7 @@ def search_stockit_items
         records = records.search(params['searchText'], params["itemId"]).page(params["page"]).per(params["per_page"])
         pages = records.total_pages
       end
-
+      records = records.where("received_quantity = 1") if is_stock_app
```
namrataukirde: Could you please check and add this to method in which we are actually fetching records from db ?

Darshan-Chauhan: Yes, Just added it...

namrataukirde: 👍

PR #277, opened 2017-04-05
Gcw 1489 mark missing should clear packages loactions record

trojanh:


app/models/package.rb
```diff
@@ -111,6 +112,10 @@ def add_location(location_id)
     end
   end
 
+  def delete_associated_packages_locations
+    self.packages_locations.destroy_all
```
namrataukirde: What if we don's use self here ?

trojanh: I think it will work. Should I remove it?

namrataukirde: Yes, It will work but I think we don't need self here.

PR #279, opened 2017-04-10
Update demo data to maintian referencial integrity

trojanh:


spec/factories/packages.rb
```diff
@@ -1,12 +1,12 @@
...
   factory :package do
-    quantity              { rand(5) + 1 }
+    quantity              { rand(10) + 1 }
     length                { rand(199) + 1 }
     width                 { rand(199) + 1 }
     height                { rand(199) + 1 }
     notes                 { FFaker::Lorem.paragraph }
     state                 'expecting'
-    received_quantity     1
+    received_quantity     10
```
steveyken: doesn't this need to be similar to quantity? Will it break if received_quantity > quantity?

trojanh: 'received_quantity' is the one that we get from Offers. We use that initially as our 'quantity' but 'quantity' keeps changing depending on factors like designation and  dispatch. Maximum 'quantity' any package can have is 'received_quantity'. So it can be greater than 'quantity' or even equal but not less because that would be like saying 'I had received 'x' quantity from an offer but now I have 'x+10' quantity' which isn't right.

So it should work.

PR #283, opened 2017-04-19
Gcw 1405 correct demo data to sync with stockit and follow lifecycle

trojanh:


lib/tasks/demo.rake
```diff
@@ -14,84 +14,151 @@ namespace :demo do
...
-        )
-        if(@orders_package.state == "designated")
-          @package.order_id = @order_id
+
+    def create_designated_packages
+      order = create_single_order
+      order.save
+      offer = create_recieved_offer
+      orders_packages_ids = []
+      offer.reload
```
namrataukirde: Instead of this we can directly write offer.reload.items.

lib/tasks/demo.rake
```diff
@@ -14,84 +14,150 @@ namespace :demo do
...
+
+    def create_receiving_offer
+      offer = create_reviewed_offer
+      offer.start_receiving
+      offer
+    end
+
+    def create_recieved_offer
+      offer = create_receiving_offer
+      offer = inventory_offer_packages (offer)
```
namrataukirde: remove space before ()

lib/tasks/demo.rake
```diff
@@ -14,84 +14,150 @@ namespace :demo do
...
+      offer.receive
+      offer
+    end
+
+    def inventory_offer_packages (offer)
+      offer.reload.items.each do |item|
+        item.accept
+        item.packages.each do |package|
+          loc = Location.all.to_a.sample.id
```
namrataukirde: Name should be proper. rename it `location_id`

namrataukirde: Use `Location.pluck(:id).sample` Instead of `Location.all.to_a.sample.id`

lib/tasks/demo.rake
```diff
@@ -14,84 +14,150 @@ namespace :demo do
...
+          package.update(inventory_number: InventoryNumber.available_code, allow_web_publish: true, location_id: loc)
+          package.build_or_create_packages_location(loc, 'create')
+
```
namrataukirde: remove unnecessary line.

lib/tasks/demo.rake
```diff
@@ -14,84 +14,150 @@ namespace :demo do
...
+          package.mark_received
+        end
       end
+      offer.update(delivered_by: ['Gogovan','Crossroads truck','Dropped off'].sample)
```
namrataukirde: ['Gogovan','Crossroads truck','Dropped off'] can be moved to constants.

lib/tasks/demo.rake
```diff
@@ -14,84 +14,150 @@ namespace :demo do
...
+
+    def create_submitted_offer
+      offer = FactoryGirl.create(:offer, :with_demo_items, :with_messages, created_by: donor)
+      offer.submit
+      offer
+    end
+
+    def create_reviewing_offer
+      offer = create_submitted_offer
+      reviewer = FactoryGirl.create(:user, :reviewer)
```
namrataukirde: this can be moved to method as it is used multiple times.

lib/tasks/demo.rake
```diff
@@ -14,84 +14,150 @@ namespace :demo do
...
-          order: @order,
-          quantity: @package.quantity,
-          updated_by: @updated_by
-        )
-        if(@orders_package.state == "designated")
-          @package.order_id = @order_id
+
+    def create_designated_packages
+      order = create_single_order
+      order.save
```
namrataukirde: Do we need order.save here ?

lib/tasks/demo.rake
```diff
@@ -14,84 +14,144 @@ namespace :demo do
...
+      offer = inventory_offer_packages(offer)
+      offer.receive
+      offer
+    end
+
+    def inventory_offer_packages(offer)
+      offer.reload.items.each do |item|
+        item.accept
+        item.packages.each do |package|
+          loc = Location.all.to_a.sample.id
```
namrataukirde: location_id = Location.pluck(:id).sample

lib/tasks/demo.rake
```diff
@@ -14,84 +14,144 @@ namespace :demo do
...
+    end
+
+    def create_closed_offer
+      offer = create_reviewed_offer
+      offer.mark_unwanted
+      offer
+    end
+
+    def create_receiving_offer
+      offer = create_reviewed_offer
```
steveyken: for methods with this similar pattern, you could look at using the `tap` function in ruby. It calls a method on an object and returns the method itself... something like this:

    def create_receiving_offer
      create_reviewed_offer.tap(&:start_receiving)
    end

Read http://ruby-doc.org/core-2.4.1/Object.html#method-i-tap and http://seejohncode.com/2012/01/02/ruby-tap-that/ (particularly this last one)

lib/tasks/demo.rake
```diff
@@ -14,84 +14,144 @@ namespace :demo do
...
+      offer.reload.items.each do |item|
+        item.accept
+        item.packages.each do |package|
+          loc = Location.pluck(:id).sample
+          package.update(inventory_number: InventoryNumber.available_code, allow_web_publish: true, location_id: loc)
+          package.build_or_create_packages_location(loc, 'create')
+          package.mark_received
+        end
       end
+      offer.update(delivered_by: DELIVERED_BY.sample)
```
steveyken: could you remove this from constants and add it as a factory girl sequence instead?

lib/tasks/demo.rake
```diff
@@ -107,17 +167,22 @@ namespace :demo do
       end
     end
 
+    def reviewer
```
steveyken: can we be sure we already have reviewers? or should this be

    User.where(permission_id: 3).first || FactoryGirl.create(:reviewer)

trojanh: It is created when we run seed rake mentioned in the demo rake but I think its safe to have `User.where(permission_id: 3).sample||FactoryGirl.create(:user, :reviewer)` rather than just `User.where(permission_id: 3).sample`. 
Thanks @steveyken for pointing this.

lib/tasks/demo.rake
trojanh: Cool. I was wondering if we could use 'create_reviewing_offer.finish_review' but it returns boolean value as 'finish_review' is a state event. But 'create_reviewing_offer.tap(&:finish_review)' works flawlessly and return exactly what i wanted. 
Thanks @steveyken .

lib/tasks/demo.rake
```diff
@@ -14,84 +14,129 @@ namespace :demo do
...
       end
+      offer.update(delivered_by: FactoryGirl.generate(:delivered_by))
+      offer
     end
 
     def create_single_order
-        @organisation = FactoryGirl.create(:organisation, organisation_type_id: OrganisationType.find_by_id(Random.rand(3)))
-        @processor = FactoryGirl.create(:user, :reviewer)
-        FactoryGirl.create(:order, :with_created_by, processed_by: @processor, organisation: @organisation)
+      organisation = FactoryGirl.create(:organisation, organisation_type_id: OrganisationType.find_by_id(Random.rand(3)))
```
namrataukirde: OrganisationType.find_by_id(Random.rand(3)) can be moved to method as it is used twice.

lib/tasks/demo.rake
```diff
@@ -14,84 +14,129 @@ namespace :demo do
...
+
+    def create_recieved_offer
+      inventory_offer_packages(create_receiving_offer).tap(&:receive)
+    end
+
+    def inventory_offer_packages(offer)
+      offer.reload.items.each do |item|
+        item.accept
+        item.packages.each do |package|
+          loc = Location.pluck(:id).sample
```
namrataukirde: Use proper name `location_id`

lib/tasks/demo.rake
```diff
@@ -14,84 +14,129 @@ namespace :demo do
...
+    end
+
+    def create_reviewing_offer
+      User.current_user = reviewer
+      create_submitted_offer.tap(&:start_review)
+    end
+
+    def create_reviewed_offer
+      offer = create_reviewing_offer
+      offer.update(reviewed_by: reviewer)
```
namrataukirde: do we need update method here? Just update_column will not work ?

lib/tasks/demo.rake
```diff
@@ -107,17 +152,22 @@ namespace :demo do
       end
     end
 
+    def reviewer
+      User.where(permission_id: 3).sample
```
namrataukirde: Add || condition as suggested by Steve.

lib/tasks/demo.rake
```diff
@@ -14,84 +14,129 @@ namespace :demo do
...
-        create_single_order
+    def create_dispatched_packages
+      orders_packages_ids = create_designated_packages
+      orders_packages_ids.each do |orders_pkg|
+        orders_package = OrdersPackage.find(orders_pkg)
+        pkg = orders_package.package
+        orders_package.dispatch_orders_package
+        pkg.dispatch_stockit_item(orders_package)
       end
+      orders_packages_ids
```
namrataukirde: Do we need to return `orders_packages_ids` from this method ?

lib/tasks/demo.rake
```diff
@@ -14,84 +14,129 @@ namespace :demo do
...
+    def create_recieved_offer
+      inventory_offer_packages(create_receiving_offer).tap(&:receive)
+    end
+
+    def inventory_offer_packages(offer)
+      offer.reload.items.each do |item|
+        item.accept
+        item.packages.each do |package|
+          loc = Location.pluck(:id).sample
+          package.update(inventory_number: InventoryNumber.available_code, allow_web_publish: true, location_id: loc)
```
namrataukirde: do we need update method here? Just update_column will not work ?

lib/tasks/demo.rake
```diff
@@ -14,84 +14,129 @@ namespace :demo do
...
+          package.build_or_create_packages_location(loc, 'create')
+          package.mark_received
+        end
       end
+      offer.update(delivered_by: FactoryGirl.generate(:delivered_by))
```
namrataukirde: do we need update method here? Just update_column will not work ?

lib/tasks/demo.rake
trojanh: it is not used but i have returned it for uniformity and future needs.

lib/tasks/demo.rake
trojanh: for offer.update_column will(reviewed_by: reviewer) 
[Validations are skipped.
Callbacks are skipped.
updated_at/updated_on are not updated.](https://cbabhusal.wordpress.com/2014/10/08/different-between-update-update_columns-update_column-update_attributes-assign_attributes/)

So i don't think we might want to use it.

namrataukirde: Yes, I know that's why i asked. If we really need to execute validations, callbacks an all in that case we should use update. otherwise if we just want to change value and dont need callbacks, validations then we can directly use update_column which will prevent unnecessary processes.  In this case if we require callbacks then keep it as it is.

lib/tasks/demo.rake
namrataukirde: If we don't need it right now then please remove it.

PR #285, opened 2017-04-21
GCW-1520-dispatch-update-for-stockit

namrataukirde:


Darshan-Chauhan:
👍

app/services/stockit/item_sync.rb
```diff
@@ -71,6 +79,20 @@ def delete
       end
     end
 
+    def dispatch
```
steveyken: yes! I noticed these methods were missing yesterday and pleased to see an endpoint for this. On the Stockit side, it should be calling `item.dispatch` which will handle all the dispatch logic internally.

PR #288, opened 2017-04-26
Gcw 1508 update rake to include logs in a rake log.txt file

trojanh:


trojanh:
Sorry for that @steveyken, I was not able to see the old branch on my machine so i could not update that branch.

PR #297, opened 2017-05-09
Refactoring to improve code climate

trojanh:
followed DRY principle for some of the repeating codes. Should help improve the code climate.


app/controllers/api/v1/schedules_controller.rb
```diff
@@ -37,12 +37,7 @@ def availableTimeSlots
     api :POST, "/v1/schedules", "Make a booking for the pick up"
     param_group :schedule
     def create
-      @schedule.attributes = schedule_params
-      if @schedule.save
-        render json: @schedule, serializer: serializer, status: 201
-      else
-        render json: @schedule.errors.to_json, status: 422
-      end
+      assign_params_and_render_object(@schedule, serializer, schedule_params)
```
steveyken: the `assign_params...` method uses the `cached_json` model but that isn't here in schedule controller and might produce unexpected effects.

The `data_object = data_object.find( params[:ids].split(",") ) if params[:ids].present?` line in `assign_params...` function is not something that happens in the `schedules#create` controller - possible unintended consequences?

app/controllers/api/v1/api_controller.rb
```diff
@@ -1,5 +1,5 @@
 module Api::V1
-  class ApiController < ApplicationController
+  class ApiController <  Api::V1::UtilityController
```
steveyken: Rather than inherit from `UtilityController` you can put the functions directly into `ApiController` and delete UtilityController. The functions are specific to the API so that's ok.

app/controllers/api/v1/holidays_controller.rb
```diff
@@ -32,12 +32,7 @@ def available_dates
     api :POST, '/v1/holidays', "Create holiday"
     param_group :holiday
     def create
-      @holiday = Holiday.new(holiday_params)
-      if @holiday.save
-        render json: @holiday, serializer: serializer, status: 201
-      else
-        render json: @holiday.errors.to_json, status: 422
-      end
+      assign_params_and_render_object(Holiday.new, serializer, holiday_params)
```
steveyken: As per CanCanCan comments, no need for Holiday.new here.

CanCanCan is effectively doing the following for you:

    @holiday = Holiday.new
    @holiday.attributes = holiday_params

The source for that is here - https://github.com/CanCanCommunity/cancancan/blob/734c70cb824c58e1efb1473401186abeab023c54/lib/cancan/controller_resource.rb start with `load_resource` method and follow it through to the `assign_attributes` method.

app/controllers/api/v1/schedules_controller.rb
trojanh: Hi @steveyken , appreciate your valuable feedback on this PR.

About this specific method call , I am calling following method for schedules#create
```
def assign_params_and_render_object(data_object, serializer, params={})
       data_object.attributes = params
       if data_object.save
         render json: data_object, serializer: serializer, status: 201
       else
         render json: data_object.errors.to_json, status: 422
       end
     end
```
So it doesn't use cached json and  `data_object = data_object.find( params[:ids].split(",") ) if params[:ids].present?` maybe you were talking about following method call which we are not calling from here

```
    def render_object(data_object, model_name, serializer, params)
      if params[:ids].blank?
         render json: model_name.cached_json
         return
       end
       data_object = data_object.find( params[:ids].split(",") ) if params[:ids].present?
       render json: data_object, each_serializer: serializer
     end
   end
```

So if we call `assign_params_and_render_object` method will it be fine or can we have some undesirable consequences?

PR #302, opened 2017-05-11
Health check code.

steveyken:
Use `rake health_checks`

Need more checks to be written now

### DIFF
--- a/lib/goodcity/health_checks/dispatched_packages_order_id_check.rb
+++ b/lib/goodcity/health_checks/dispatched_packages_order_id_check.rb
@@ -1,0 +1,19 @@
+require 'goodcity/health_checks/base'
+
+module Goodcity
+  class HealthChecks
+
+    class DispatchedPackagesOrderIdCheck < Base
+      desc "Dispatched packages should contain an order_id reference."
+      def run
+        ids = Package.where('stockit_sent_on is not null and order_id is null').pluck(:id)
+        if ids.count == 0
+          pass!
+        else
+          fail_with_message!("GoodCity Dispatched Packages with nil sent_on or order_id: #{ids.join(', ')}")
+        end
+      end
+    end
+
+  end
+end
--- a/lib/goodcity/health_checks/package_stockit_id_nil_check.rb
+++ b/lib/goodcity/health_checks/package_stockit_id_nil_check.rb
@@ -1,0 +1,19 @@
+require 'goodcity/health_checks/base'
+
+module Goodcity
+  class HealthChecks
+
+    class PackageStockitIdNilCheck < Base
+      desc "Packages should contain a stockit_id reference."
+      def run
+        ids = Package.where(stockit_id: nil).pluck(:id)
+        if ids.count == 0
+          pass!
+        else
+          fail_with_message!("GoodCity Packages with nil stockit_id: #{ids.join(', ')}")
+        end
+      end
+    end
+
+  end
+end
--- a/spec/lib/goodcity/health_checks/dispatched_packages_order_id_check_spec.rb
+++ b/spec/lib/goodcity/health_checks/dispatched_packages_order_id_check_spec.rb
@@ -1,0 +1,34 @@
+require 'rails_helper' # needed to configure transaction rollback
+require 'goodcity/health_checks/dispatched_packages_order_id_check'
+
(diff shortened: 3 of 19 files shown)


lib/goodcity/health_checks/dispatched_packages_order_id_check.rb
```diff
@@ -0,0 +1,19 @@
+require 'goodcity/health_checks/base'
+
+module Goodcity
+  class HealthChecks
+
+    class DispatchedPackagesOrderIdCheck < Base
+      desc "Dispatched packages should contain an order_id reference."
+      def run
+        ids = Package.where(stockit_sent_on: nil, order_id: nil).pluck(:id)
```
Darshan-Chauhan: @steveyken , this query doesn't check for dispatched packages as **stockit_sent_on: nil and order_id: nil** will only give received packages. 
I think **stockit_sent_on should be checked with not nil** here.

lib/goodcity/health_checks/package_stockit_id_nil_check.rb
```diff
@@ -0,0 +1,19 @@
...
+    class PackageStockitIdNilCheck < Base
+      desc "Packages should contain a stockit_id reference."
+      def run
+        ids = Package.where(stockit_id: nil).pluck(:id)
+        if ids.count == 0
+          pass!
+        else
+          fail_with_message!("GoodCity Packages with nil stockit_id: #{ids.join(', ')}")
+        end
+      end
```
trojanh: these same lines are used multiple times above for `PackageStockitIdNilCheck` `OrdersPackagesOrderIdCheck`, `OrderStockitIdNilCheck` and `LocationStockitIdNilCheck` so maybe we can create a common method for them.

steveyken: Yes I know what you mean. It's a tough call... too much abstraction and it becomes hard to read. Too little and nothing is DRY!

There is a pattern here with some checks where if there are any results found that then is bad and a query resulting in an empty array is good. Perhaps something like:

    query = Package.where(stockit_id: nil)
    fail_if_not_empty(query, failure_message)


    def fail_if_not_empty(query, failure_message = '')
      query.any? ? fail_with_message!(message) : pass!
    end
Could take it a bit further and say 'fail if less than X items' but we don't have that use case yet.

spec/lib/goodcity/health_checks/dispatched_packages_order_id_check_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  context "passes" do
+    it "with stockit_sent_on" do
+      create :package, stockit_sent_on: Date.today, order_id: order.id
+      subject.run
+      expect(subject.passed?).to eql(true)
+    end
+    it "with nil stockit_sent_on" do
+      create :package, stockit_sent_on: nil, order_id: order.id
+      subject.run
+      expect(subject.passed?).to eql(true)
```
steveyken: this should be failing now

lib/goodcity/health_checks/dispatched_packages_order_id_check.rb
steveyken: good catch, thanks for fixing

spec/lib/goodcity/health_checks/dispatched_packages_order_id_check_spec.rb
Darshan-Chauhan: @steveyken it won't fail as it will be just designated package with stockit_sent_on nil so it's fine.

lib/goodcity/health_checks/package_stockit_id_nil_check.rb
trojanh: okay, Thanks @steveyken sounds good. :+1:

steveyken: Decided to back out of these changes for now... the failure message that is passed in required access to the query to list the ids so not possible to defer execution of query till later.

spec/lib/goodcity/health_checks/dispatched_packages_order_id_check_spec.rb
steveyken: thanks

PR #303, opened 2017-05-12
checked if received_quantity is null before assigning

trojanh:


steveyken:
sorry rajan i missed this PR. I've made a change to the same file and also moved to sql functions. Could you see PR #304

trojanh:
That's fine @steveyken , i was just wondering if we any Package with quantity=0 so that our rake doesn't fail anywhere and fortunately there is no such Package which has quantity=0 on production. So that rake you have modified should work.

PR #304, opened 2017-05-13
Using SQL to speed up the rake task.

steveyken:
I've experiemented with using direct SQL updates for fixing quantity as this speeds things up. It makes a few assumptions in order to do so:

Assumes that:
* a Stockit update does not need to be sent
* a SocketIO notification won't be sent (admin users will need to F5 refresh and reload their mobile apps
* No other active record life cycle events need to be triggered
* When quantity is set to 0 for sent packages, Stockit does not need to be updated.

Can you review and confirm if the above assumptions are correct?

### DIFF
--- a/lib/tasks/packages_add_received_quantity_data.rake
+++ b/lib/tasks/packages_add_received_quantity_data.rake
@@ -11,17 +11,12 @@ namespace :goodcity do
 
     count = 0
     bar = RakeProgressbar.new(Package.count)
-    Package.find_each(batch_size: 100) do |package|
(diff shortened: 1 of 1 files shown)


Darshan-Chauhan:
@steveyken , Yes I think it should be fine If we don't send updates to StockIt(For this rake) even for quantity because when we run rake task which creates OrdersPackages it will updates Packages(qty) which will sent update to StockIt and also StockIt doesn't have received_quantity so its ok.

👍

PR #309, opened 2017-05-22
Fix code climate issues

namrataukirde:
https://github.com/bbatsov/ruby-style-guide#indent-conditional-assignment


shivanibhanwal:
👍

PR #310, opened 2017-05-24
Gcw 1544 specs

namrataukirde:


Gemfile
```diff
@@ -90,4 +90,5 @@ group :test do
   gem "shoulda-callback-matchers"
   gem "codeclimate-test-reporter", require: nil if ENV["CI"]
   gem 'rspec_junit_formatter'
+  gem 'timecop'
```
trojanh: :+1: 
nice discovery.

spec/models/package_spec.rb
```diff
@@ -491,6 +491,75 @@
...
+    let(:dispatched_location) { create :location, :dispatched }
+    let(:order) { create :order }
+    let(:orders_package) { create :orders_package, state: 'dispatched', package: package, order: order }
+    let(:dispatched_location) { create :location, :dispatched }
+
+    it 'creates dispatched packages location record against package if do not exist' do
+      expect{
+        package.create_dispatched_packages_location_from_gc(dispatched_location, orders_package.id, 1)
+      }.to change(PackagesLocation, :count).by(1)
+      expect(package.reload.packages_locations.first.location).to eq dispatched_location
```
steveyken: perhaps remove repetition here:

    first_location = package.reload.package_locations.first.location
    expect(first_location).to eq(...)

spec/controllers/api/v1/packages_controller_spec.rb
```diff
@@ -46,6 +51,37 @@ def test_orders_packages(package, stockit_request, count)
     end
   end
 
+ describe "PUT move_full_quantity" do
+    before { generate_and_set_token(user) }
+
+    before(:all) do
+      Timecop.freeze('2013-04-24'.to_date) { Time.zone.now.to_date }
```
steveyken: appear to be calling this with params and a block... does this make sense?

spec/models/package_spec.rb
namrataukirde: Sure @steveyken , Thanks

spec/controllers/api/v1/packages_controller_spec.rb
namrataukirde: Whenver we will call Time.zone.now.to_date then it will return '2013-04-24'.to_date

PR #311, opened 2017-05-25
Refactoring to improve code climate

trojanh:
Have used 'has_saved' to check if the object has been saved or not as sometimes it is possible to have no errors but unsuccessful save in that case we'll render the old data without throwing the 422 error. So instead of checking object.error.any? I am using 'has_saved' which has the value object.save method.


app/controllers/api/v1/api_controller.rb
```diff
@@ -8,6 +8,28 @@ class ApiController < ApplicationController
     rescue_from Apipie::ParamInvalid, with: :invalid_params
     rescue_from Apipie::ParamMissing, with: :invalid_params
 
+    def serializer_for(object)
+      # using safe_constantize as it will return nil if no object is found
```
namrataukirde: Do not add comment inside method. If you want this comment then add it before method start. Also i don't think we need comment here as its behaviour of `safe_constantize`.

app/controllers/api/v1/addresses_controller.rb
```diff
@@ -27,12 +27,8 @@ class AddressesController < Api::V1::ApiController
     api :POST, '/v1/addresses', "Create an address"
     param_group :address
     def create
-      @address.attributes = address_params
-      if @address.save
-        render json: @address, serializer: serializer, status: 201
-      else
-        render json: @address.errors.to_json, status: 422
-      end
+      has_saved = @address.save
```
namrataukirde: 👍

namrataukirde:
Overall good refactoring. Good work @trojanh 👍

app/controllers/api/v1/donor_conditions_controller.rb
```diff
@@ -16,12 +16,7 @@ class DonorConditionsController < Api::V1::ApiController
     api :GET, '/v1/donor_conditions', "List all donor conditions."
     param :ids, Array, of: Integer, desc: "Filter by donor condition ids e.g. ids = [1,2,3,4]"
     def index
```
steveyken: If we abstract away too much code, it makes it harder to introduce slight variations later and the code is less flexible. I would break this down in to a couple of helper methods.

    render_and_return_cached_json if params[:ids].blank?
    @donor_conditions = @donor_conditions.where(id: params[:ids].split(",") ) if params[:ids].present?
    render json: @donor_conditions, each_serializer: serializer

    def render_and_return_cached_json
      controller.model.cached_json and return
    end

app/services/stockit/base.rb
```diff
@@ -3,27 +3,21 @@
 module Stockit::Base
 
   def post(url, params = {}, options = {})
```
steveyken: this is a good refactor

app/controllers/api/v1/donor_conditions_controller.rb
trojanh: We have this same code repeating three times in district_controller, rejection_reasons and donor_condition controllers so i think its worth making a parent method. We can have a custom code incase we need to modify the above code for any controller. 

I have done all other changes. Do let me know if i need to make any changes.

PR #312, opened 2017-05-30
Allow packages on browse app only if they have Quantity!=0

trojanh:


Darshan-Chauhan:
This will only show the packages with 0 qty. We need to filter packages whose qty is 0

app/models/package.rb
```diff
@@ -48,6 +48,7 @@ class Package < ActiveRecord::Base
   scope :received, -> { where(state: 'received') }
   scope :expecting, -> { where(state: 'expecting') }
   scope :inventorized, -> { where.not(inventory_number: nil) }
+  scope :hasZeroQuantity, -> { where(quantity: 0) }
```
Darshan-Chauhan: Also can you rename the scope to qtyPackage or something? Because {{hasZeroQuantity}} sounds like we're returning boolean value, which is not true

PR #313, opened 2017-05-31
Gcw 1514 turn on qty feature with query param

namrataukirde:


app/models/package.rb
```diff
@@ -64,11 +64,12 @@ class Package < ActiveRecord::Base
 
   attr_accessor :skip_set_relation_update
 
-  def self.search(search_text, item_id)
+  def self.search(search_text, item_id, show_quantity_item = false)
```
steveyken: Perhaps chaining scopes would be cleaner here?

    scope = if item_id.presence
      where("item_id = ? #{received_quantity_query}", item_id)
    else
       where("inventory_number ILIKE :query #{received_quantity_query}", query: "%#{search_text}%")
    end
    scope.where('received_quantity = 1') unless show_quantity_item == 'true'

namrataukirde: Yes, Its looks good. Will change it. Thanks @steveyken

PR #315, opened 2017-06-05
Gcw 1555 add organisations users to allow organisations to browse items and order them

trojanh:


lib/tasks/demo.rake
```diff
@@ -145,6 +146,15 @@ namespace :demo do
       end
     end
 
+    def add_organisations_users(count)
+      count.times do |i|
+        org = FactoryGirl.create(:organisation)
+        user = FactoryGirl.create(:user, mobile: "+8525500000"+(i+1).to_s)
+        FactoryGirl.create(:organisations_user, organisation: org, user: user, role: "Staff")
```
steveyken: How is `role` used currently?

trojanh: It is just a string to represent what position the of the person in the organisation holds i guess.  As we were not using it so far , now is when it'll start mattering as we'll use it on Browse app.

PR #316, opened 2017-06-05
Added rake to onboard new users

trojanh:


lib/tasks/add_browse_users.rake
```diff
@@ -0,0 +1,42 @@
+# rake organisation:add_browse_users
+namespace :organisation do
+  task add_browse_users: :environment do
```
steveyken: please add a description block here so that it shows up when you type 'rake -T'.

lib/tasks/add_browse_users.rake
```diff
@@ -0,0 +1,42 @@
...
+    create_user_for_organisation(f_name, l_name, mobile, o_name)
+  end
+
+  def create_user_for_organisation(f_name, l_name, mobile, o_name)
+    if(org = Organisation.find_by_name_en(o_name))
+      user = User.new(first_name: f_name, last_name: l_name, mobile: mobile)
+      if(user.save)
+        OrganisationsUser.create(organisation: org, user: user )
+      else
+        puts "Incorrect MOBILE_NUMBER"
```
steveyken: You could return the validation errors here instead of assuming bad mobile number.

    user.errors.full_messages

http://api.rubyonrails.org/classes/ActiveModel/Errors.html#method-i-full_messages

lib/tasks/add_browse_users.rake
steveyken: Provide some example usage in the description. If the environment variables are missing, return a message with the usage information.

Could you also change the namespace and task name to `goodcity:add_user` - we'll use it for more than just browse.

lib/tasks/add_browse_users.rake
trojanh: Yeah sure @steveyken Thanks

lib/tasks/add_browse_users.rake
trojanh: Yeah sure @steveyken

lib/tasks/add_browse_users.rake
```diff
@@ -0,0 +1,39 @@
...
+
+  def incorrect_command_prompt
+    puts "Incorrect command:\tEnter command in one of the following format:"
+    puts "\t MOBILE=+85261111111 FIRST_NAME=Steve LAST_NAME=Jobs ORGANISATION=Apple cap production invoke:rake TASK=goodcity:add_user"
+    puts "\t rake goodcity:add_user MOBILE=+85261111111 FIRST_NAME=Steve LAST_NAME=Jobs ORGANISATION=Apple"
+  end
+
+  def add_user_to_organisation(user, o_name)
+    if(org = Organisation.find_by_name_en(o_name))
+      organisation.users << user if(user.save!)
```
steveyken: Code dies at this point because `organisation` variable doesn't exist. A test would have picked this up immediately. For rake tasks, I really would like to see rake as a wrapper to a `lib/` class that performs the actions required. That way, it's easy to write tests to prove it works, easy to call via some other method (e.g. rails console or script) and generally much more flexible.

Suggest the following:
* Move all methods to lib/goodcity/add_user.rb `AddUser` class.
* AddUser class handles user creation and assignment to correct organisation. It's essentially a utility class.
* Rake passes the ENV vars through to the AddUser class
* AddUser is well tested with rspecs.

PR #319, opened 2017-06-22
Gcw 1593 improve inventory number generation code

trojanh:


trojanh:
Thanks @steveyken , your commit optimizes the queries and makes things more clear. 

Just wanted to point one thing here, I think that'll optimize the query even more. I have left a comment on the commit.

app/models/inventory_number.rb
```diff
@@ -1,16 +1,29 @@
...
+    number.to_s.rjust(6, "0")
   end
 
-  def self.latest_code
-    select("CAST(code AS integer)").order("code desc").
-      first.try(:code).try(:to_i) || 1
+  def self.missing_code
+    sql_for_missing_code =
+      "SELECT s.i AS first_missing_code
+        FROM generate_series(1,#{max_code}) s(i)
```
trojanh: Suppose we have following codes in our `InventoryNumber` table: ["000001"," 000002", "999999"], as per this query we'll get `max_code` = 999999, then we generate series from 1 to 999999. and select least non existent number in `missing_code'.

How about taking `max_code` = `InventoryNumber.count` here? 

suppose we have 
["000001", ... , "000096", "000098", "000999", "999999"] 
As we can also enter InventoryNumber code manually , consider the last code entered was "999999" then total`InventoryNumber.count` = 10000 and we'll have to look for missing number between 1 to 100. 
It really not necessary to generate this series `generate_series(1,#{max_code})` from 1 to 999999. I think it is enough if we go from 1 to InventoryNumber.count. 

What do you think?, let me know your thoughts.

steveyken: Yes, this is good thinking. I had thought of another case that wasn't going to work but I realise it was invalid. Can you change to `count` and then merge the code? Thanks

trojanh: Sure @steveyken, Thanks for checking the edge case and I am glad it didn't fail.
i'll update the code and merge it to master.

PR #322, opened 2017-06-26
Moved donor condition to service

namrataukirde:


namrataukirde:
Sure @steveyken , Thanks. Will do this change.

spec/services/package_condition_mapper_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+      expect(PackageConditionMapper.to_stockit("Lightly Used")).to eq('M')
+    end
+  end
+
+  describe '.to_condition' do
+    it 'returns donor condtions for stockit mappings' do
+      expect(PackageConditionMapper.to_condition('N')).to eq('New')
+    end
+  end
+end
```
steveyken: Could you test for empty params too?

app/models/package.rb
```diff
@@ -539,6 +539,10 @@ def dispatched_orders_package
     orders_packages.get_dispatched_records_with_order_id(order_id).first
   end
 
+  def donor_condition_name
+    try(:donor_condition).try(:name_en) || try(:item).try(:donor_condition).try(:name_en)
+  end
```
steveyken: This feels very defensively programmed! Do we need all these 'trys'?

    donor_condition.try(:name_en) || item.try(...)

spec/services/package_condition_mapper_spec.rb
namrataukirde: Sure @steveyken

app/models/package.rb
namrataukirde: We can remove `.try(:name_en)` here.

PR #324, opened 2017-06-27
solve Useless assignment

namrataukirde:


lib/goodcity/compare.rb
```diff
@@ -143,7 +143,7 @@ def compare_orders
     def preload_goodcity_objects(klass)
       memo_key = "preload_#{klass.to_s.underscore}"
       memo = instance_variable_get("@#{memo_key}")
-      memo ||= begin
+      memo || begin
```
steveyken: Actually this is important. Keep the `||=` here. We are effectively saying, "If memo hasn't been set, compute it and assign the result to memo and return memo." If we don't assign the result to `memo` then it will always be `nil` and we'll always re-compute the value.

lib/goodcity/compare.rb
```diff
@@ -155,8 +155,7 @@ def preload_goodcity_objects(klass)
     # Remember which stockit ids we've seen
     def seen_stockit_ids_for(klass)
       memo_key = "seen_stockit_ids_for_#{klass.to_s.underscore}"
-      memo = instance_variable_get("@#{memo_key}")
-      memo || []
+      instance_variable_get("@#{memo_key}") || []
```
steveyken: :+1:

lib/tasks/cloudinary.rake
```diff
@@ -17,11 +17,12 @@ namespace :cloudinary do
   task list_tags: :environment do
     tags = []
     next_cursor = nil
-    while true do
+    is_next_cursor= true
```
steveyken: Much clearer, thank you

lib/goodcity/compare.rb
namrataukirde: Okay... yeah we did memoization here.. 👍  Thanks @steveyken

PR #328, opened 2017-07-05
Gcw 1600 prohibit any invalid package quantity data

trojanh:


app/models/concerns/package_quantity_validator.rb
```diff
@@ -0,0 +1,13 @@
+class PackageQuantityValidator < ActiveModel::Validator
+  def validate(record)
+    package = record.package
+    if(package.present? && is_invalid_total_quantity?(record))
+      record.errors.add(:quantity, "cannot be greater than package quantity")
```
steveyken: Can this say "cannot be greater than total package quantity"

app/models/concerns/package_quantity_validator.rb
```diff
@@ -0,0 +1,13 @@
...
+    end
+  end
+
+  def is_invalid_total_quantity?(record)
+    total = record.package.try(record.class.table_name).where.not(id: record.id).pluck(:quantity).sum + record.quantity
```
steveyken: What happens if it's a new record and `record.id` is nil? You'll get all records in the DB.

trojanh: I think this will work in that case because 
`record.package.try(record.class.table_name).where.not(id: record.id).pluck(:quantity).sum` is actually either of following two. 
```
record.package.try("packages_locations").where.not(id: record.id).pluck(:quantity).sum
or 
record.package.try("orders_packages").where.not(id: record.id).pluck(:quantity).sum
``` 
So if its a new record and `record.id` is nil then we'll get `record.package.try("orders_packages").where.not(id: record.id) = [ ]` and if we use `pluck(:quantity).sum` on this we get answer as 0. 
Still to be sure i tried following on my local db:
```
2.2.2 :009 > Package.first.packages_locations
 => #<ActiveRecord::Associations::CollectionProxy []>

2.2.2 :011 > Package.first.packages_locations.where.not(id: nil).pluck(:quantity).sum
 => 0 
```

So by the above check I think its safe as we trying to search for "orders_package" or "packages_location" on the parent package and if there isn't any record, `pluck` will return nil and `sum` will turn that into 0. So I think its safe. Do let me know if there is something i am missing.

steveyken: ok yes that's fine. make sure there is a test for it though - i.e. what you did on the console

namrataukirde:
👍

spec/models/concerns/package_quantity_validator_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    expect(@package_quantity_validator.validate(packages_location)).to eq(@error_message)
+  end
+
+  it 'validate record when quantity is less than package received_quantity and there is no record' do
+    @package.packages_locations.destroy_all
+    packages_location = PackagesLocation.new(quantity: 10, package: @package)
+    expect(@package_quantity_validator.validate(packages_location)).to be_nil
+  end
+
+  it 'validate mutplile records only when quantity is less than package received_quantity ' do
```
steveyken: small typo here 'multiple'

trojanh: will correct that. Thanks for pointing.

PR #329, opened 2017-07-07
Move designation to service

steveyken:
Making this a PR so I can add some comments. I know the branch isn't complete yet.


app/models/orders_package.rb
```diff
@@ -4,8 +4,8 @@ class OrdersPackage < ActiveRecord::Base
   belongs_to :updated_by, class_name: 'User'
 
   after_initialize :set_initial_state
-  after_create -> { recalculate_quantity("create") }
-  after_update -> { recalculate_quantity("update") }
+  # after_create -> { recalculate_quantity("create") }
```
steveyken: good to see this - a great plus when separating concerns.

app/models/package.rb
```diff
@@ -539,6 +543,14 @@ def dispatched_orders_package
     orders_packages.get_dispatched_records_with_order_id(order_id).first
   end
 
+  def get_designated_orders_packages
```
steveyken: can these just be scopes on orders_packages?

Not a huge fan of the word 'get' - feels very much like Java and not really necessary.

Would be good to see:

    orders_packages.designated

Same re dispached below

app/models/package.rb
```diff
@@ -516,7 +516,11 @@ def is_singleton_package?
   end
 
   def update_location_quantity(total_quantity, location_id)
-    packages_locations.where(location_id: location_id).first.update_column(:quantity, total_quantity)
+    get_packages_location_with_location_id(location_id).update_quantity(total_quantity)
+  end
+
+  def get_packages_location_with_location_id(location_id)
```
steveyken: Do we use the function `get_packages_location_with_location_id` anywhere else? If we're just calling it from the line above then I think it's not necessary to turn it into a one line function. It makes it hard to trace the functionality and flow when reading through the code.

I think in general, I'd prefer not to have one-line functions unless they are called from at least 2 separate places... I know it's a style thing but I think it's a trade off of readability and functionality.

app/services/designation_and_undesignation/base.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    def dispatched_location_id
+      id ||= Location.dispatch_location.id
+    end
+
+    def dispatched_orders_packages
+      package.get_dispatched_orders_packages
+    end
+
+    def initialize(package, order_id, quantity, *args)
```
steveyken: Could you `initialize` this at the top of the class - it's one of the most descriptive methods for answering the question "How do I use this class"?

app/services/designation_and_undesignation/base.rb
```diff
@@ -0,0 +1,62 @@
...
+    def designate_stockit_item
+      package.designate_to_stockit_order(order_id)
+    end
+
+    def designated_orders_packages
+      package.get_designated_orders_packages
+    end
+
+    def dispatched_location_id
+      id ||= Location.dispatch_location.id
```
steveyken: Are you using this to memoize the result? If so, use @id otherwise `id` is just a local variable and it's value is lost when the method returns.

app/services/designation_and_undesignation/same_designation.rb
```diff
@@ -0,0 +1,51 @@
+module DesignationAndUndesignation
```
steveyken: This is really good. Can you indicate with methods are part of the public API by putting them at the top of the class (underneath `initialize`) and then using `private` or `protected` (for base class). Same for all subclasses. Then it will be clear what methods we can/can't use in the rest of the app.

app/services/designation_and_undesignation/base.rb
namrataukirde: Yes I am using it for memoization.

app/services/designation_and_undesignation/base.rb
namrataukirde: Okay Sure, Will do that. I added methods in alphabetize manner.

namrataukirde:
Thanks @steveyken , Really appreciate that you spent time to review it before it completes  and boosts confidence about things we are doing. I will check all your comments while refactoring code and will get back to you in case of any queries or if our I have any thoughts. Thanks alot :)

app/models/package.rb
namrataukirde: Yes Sure. Changing it 👍

app/models/package.rb
namrataukirde: No, Currently we are not using it anywhere else. Will remove this method and we can make it separate whenever we will use it multiple times.

namrataukirde:
Closing this PR as I am not currently working on it. I will reopen it when I will again start working on it.

PR #334, opened 2017-07-14
GCW 1619 prevent not inventoried item from being designated 

trojanh:


app/models/orders_package.rb
```diff
@@ -162,4 +164,14 @@ def update_designation_of_package
   def destroy_stockit_record(operation)
     StockitSyncOrdersPackageJob.perform_now(package.id, self.id, operation) unless package.is_singleton_package?
   end
+
+  def valid_package_state?
+    package.received? && package.inventory_number? && package.locations.present?
```
steveyken: could we split these up into separate validations?

    validates :inventory_number, :presense
    validates :locations, :presense

trojanh: Yeah that will make the code more readable. Thanks @steveyken

app/models/orders_package.rb
```diff
@@ -162,4 +166,22 @@ def update_designation_of_package
   def destroy_stockit_record(operation)
     StockitSyncOrdersPackageJob.perform_now(package.id, self.id, operation) unless package.is_singleton_package?
   end
+
+  def package_inventory_number
+    if !package.inventory_number?
```
namrataukirde: Can we use unless here instead of if ! ?

app/models/orders_package.rb
```diff
@@ -3,7 +3,11 @@ class OrdersPackage < ActiveRecord::Base
   belongs_to :package
   belongs_to :updated_by, class_name: 'User'
 
+  validate :package_inventory_number, :on => :create
```
namrataukirde: Use new syntax here.  on:  :create

trojanh: :+1:

app/models/orders_package.rb
trojanh: :+1:

namrataukirde:
Closing this PR as we are not currently working on it.

PR #339, opened 2017-07-27
Gcw 1652 add package type validation

Darshan-Chauhan:


trojanh:
We have planned to put this on hold until we fix all the prod data issues mentioned on [GCW-1652](https://jira.crossroads.org.hk/browse/GCW-1652)

PR #361, opened 2017-10-13
add organisation in order serializer

viveksriram1:
In this PR order serializer will send organisation_id with order


app/controllers/api/v1/orders_controller.rb
```diff
@@ -54,6 +54,7 @@ def show
         root: root,
         exclude_code_details: true,
         include_packages: true,
+        include_organisation: true,
```
Darshan-Chauhan: where is this used?

PR #364, opened 2017-10-27
Gcw 1711 add mixed donor condition for nil door condtion

viveksriram1:


lib/tasks/assign_default_donor_condition_to_packages.rake
```diff
@@ -0,0 +1,25 @@
...
+    log = Goodcity::RakeLogger.new("assign_default_donor_condition")
+    lightly_used = DonorCondition.find_by(name_en: 'Lightly Used')
+
+    nil_donor_condition_count = Package.where(donor_condition_id: nil).count
+    log.info(": #{nil_donor_condition_count} records has nil donor_condition")
+
+    count = 0
+
+    begin
+      Package.where(donor_condition_id: nil).find_each(batch_size: 50) do |package|
```
Darshan-Chauhan: Hi @viveksriram1 
- I don't think we need to fire **Package.where** twice, once while counting and second time here.
It would be better if we can store **Package.where(donor_condition_id: nil)** in variable `xyz` and create another variable for count.
- We can use the variable **`xyz.find_each(batch_size: 50)`**.
- This would also make rake task faster.
- count query makes rake task significantly slower and I don't think it's that important to show how many such records are there because we're already showing how many records are updated.

What do you think?

CC: @steveyken , @namrataukirde

namrataukirde: I think we should not store result of where query in variable as it will try to dump all records in memory which we do not want. I  think we can only count total number of records updated on completion of rake task.

viveksriram1:
Yes @steveyken It will update Stockit

PR #365, opened 2017-10-30
GCW-1728 assign values for code created from stockit

namrataukirde:
Hi Team,

This PR is to create new package_type  record on api with all details when new code is created from stockit.

https://jira.crossroads.org.hk/browse/GCW-1728

Thanks.


namrataukirde:
Sure @steveyken Adding test cases for same. Thanks.

PR #381, opened 2017-12-01
#DecemberRelease1

namrataukirde:
Hi Team,

This PR contains changes for rollbar and rake to add new permission.

Thanks.

### DIFF
--- a/app/models/package.rb
+++ b/app/models/package.rb
@@ -183,7 +183,7 @@ class Package < ActiveRecord::Base
   end
 
   def build_or_create_packages_location(location_id, operation)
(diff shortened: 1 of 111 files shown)


app/models/package.rb
```diff
@@ -258,7 +259,7 @@ def designation
   end
 
   def cancel_designation
-    designation and designation.cancel!
+    designation ** designation.cancel!
```
steveyken: Is ** correct here?

namrataukirde: Its wrong. It should be &&. Thanks @steveyken 👍

PR #410, opened 2018-04-05
Gcw 1789 permission model structure changes

namrataukirde:
Hi Team,

This PR has changes to change permission model structure. Please review and let me know if any changes are required.

https://jira.crossroads.org.hk/browse/GCW-1774

Thanks.


app/models/ability.rb
```diff
@@ -205,62 +238,66 @@ def public_ability
...
-    can [:create, :remove_number], InventoryNumber if api_user_or_staff?
+    if can_add_or_remove_inventory_number? || @api_user
+      can [:create, :remove_number], InventoryNumber
+    end
+  end
+
+  def schedule_abilities
+    can [:create, :availableTimeSlots], Schedule
+    can [:index, :show], Schedule, deliveries: { offer_id: @user_offer_ids }
+    can [:index, :show], Schedule if can_read_schedule?
```
steveyken: Can we make this `can_read_schedules?` just to be clear

app/models/ability.rb
```diff
@@ -205,62 +238,66 @@ def public_ability
...
+      can [:create, :index], StockitActivity
+    end
   end
 
   def user_abilities
-    can [:show, :update], User, id: @user_id
-    can [:index, :show, :update], User if staff?
     can :current_user_profile, User
+    can [:show, :update], User, id: @user_id
+    can [:index, :show, :update], User if can_read_or_modify_user?
```
steveyken: Can we make this `can_read_or_update_users?` 'Modify' is fine but we already use 'update'

app/models/organisations_user.rb
```diff
@@ -4,11 +4,16 @@ class OrganisationsUser < ActiveRecord::Base
...
-  after_create :send_welcome_msg
+  after_create :send_welcome_msg, :create_user_role
 
   private
 
   def send_welcome_msg
     TwilioService.new(user, organisation).send_welcome_msg
   end
+
+  def create_user_role
```
steveyken: Perhaps clearer if this was `def add_charity_role`

app/models/role_permission.rb
```diff
@@ -0,0 +1,4 @@
+class RolePermission < ActiveRecord::Base
```
steveyken: Shouldn't this be the `PermissionsRoles` class?

app/models/user.rb
```diff
@@ -136,8 +141,20 @@ def recent_active_offer_id
...
+  def create_or_remove_user_roles(role_ids)
+    remove_user_roles(role_ids)
+    role_ids.each do |role_id|
+      user_roles.where(role_id: role_id).first_or_create
+    end
+  end
+
   private
 
+  def remove_user_roles(role_ids)
```
steveyken: We don't appear to be using this method anywhere else apart from in `create_or_remove_user_roles` so I would be tempted to merge this into `create_or_remove_user_roles` and change the method name to `set_user_roles`. This is because when I read `remove_user_roles(role_ids)` I think that `role_ids` is the list of roles to remove from a user but actually that is not the case here, its the list of roles to keep! E.g. if `role_ids` is `nil` then it will remove *all* roles rather than none.

Also, shouldn't `roles.pluck(:id) - role_ids` be the other way around? To get the list of roles to remove.

```ruby
def set_user_roles(role_ids)
  role_ids_to_remove =  role_ids - roles.pluck(:id)
  user_roles.where(role_id: role_ids_to_remove).destroy_all
  role_ids.each do |role_id|
    user_roles.where(role_id: role_id).first_or_create
  end
end
```

app/models/ability.rb
```diff
@@ -25,101 +44,86 @@ def initialize(user)
...
-
-  def stockit_contact_abilities
-    can [:create, :index], StockitContact if @api_user
-  end
-
-  def stockit_local_order_abilities
-    can [:create, :index], StockitLocalOrder if @api_user
+  def gogovan_order_abilities
+    can [:calculate_price, :confirm_order, :destroy], GogovanOrder, delivery: { offer_id: @user_offer_ids }
+    can [:calculate_price, :confirm_order, :destroy], GogovanOrder if can_handle_gogovan_order?
```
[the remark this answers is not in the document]
namrataukirde: Sure @steveyken

app/models/ability.rb
```diff
@@ -187,7 +214,13 @@ def package_abilities
     end
     can :create, Package if @api_user
     can :destroy, Package, item: { offer: { created_by_id: @user_id }, state: 'draft' }
-    can :destroy, Package, item: { state: 'draft' } if @reviewer
+    can :destroy, Package, item: { state: 'draft' } if can_destroy_package_with_specific_states?
+  end
+
+  def package_type_abilities
+    if can_add_package_types? || @api_user
```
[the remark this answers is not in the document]
namrataukirde: You mean we can create record in role table with name `api_user` ?

app/models/ability.rb
```diff
@@ -205,62 +238,66 @@ def public_ability
...
 
-  def address_abilities
-    # User address
-    can [:create, :show], Address, addressable_type: "User", addressable_id: @user_id
-
-    # Offer delivery address
-    can [:create, :show, :destroy], Address, addressable_type: "Contact", addressable: { delivery: { offer_id: @user_offer_ids } }
-    can [:create, :show, :destroy], Address, addressable_type: "Contact" if staff?
+  def packages_locations_abilities
+    if can_access_packages_locations? || @api_user
```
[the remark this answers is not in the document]
namrataukirde: Sure @steveyken 👍

app/models/role_permission.rb
namrataukirde: Yes, It can be. I found this name more reliable. But we can change it. 👍

app/controllers/api/v1/users_controller.rb
```diff
@@ -37,6 +37,7 @@ def show
       end
       def update
         @user.update_attributes(user_params)
+        @user.create_or_remove_user_roles(params["user"]["user_role_ids"])
```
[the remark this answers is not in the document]
namrataukirde: @steveyken, I think yes, When user is logged in as supervisor then he can assign himself as admin and can remove supervisor role. Thus when request will come to api, at that time that user will be supervisor (i.e when we are checking user is supervisor or not). After that user will no more supervisor and he will be admin.

In this case we need to check what will happen with user as he was already logged in. 

Its an amazing h-case to test. I would like to know what is expected behaviour here and thoughts on this.

CC: @mattgow

app/models/ability.rb
```diff
@@ -205,62 +238,66 @@ def public_ability
...
   end
 
   def taxonomies
     can :register, :device
     can [:index, :show], DonorCondition
     can [:index, :show], SubpackageType
     can [:index, :show], RejectionReason
+    can [:index, :show], Role
     can [:index, :show], Permission
+    can [:index, :show], UserRole
```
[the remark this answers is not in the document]
namrataukirde: Yes we can add one clause here which will check if user has admin role or not. 👍

app/models/ability.rb
```diff
@@ -25,101 +44,86 @@ def initialize(user)
...
+      can [:index, :show, :create, :update, :destroy, :delete_cloudinary_image], Image
+    else
+      can [:index, :show, :create, :update, :destroy], Image, Image.donor_images(@user_id) do |record|
+        record.imageable.offer.created_by_id == @user_id
+      end
+    end
+    can :destroy, Image, imageable: { offer: { created_by_id: @user_id },
+      state: ['draft', 'submitted', 'scheduled'] }
+    can :destroy, Image, imageable: {
+      state: ['draft', 'submitted', 'accepted', 'rejected', 'scheduled'] } if can_destroy_image_for_imageable_states?
```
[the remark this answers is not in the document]
namrataukirde: We can remove condition can_destroy_image_for_imageable_states? And reviewer should have can_manage_images  permisssion.

app/models/role_permission.rb
steveyken: Doesn't the Rails magic happen if it is `PermissionsRoles` ?

app/models/user.rb
steveyken: good catch for `not in []` ! as long as it is well tested, this will work

app/models/user_role.rb
```diff
@@ -0,0 +1,8 @@
+class UserRole < ActiveRecord::Base
```
steveyken: `RolesUsers` ?? for rails conventions to work properly?

app/serializers/api/v1/role_permission_serializer.rb
```diff
@@ -0,0 +1,9 @@
+module Api::V1
+  class RolePermissionSerializer < ApplicationSerializer
+    embed :ids, include: true
+
+    attributes :id, :role_id
+
+    has_one :permission, serializer: PermissionSerializer
```
steveyken: `has_one`? aren't there many permissions per role?

db/migrate/20180214104436_create_role_permissions.rb
```diff
@@ -0,0 +1,10 @@
+class CreateRolePermissions < ActiveRecord::Migration
+  def change
+    create_table :role_permissions do |t|
+      t.integer :role_id
+      t.integer :permission_id
```
steveyken: need to add indexes?

spec/controllers/api/v1/organisations_user_controller_spec.rb
```diff
@@ -39,6 +39,13 @@
       expect(JSON.parse(response.body)["errors"]).to eq("Mobile is invalid")
     end
 
+    it "assigns 'charity' role to user and creates user_role", :show_in_doc do
+      expect {
+        post :create, format: :json, organisations_user: organisations_user_params
+      }.to change(UserRole, :count).by(1)
+      expect(OrganisationsUser.last.user.roles.pluck(:name)).to include('Charity')
```
steveyken: what about `expect(OrganisationsUser.last.user.roles).to eql(charity_role)` ?

spec/controllers/api/v1/roles_controller_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  describe "GET roles" do
+    before { generate_and_set_token(user) }
+    it "returns 200" do
+      get :index
+      expect(response.status).to eq(200)
+    end
+
+    it "return serialized offers", :show_in_doc do
```
steveyken: `return serialized roles`

spec/factories/roles.rb
```diff
@@ -0,0 +1,151 @@
...
+
+  factory :role do
+    name            { %w( Reviewer Supervisor Administrator ).sample }
+    initialize_with { Role.find_or_initialize_by(name: name) } # limits us to our sample of permissions
+
+    transient do
+      permissions { %w(can_manage_offers, can_manage_packages)}
+    end
+
+    trait :with_dynamic_permission do
```
steveyken: nice!

steveyken:
Some amazing work in here @namrataukirde Thanks for pushing through such a massive job!

spec/controllers/api/v1/roles_controller_spec.rb
namrataukirde: 😅  👍

spec/factories/roles.rb
namrataukirde: Thanks :)

PR #411, opened 2018-04-24
GCW-1861 fix issue: remove all roles for user should work

namrataukirde:


Darshan-Chauhan:
Yes @steveyken , I have added model test cases previously for same. But controller test case is missing for this specific scenario. I will adding it today. Thanks :)

namrataukirde:
Yes @steveyken , I have added model test cases previously for same. But controller test case is missing for this specific scenario. I will adding it today. Thanks :)

PR #438, opened 2018-05-08
Access restricted to charity users for browse app

Darshan-Chauhan:


namrataukirde:
@Darshan-Chauhan for specs 👍

PR #451, opened 2018-05-11
GCW-1744 Package column updates on Stockit designate and dispatch

Darshan-Chauhan:
Hi Team,

This PR contains changes for GCW:1744:- set values for stockit_designated_by_id and stockit_sent_by_id columns	

Jira ticket link:- https://jira.crossroads.org.hk/browse/GCW-1744


app/controllers/api/v1/packages_controller.rb
```diff
@@ -312,6 +312,8 @@ def package_record
           @package.inventory_number = inventory_number
           @package.box_id = box_id
           @package.pallet_id = pallet_id
+          @package.stockit_designated_by_id = designated_from_stockit ? current_user_id : nil
```
namrataukirde: Hi @Darshan-Chauhan It would be good to write instance method in package model and perform this operation in that method. 

Just a suggestion. Correct me if you have any other thoughts on it. :)

namrataukirde:
Specs looks good. 👍 

You can keep those specs. It would be good if add same specs for other scenario's of designation from stockit.

namrataukirde:
Closing this PR for now.

PR #455, opened 2018-05-15
Gcw 1872 fix offers load for supervisor in donor app

namrataukirde:


spec/controllers/api/v1/offers_controller_spec.rb
```diff
@@ -86,6 +87,35 @@
...
+        context "for donor app"
+          let!(:offer1) { create :offer, created_by: supervisor }
+          let!(:offer2) { create :offer }
+          before { generate_and_set_token(supervisor) }
+
+          it 'returns offers created_by supervisor' do
+            request.headers["X-GOODCITY-APP-NAME"] = "app.goodcity"
+            get :index
+            expect(assigns(:offers).to_a).to eql([offer1])
+            expect(assigns(:offers).to_a).not_to include(offer2)
```
steveyken: technically the line above does the same check

namrataukirde: Yup.. Thanks @steveyken . I will remove that unnecessary check. :)

PR #465, opened 2018-05-24
allow_web_publish is set to false

abulsayyad123:
Hi,
This is PR for GCW-1888-Undesignate should live update items availability on Browse.
Thanks


app/models/package.rb
```diff
@@ -358,6 +360,7 @@ def undesignate_from_stockit_order
     self.order = nil
     self.stockit_designated_on = nil
     self.stockit_designated_by = nil
+    last_allow_web_published and self.allow_web_publish = last_allow_web_published
```
steveyken: Can we update this to ```self.allow_web_publish = last_allow_web_published if last_allow_web_published```

namrataukirde:
Closing this PR for now. We will work on different approach when we will resume work on it.

PR #471, opened 2018-05-29
GCW-1910 stock order workflow

Darshan-Chauhan:
Hi Team,

This PR adds state workflow for Order processing.

Jira Ticket:- https://jira.crossroads.org.hk/browse/GCW-1910

### DIFF
--- a/app/controllers/api/v1/orders_controller.rb
+++ b/app/controllers/api/v1/orders_controller.rb
@@ -59,6 +59,21 @@ module Api
           exclude_stockit_set_item: true
       end
 
+      def start_process
+        @order.start_processing(current_user) if @order.submitted?
+        render json: @order, serializer: serializer
+      end
+
+      def finish_process
+        @order.finish_processing(current_user) if @order.processing?
+        render json: @order, serializer: serializer
+      end
+
+      def cancel_order
+        @order.cancel(current_user) if @order.processing?
+        render json: @order, serializer: serializer
+      end
+
       def update
         @order.assign_attributes(order_params)
         # use valid? to ensure submit event errors get caught
--- a/app/controllers/api/v1/orders_controller.rb
+++ b/app/controllers/api/v1/orders_controller.rb
@@ -60,22 +60,22 @@ module Api
       end
 
       def start_process
-        @order.start_processing(current_user) if @order.submitted?
+        @order.start_processing
         render json: @order, serializer: serializer
       end
 
       def finish_process
-        @order.finish_processing(current_user) if @order.processing?
+        @order.finish_processing
         render json: @order, serializer: serializer
       end
 
       def cancel_order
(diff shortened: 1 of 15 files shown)


app/controllers/api/v1/orders_controller.rb
```diff
@@ -59,6 +59,26 @@ def show
           exclude_stockit_set_item: true
       end
 
+      def start_process
+        @order.start_processing(current_user) if @order.submitted?
```
namrataukirde: Can we move `if @order.submitted?` to state transition callback ?

```
before_transition on: :start_processing do |order|
      if @order.submitted?
        order.processed_at = Time.now
        processed_by_id = User.current_user
      end
end
```

namrataukirde:
👍

app/controllers/api/v1/orders_controller.rb
Darshan-Chauhan: Thanks

app/models/concerns/push_updates.rb
```diff
@@ -8,14 +8,22 @@ module PushUpdates
   end
 
   def update_client_store(operation)
+    type  = self.class.name
```
steveyken: I wonder if it would be best to have a separate OrderPushUpdate class here. Perhaps inheriting from PushUpdateBase. Let's talk this one through and come up with a plan.

app/controllers/api/v1/orders_controller.rb
```diff
@@ -59,6 +59,26 @@ def show
           exclude_stockit_set_item: true
       end
 
+      def start_process
```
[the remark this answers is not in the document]
Darshan-Chauhan: Thanks @steveyken , we don't have methods for each transition. It's just a name of controller method.
We've directly called `@order.start_processing` which is a state event and not a method.

app/models/concerns/push_updates.rb
Darshan-Chauhan: Sure, It would be better as later we might include pusUpdates for other models also.

app/controllers/api/v1/orders_controller.rb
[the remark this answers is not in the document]
steveyken: Sorry, I wasn't very clear before, I was thinking instead of controller actions for every transition, why don't we just have one controller action and pass the transition as a parameter. E.g.

```
OrdersController
    def transition
      @transition = params['transition']   # e.g. 'start_processing'
      if @order.state_events.include?(@transition)
        @order.fire_state_event(@transition)
      end
      ...
    end
```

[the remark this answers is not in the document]
Darshan-Chauhan: Thanks @steveyken , looks much better.
Will do it tonight. :)

PR #497, opened 2018-06-07
orders_package delete method added

abulsayyad123:
Hi,
This is pull request for issue GCW-1936 and GCW-1937
Thanks


app/models/orders_package.rb
```diff
@@ -26,6 +27,10 @@ def set_initial_state
     self.state ||= :requested
   end
 
+  def delete_order_if_no_package
```
steveyken: @abulsayyad123 how do we check that there are no packages?

PR #500, opened 2018-06-08
notify order fulfilment user about new order submission from browse app

namrataukirde:
Hi Team, 

This PR is to notify order fulfilment user about new order submission from goodcity.

https://jira.crossroads.org.hk/browse/GCW-1909

Please review and let me know if any changes required.

Thanks

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -169,8 +169,22 @@ class Order < ActiveRecord::Base
     end
 
     after_transition on: :submit do |order|
-      order.designate_orders_packages if order.detail_type == "GoodCity"
+      if order.detail_type == "GoodCity"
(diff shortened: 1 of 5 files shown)


app/models/order.rb
```diff
@@ -169,10 +169,24 @@ def set_initial_state
...
     after_transition on: :submit do |order|
-      order.designate_orders_packages if order.detail_type == "GoodCity"
+      if order.detail_type == "GoodCity"
+        order.designate_orders_packages
+        order.send_new_order_notificationen
+      end
     end
   end
 
+  def send_new_order_notificationen
```
steveyken: maybe rename this to `send_new_order_notification` now that it is i18n'd?

PR #513, opened 2018-06-20
GCW 2005 order submission details

Darshan-Chauhan:
Hi Team,

This PR adds functionality to add timestamp and user detail when GC Order is Submitted.
It also updates OrderSerializer to send the same details.

Jira ticket:- https://jira.crossroads.org.hk/browse/GCW-2005

Note: This PR also upgrade `sprockets` gem to `3.7.2`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -43,6 +43,7 @@ gem 'oj', '2.10.2' # 2.10.3 causes a 'too deeply nested' error
 gem 'oj_mimic_json'
 gem 'redis'
 gem 'redis-rails', '~> 5.0.2'
+gem 'sprockets', '~>3.7.2'
 gem 'rollbar'
 gem 'apipie-rails' , git: "https://github.com/Apipie/apipie-rails.git", branch: 'master'
 gem "go_go_van_api", git: "git@github.com:crossroads/go_go_van_api.git", branch: 'master'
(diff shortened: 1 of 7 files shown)


app/models/order.rb
```diff
@@ -115,6 +116,8 @@ def set_initial_state
     end
 
     before_transition on: :submit do |order|
+      order.submitted_at = Time.now
+      order.submitted_by_id = User.current_user.id
```
namrataukirde: Can we directly assign object here?

`order.submitted_by = User.current_user`

db/migrate/20180620045951_add_submitted_at_to_orders.rb
```diff
@@ -0,0 +1,5 @@
+class AddSubmittedAtToOrders < ActiveRecord::Migration
+  def change
+    add_column :orders, :submitted_at, :datetime
```
namrataukirde: Can we do it in single migration instead of separate migrations ?

Gemfile
```diff
@@ -43,6 +43,7 @@ gem 'oj', '2.10.2' # 2.10.3 causes a 'too deeply nested' error
 gem 'oj_mimic_json'
 gem 'redis'
 gem 'redis-rails', '~> 5.0.2'
+gem 'sprockets', '~>3.7.2'
```
namrataukirde: Do we need this gem ?

Darshan-Chauhan: Yes, build was failing as it was showing `sprockets` had vulnerability and needed to upgrade to `3.7.2`

namrataukirde: Ohh Okay, Thanks for clarification 👍

PR #518, opened 2018-06-26
Gcw 1919 upgrade ruby 2 5

steveyken:
Work to upgrade ruby to v2.5.1

- removed unnecessary to_json calls in controller renders
- used json hash to re-implement `{'meta': search ... }` with object hashes
- fixed failing tests - need to use as_json to return object hash
- upgraded FactoryBot and other various gems


steveyken:
Still need to test against apps to confirm API is consistent. Perhaps can be done on staging but might be good to wait and test on dev machines before we merge. Is anyone able to do this?

I have installed ruby-2.5.1 on staging server so api will run and deploy ok when this branch is merged.

Gemfile.lock
```diff
@@ -434,8 +433,8 @@ DEPENDENCIES
   dotenv-rails (= 0.11.1)
   easyzpl!
   execjs
-  factory_girl_rails
-  ffaker (~> 2.8.0)
+  factory_bot_rails
```
namrataukirde: Hi @steveyken , Looks like we have replaced `factory_girl_rails` with `factory_bot_rails`. Is that ruby 2.5.1 do not support it or its better than factorygirl ? Could you please tell me reason behind it. 

It would be good to know. :)

spec/controllers/api/v1/inventory_numbers_controller_spec.rb
```diff
@@ -12,17 +13,18 @@
         post :create
       }.to change(InventoryNumber, :count).by(1)
       expect(response.status).to eq(200)
+      expect(parsed_body['inventory_number']).to eql(InventoryNumber.last.code)
     end
   end
 
   describe "PUT /v1/inventory_numbers Delete InventoryNumber" do
     it 'deletes existing inventory_number and returns blank json' do
+      inventory_number
```
namrataukirde: Hi @steveyken Do we need to add this ? Because in post request we are calling it right ? So i think object will get created. 🤔 

I think this line is added due to use of `let`.

namrataukirde: Would like to know this as it will be helpful for me. Nothing else 😅

namrataukirde:
Just 1 thing to mention, I have installed new ruby version on local on friday. I have started checking this branch on local today.

namrataukirde:
Hi @steveyken , I have tested this PR on local with admin app.

Checked json reponse without to_json method which we removed at most of places. I feel this PR is safe to go on staging and we can do further testing on staging.

Thanks.

Gemfile
```diff
@@ -23,13 +19,13 @@ gem 'dotenv-rails', '0.11.1' # v1.0.2 of dotenv-rails doesn't preload ENV before
 gem 'cancancan'
 gem 'loofah'
 gem 'cloudinary'
-gem 'factory_girl_rails' # used in rake db:seed in production
-gem 'ffaker', '~> 2.8.0'
+gem 'factory_bot_rails' # used in rake db:seed in production
```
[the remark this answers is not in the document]
steveyken: Factory girl has been replaced with factory bot (namespace change). All new releases will be as factory bot so we need to change to stay up to date.

spec/controllers/api/v1/inventory_numbers_controller_spec.rb
steveyken: Yes I replaced "let!" - I'm not a big fan of it because it's easy to miss. "Let" is much clearer

steveyken:
Thanks, yes happy for you to merge once other changes have been pushed to live. I can merge master in to this branch to resolve conflicts.

We can remove `to_json` in render statements because it is called by the underlying Json renderer if we provide a hash instead.

app/controllers/api/v1/cancellation_reasons_controller.rb
```diff
@@ -4,11 +4,11 @@ class CancellationReasonsController < Api::V1::ApiController
       load_and_authorize_resource :cancellation_reason, parent: false
 
       def index
-        if params[:ids].blank?
-          render json: CancellationReason.visible.cached_json
-          return
+        if params["ids"].present?
+          @cancellation_reasons = @cancellation_reasons.where(id: params["ids"].split(",").flatten)
```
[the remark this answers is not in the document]
steveyken: Yes, I don't think we really use this much at all in the apps but it's there if we want to request multiple objects by id. Ember Data can do this in some cases when related data isn't loaded and it needs to get it.

Gemfile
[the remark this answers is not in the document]
namrataukirde: 👍

spec/controllers/api/v1/inventory_numbers_controller_spec.rb
namrataukirde: Okay, Sure 👍

app/controllers/api/v1/cancellation_reasons_controller.rb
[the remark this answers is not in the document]
namrataukirde: 👍

namrataukirde:
Thanks @steveyken. Appreciate your efforts on this. :)

PR #525, opened 2018-07-04
Slack connector for messages

steveyken:


app/services/slack_pin_service.rb
```diff
@@ -0,0 +1,16 @@
+class SlackPinService
+
+  attr_accessor :user
+
+  def initialize(user)
+    @user = user
+  end
+
+  def send_otp
+    channel = ENV['SLACK_PIN_CHANNEL']
```
namrataukirde: This we are going to add to .env file. Right ?

namrataukirde: Can we add this variable to .env file with dummy channel name ?

PR #530, opened 2018-07-16
GCW-2033 new order sms notification for order fulfilment users

Darshan-Chauhan:
Hi Team,

This PR adds feature of sending sms notification to Users with "order fulfilment" role whenever any new GoodCity order is submitted.

Jira ticket:- https://jira.crossroads.org.hk/browse/GCW-2046

Please review the PR and let me know if any changes are required.
Thanks

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -219,7 +219,9 @@ class Order < ActiveRecord::Base
   end
 
   def send_order_placed_sms_to_order_fulfilment_users
-    User.order_fulfilment.map { |user| TwilioService.new(user).order_submitted_sms_to_order_fulfilment_users(self) }
+    User.order_fulfilment.each do |user|
+      TwilioService.new(user).order_submitted_sms_to_order_fulfilment_users(self)
+    end
   end
 
   def nullify_columns(*columns)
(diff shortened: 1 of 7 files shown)


app/models/order.rb
```diff
@@ -213,8 +214,12 @@ def send_new_order_notificationen
     }
   end
 
-  def send_new_order_sms
-    TwilioService.new(submitted_by).send_new_order_sms(self)
+  def send_new_order_confirmed_sms_to_charity
+    TwilioService.new(submitted_by).order_confirmed_sms_to_charity(self)
```
namrataukirde: Can we pass order object to initializer instead of argument to method ?

Darshan-Chauhan: We can but what difference will it make?? I mean whats the advantage of passing it to initializer instead of passing to method?
I only need to use it in one method so I think passing to initializer would be more useful if it's needed in the Class somewhere else too.

What do you think @namrataukirde ?

namrataukirde: Ohh okay, So we are using it in single method. Then approach you have used is correct. Thanks for clarification @Darshan-Chauhan :)

Darshan-Chauhan: Sure :)

Darshan-Chauhan:
Thanks for adding the specs @namrataukirde :)

PR #536, opened 2018-07-25
Gcw 2055 transport details information needed

abulsayyad123:
Hi,
This is PR for jira ticket GCW-2055-Stock: Order information, transport detail information need more details.
https://jira.crossroads.org.hk/browse/GCW-2055.
Please review it. If any change is required then please let me know.
Thanks


app/serializers/api/v1/contact_serializer.rb
```diff
@@ -1,7 +1,15 @@
 module Api::V1
   class ContactSerializer < ApplicationSerializer
     embed :ids, include: true
-    attributes :id, :name, :mobile
+    attributes :id, :name, :mobile, :mobile_phone_number
     has_one :address, serializer: AddressSerializer
+
+    def mobile_phone_number
```
Darshan-Chauhan: @abulsayyad123  Why do we need to send ** mobile_phone_number** when we're already sending ** mobile??**

app/serializers/api/v1/order_serializer.rb
```diff
@@ -9,7 +9,7 @@ class OrderSerializer < ApplicationSerializer
       :dispatch_started_by_id, :submitted_at, :submitted_by_id
 
     has_one :created_by, serializer: UserProfileSerializer, root: :user
-    has_one :stockit_contact, serializer: StockitContactSerializer, root: :contact
+    has_one :stockit_contact, serializer: StockitContactSerializer
```
Darshan-Chauhan: @abulsayyad123  What was the reason for removing **root: :contact?**

abulsayyad123: because the main contact serializer was getting blocked by stockit_contact root:contact.

Darshan-Chauhan: removing that is useful for GC orders but I think **StockItLocalOrders** might be using ** stockit_contact** . What will happen to LocalOrders on front-end because of this change?

PR #543, opened 2018-08-06
Gcw 1917 allow multi quantity sync from admin

namrataukirde:
Hi Team,

This PR is to allow multi quantity sync from admin when package received from admin with inventory number.

https://jira.crossroads.org.hk/browse/GCW-1917

Please review and let me know in case of any changes.

Thanks.

### DIFF
--- a/app/jobs/stockit_update_job.rb
+++ b/app/jobs/stockit_update_job.rb
@@ -1,11 +1,11 @@
 class StockitUpdateJob < ActiveJob::Base
   queue_as :stockit_updates
 
-  def perform(package_id, request_from_admin = nil)
+  def perform(package_id, allow_sync = nil)
     package = Package.find_by(id: package_id)
 
     if package
-      response = Stockit::ItemSync.update(package, request_from_admin)
+      response = Stockit::ItemSync.update(package, allow_sync)
 
       if response && (errors = response["errors"] || response[:errors])
         log_text = "Inventory: #{package.inventory_number} Package: #{package_id}"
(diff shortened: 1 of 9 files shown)


app/jobs/stockit_update_job.rb
```diff
@@ -1,11 +1,11 @@
 class StockitUpdateJob < ActiveJob::Base
   queue_as :stockit_updates
 
-  def perform(package_id)
+  def perform(package_id, allow_sync = nil)
     package = Package.find_by(id: package_id)
 
     if package
-      response = Stockit::ItemSync.update(package)
+      response = Stockit::ItemSync.update(package, allow_sync)
```
steveyken: is there a need to pass `allow_sync`  into `Stockit::ItemSync` - can you just say `if package && allow_sync` here instead?

namrataukirde: Agree. Thanks @steveyken 👍

namrataukirde: Update from discussion : 

I feel right now we should pass `allow_sync` attribute to item_sync method. Because at some places that service is called directly and not through any job.

- If we do not pass that parameter for Stockit::ItemSync.update(id, allow_sync) . Then we can create one another method which will check for (singleton || request_from_admin). If that method is not used with `if` for calling item sync in any scenario then it will cause unwanted sync. (i.e if someone calls item_sync code newly in any of code).
- But if we pass it as argument then person will check what it is and such scenario can be taken care or will be noticed.

Thus keeping it as it is.

Thanks.

PR #545, opened 2018-08-27
Gcw 2069 add ngo user as charity user

abulsayyad123:
Hi Team,
This is PR for jira ticket GCW-2069- Cannot assign NGO staffs who are registered in the donor app as a charity user.
https://jira.crossroads.org.hk/browse/GCW-2069.
I have covered all the possible scenarios.
Please review the code and let me know if any change is required.

Thanks
@steveyken @namrataukirde @Darshan-Chauhan

### DIFF
--- a/app/controllers/api/v1/organisations_users_controller.rb
+++ b/app/controllers/api/v1/organisations_users_controller.rb
@@ -27,11 +27,11 @@ module Api
       api :POST, "/v1/organisations_user", "Create a package"
       param_group :organisations_user
       def create
-        organisation_user = OrganisationsUserBuilder.new(@organisations_user).build
-        if organisation_user.valid?
+        builder = OrganisationsUserBuilder.new(@organisations_user).build
+        if builder['result'] == true
           render json: organisation_user, serializer: serializer, status: 201
         else
-          render_error(organisation_user.errors.full_messages.join('. '))
+          render_error(builder['errors'])
         end
       end
 
--- a/app/controllers/api/v1/organisations_users_controller.rb
+++ b/app/controllers/api/v1/organisations_users_controller.rb
@@ -27,7 +27,8 @@ module Api
       api :POST, "/v1/organisations_user", "Create a package"
       param_group :organisations_user
       def create
-        save_and_render_object_with_errors(@organisations_user)
+        org_user = OrganisationsUser.organisation_user?(@organisations_user)
+        save_and_render_object_with_errors(org_user)
       end
 
       private
--- a/app/controllers/api/v1/organisations_users_controller.rb
+++ b/app/controllers/api/v1/organisations_users_controller.rb
@@ -27,8 +27,7 @@ module Api
       api :POST, "/v1/organisations_user", "Create a package"
       param_group :organisations_user
       def create
-        organisation_user = OrganisationsUser.create_or_update_existing_user_object(@organisations_user)
-        save_and_render_object_with_errors(organisation_user)
+        save_and_render_object_with_errors(@organisations_user.create_or_update_existing_organisation_user)
       end
 
       private
--- a/app/controllers/api/v1/organisations_users_controller.rb
+++ b/app/controllers/api/v1/organisations_users_controller.rb
@@ -1,8 +1,7 @@
 module Api
   module V1
     class OrganisationsUsersController < Api::V1::ApiController
-      load_and_authorize_resource :organisations_user, parent: false
-
+      authorize_resource :organisations_user, parent: false
       resource_description do
         short "Get Organisations Users."
         formats ["json"]
@@ -27,8 +26,8 @@ module Api
       api :POST, "/v1/organisations_user", "Create a package"
       param_group :organisations_user
       def create
-        builder = OrganisationsUserBuilder.new(@organisations_user).build
-        if builder['result'] == true
+        builder = OrganisationsUserBuilder.new(params['organisations_user']).build
+        if builder[:result]
           render json: organisation_user, serializer: serializer, status: 201
         else
           render_error(builder['errors'])
--- a/app/controllers/api/v1/organisations_users_controller.rb
+++ b/app/controllers/api/v1/organisations_users_controller.rb
@@ -27,7 +27,12 @@ module Api
       api :POST, "/v1/organisations_user", "Create a package"
       param_group :organisations_user
       def create
-        save_and_render_object_with_errors(@organisations_user.create_or_update_existing_organisation_user)
+        organisation_user = OrganisationsUserBuilder.new(@organisations_user).build
+        if organisation_user.valid?
+          render json: organisation_user, serializer: serializer, status: 201
+        else
+          render_error(organisation_user.errors.full_messages.join('. '))
+        end
       end
 
       private
--- a/app/controllers/api/v1/organisations_users_controller.rb
+++ b/app/controllers/api/v1/organisations_users_controller.rb
@@ -27,8 +27,8 @@ module Api
       api :POST, "/v1/organisations_user", "Create a package"
       param_group :organisations_user
       def create
-        org_user = OrganisationsUser.organisation_user?(@organisations_user)
-        save_and_render_object_with_errors(org_user)
+        organisation_user = OrganisationsUser.create_or_update_existing_user_object(@organisations_user)
+        save_and_render_object_with_errors(organisation_user)
       end
 
       private
--- a/app/controllers/api/v1/organisations_users_controller.rb
+++ b/app/controllers/api/v1/organisations_users_controller.rb
@@ -28,7 +28,7 @@ module Api
       def create
         builder = OrganisationsUserBuilder.new(params['organisations_user']).build
         if builder[:result]
-          render json: organisation_user, serializer: serializer, status: 201
+          save_and_render_object_with_errors(builder[:organisations_user])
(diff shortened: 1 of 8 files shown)


app/models/organisations_user.rb
```diff
@@ -6,6 +6,27 @@ class OrganisationsUser < ActiveRecord::Base
...
+    if existing_user && already_exist_in_same_organisation?(existing_user)
+       self
+    elsif existing_user
+       existing_user.first_name = user.first_name
+       existing_user.last_name = user.last_name
+       existing_user.email = user.email
+       self.user_id = existing_user.id
+       existing_user.save
+       existing_user
+     else
```
namrataukirde: Alignment is not proper.

namrataukirde: Also there is scope to refactor this code.

spec/controllers/api/v1/organisations_user_controller_spec.rb
```diff
@@ -65,5 +68,24 @@
       expect(response.status).to eq(422)
       expect(subject["errors"]).to eq("User email is invalid")
     end
+
+    it "add organisation user with existing mobile from User Model" do
```
namrataukirde: We should replace `add` word with `adds`

namrataukirde: Also, I think we need to check other data as well like what happens in case of first_name, last_name. etc.  Let me know your thoughts.

spec/controllers/api/v1/organisations_user_controller_spec.rb
```diff
@@ -65,5 +68,24 @@
...
+
+    it "do not add organisation user if mobile number already exists in organisation", :show_in_doc do
+      organisations_user = create :organisations_user
+      new_organisations_user_params[:organisation_id] = organisations_user.organisation_id
+      new_organisations_user_params[:user_attributes][:mobile] = organisations_user.user.mobile
+      expect {
+        post :create, format: :json, organisations_user: new_organisations_user_params
+      }.to change(OrganisationsUser, :count).by(0)
+      expect(response.status).to eq(422)
+    end
```
namrataukirde: We can write spec for scenario where user do not exists in db.

Darshan-Chauhan:
I agree with @namrataukirde 's comments

spec/controllers/api/v1/organisations_user_controller_spec.rb
abulsayyad123: You mean different test case without first_name and last_name ?

app/models/organisations_user.rb
```diff
@@ -6,6 +6,22 @@ class OrganisationsUser < ActiveRecord::Base
 
   after_create :send_welcome_msg, :create_user_role
 
+  def create_or_update_existing_organisation_user
+    existing_user = User.find_by_mobile(user.mobile)
```
steveyken: The function name `already_exist_in_same_organisation?` is the reverse of the logic inside it so I think there is a little confusion there.

Overall, I think this code lends itself nicely to the Builder pattern where the logic of creating an object is moved to it's own class that has a single purpose - to ensure the user is either found or created and then added to the organisation if it isn't already. An SMS also needs to be sent and the charity_role added to the user if they don't already have it.

Given these requirements, moving to a separate logical class keeps the building code away from the data/model layer and allows the `OrganisationsUser` class here to remain focussed on it's job - representing the `organisations_users` table. Here's an example:

```

# controller
OrganisationsUserBuilder.new(organisation_id, user_attributes).build

# new class
class OrganisationsUserBuilder

  def initialize(organisation_id, user_attributes)
      @organisation_id = organisation_id
      @user_attributes = user_attributes
  end

  def build
      mobile = user_attributes.mobile
      raise ValueError("No mobile provided") unless mobile.present?
      user = User.where(mobile: mobile).first_or_create(user_attributes)
      if user and !user_belongs_to_organisation(user)
        user.organisation_ids << @organisation_id
        user.roles << charity_role unless user.role_ids.include?(charity_role.id)
        TwilioService.new(user).send_welcome_msg
      end
  end

  private

  def user_belongs_to_organisation(user)
      user.pluck(organisation_ids).include?(@organisation_id)
  end

  def charity_role
    @charity_role ||= Role.charity.id
  end

end
```

steveyken: For reference, here's a good example of builder pattern https://bogdanvlviv.com/posts/ruby/patterns/design-patterns-in-ruby.html#builder

lib/classes/organisations_user_builder.rb
```diff
@@ -0,0 +1,33 @@
+class OrganisationsUserBuilder
+
+  def initialize(organisations_user)
```
steveyken: @abulsayyad123 what data is inside `organisations_user`? I think I'm missing something here and it would be helpful to get a sense of it. I was expecting something like `{organisation_id: 1, mobile: "+85212345678", name: "Bob Jones", email: "bob@example.com"}`

Then the mobile would be the key, whether the user exists or not? Perhaps you could clarify for me what data the ember controller posts to the API in both cases. 1) when the user already exists, and 2) when they don't exist. Thanks

spec/lib/classes/organisations_user_builder_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    it { expect(organisations_user_builder.instance_variable_get("@mobile")).to eql(user_attributes[:mobile]) }
+    it { expect(organisations_user_builder.instance_variable_get("@position")).to eql(position) }
+  end
+
+  context "build" do
+    it "adds new user if mobile does not exist and associates it with organisation" do
+      User.current_user = current_user
+      organisations_user_builder.build
+      expect(User.count).to eq(3)
+      expect(OrganisationsUser.count).to eq(1)
```
namrataukirde: Could you please try below code here?

expect{
 organisations_user_builder.build}
}.to change{User.count}.by(1).and change{OrganisationUser.count}.by(1)

app/controllers/api/v1/organisations_users_controller.rb
```diff
@@ -1,8 +1,7 @@
 module Api
   module V1
     class OrganisationsUsersController < Api::V1::ApiController
-      load_and_authorize_resource :organisations_user, parent: false
-
+      authorize_resource :organisations_user, parent: false
```
steveyken: 👍

config/locales/en.yml
```diff
@@ -64,3 +64,12 @@ en:
...
     already_dispatched: "This has been already dispatched."
+  organisations_user_builder:
+    organisation:
+      blank: "Organisation can't be blank"
+      not_found: "Organisation not found"
+    user:
+      mobile:
+        blank: "Mobile can't be blank"
+    existing_user:
+      present: "User's already exist in organisation"
```
steveyken: Could we change the English slightly here - "User already exists in this organisation". Thanks!

config/locales/zh-tw.yml
```diff
@@ -59,4 +59,13 @@ zh-tw:
...
     already_dispatched: "This has been already dispatched."
+  organisations_user_builder:
+    organisation:
+      blank: "Organisation can't be blank"
+      not_found: "Organisation not found"
+    user:
+      mobile:
+        blank: "Mobile can't be blank"
+    existing_user:
+      present: "User's already exist in organisation"
```
steveyken: same re "User already exists in this organisation".

lib/classes/organisations_user_builder.rb
```diff
@@ -0,0 +1,59 @@
...
+    @organisation_id = params[:organisation_id].to_i
+    @user_attributes = params[:user_attributes]
+    @mobile = @user_attributes[:mobile]
+    @position = params[:position]
+    fail_with_error(I18n.t('organisations_user_builder.organisation.blank')) unless @organisation_id.present?
+    fail_with_error(I18n.t('organisations_user_builder.user.mobile.blank')) unless @mobile.present?
+  end
+
+  def build
+    user = User.where(mobile: @mobile).first_or_create(@user_attributes.to_hash)
```
steveyken: I think it's probably better to move the `to_hash` up into the controller. The builder should expect a normal hash so that is can also be valid to call this code from elsewhere and not be tied to having controller params.

app/controllers/api/v1/organisations_users_controller.rb
```diff
@@ -27,7 +26,12 @@ class OrganisationsUsersController < Api::V1::ApiController
       api :POST, "/v1/organisations_user", "Create a package"
       param_group :organisations_user
       def create
-        save_and_render_object_with_errors(@organisations_user)
+        builder = OrganisationsUserBuilder.new(params['organisations_user']).build
```
steveyken: Put `to_hash` here

app/controllers/api/v1/organisations_users_controller.rb
```diff
@@ -27,7 +26,12 @@ class OrganisationsUsersController < Api::V1::ApiController
...
+        if builder[:result]
+          save_and_render_object_with_errors(builder[:organisations_user])
```
steveyken: be good to use builder['result'] - string instead of symbols... are we using `hash_with_indifferent_access`? Same for `builder[:organisations_user]` etc etc

steveyken:
Good to go. Thanks Abul

PR #551, opened 2018-09-08
GCW-2060 systemctl sidekiq

steveyken:
This PR enables capistrano to interact with the new systemd sidekiq service. The deployment user is now able to execute sidekiq systemctl commands to control the background job runner and this is now integrated into deployment so that sidekiq is reloaded after a deployment.

    sudo systemctl start sidekiq
    sudo systemctl stop sidekiq
    sudo systemctl restart sidekiq
    sudo systemctl status sidekiq


steveyken:
Closing this PR as it starts from the wrong base. Will open a new one.

PR #555, opened 2018-09-12
Gcw 2125 make gc orders searchable on item designation

namrataukirde:
Hi Team,

This PR is to make gc orders searchable while designating item from stock app.

https://jira.crossroads.org.hk/browse/GCW-2125

Please review and let me know in case of any changes.

Thanks.


namrataukirde:
Yes @steveyken good point. We need to think over it. We will do that in the coming days.

spec/controllers/api/v1/orders_controller_spec.rb
```diff
@@ -75,6 +76,30 @@
         expect(parsed_body['designations'].count).to eq(0)
         expect(parsed_body['meta']['total_pages']).to eql(0)
       end
+
+      it 'returns goodcity order if search text is non draft goodcity order with toDesignateItem params' do
+        request.headers["X-GOODCITY-APP-NAME"] = "stock.goodcity"
```
patrixr: This line could be moved to before{} on each test. not an issue, just mentioning it :)

PR #557, opened 2018-10-03
Gcw 2132 -New user can sign up feature added

abulsayyad123:
Hi Team,
This is PR for New User Sign up feature. https://jira.crossroads.org.hk/browse/GCW-2132.
And some other changes added which includes.

- Organisations User edit method added.
- Permission changed for allowing new user and existing user to sign up for browse app.
- New title field added in User Model.
- Specs added according to new changes.

Please review this PR and let me know if any change is required.

Thanks


app/controllers/api/v1/authentication_controller.rb
```diff
@@ -70,14 +70,12 @@ def send_pin
           return render_error(@mobile.errors.full_messages.join('. '))
         end
 
-        @user = User.find_by_mobile(@mobile.mobile)
-
-        if @user && @user.allowed_login?(app_name)
-          @user.send_verification_pin
-        elsif @user
+        @user = User.find_or_create_for_browse(is_browse_app?, @mobile.mobile)
```
patrixr: I think we should try to avoid having endpoints with different behaviours based on the app context (is_browse_app). It creates inconsistencies and makes it harder to test. Was talking about that recently with @steveyken

The name of the method should be explicit on what it does, "send_pin" should not secretly create a user.

I would be in favor of having a new method to the controller which does this "login or create user", without checking if it's the browse app. That way we can one day make use of it from another app without adding "or is_x_app"

Darshan-Chauhan: I agree 👍

steveyken: Yes, if we can start making the api less dependent on the calling app that will help us a lot in the future. (Good design principles). Agree re making a new endpoint. Something like `sign_up` would be a more suitable endpoint name.

app/models/user.rb
```diff
@@ -139,6 +141,14 @@ def channels
     channels
   end
 
+  def self.find_or_create_for_browse(is_browse_app, mobile)
```
steveyken: Would be great to write this without the need for `is_browse_app`. The model is currently too tightly coupled to the app type.

config/initializers/constants.rb
```diff
@@ -1,3 +1,4 @@
+TITLE_OPTIONS = %w(Mr Mrs Miss Ms)
```
steveyken: How will we do translations?

db/migrate/20180924082712_add_title_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddTitleToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :title, :string
```
steveyken: re translation, we can do `title_en` and `title_zh_tw`

app/controllers/api/v1/authentication_controller.rb
namrataukirde: Yeah.. Good to add separate endpoint. Agree with it. 👍

db/migrate/20180924082712_add_title_to_users.rb
steveyken: Based on comments in GCW-2132 I'll retract this suggestion. @abulsayyad123 is correct - it doesn't really make sense to have title_en and title_zh_tw (consider first_name and last_name also)

abulsayyad123: Thanks @steveyken . I have rollback migration for those fields.

config/initializers/constants.rb
abulsayyad123: No issues.. We are not handling translation here.

lib/classes/organisations_user_builder.rb
```diff
@@ -11,11 +11,12 @@ class OrganisationsUserBuilder
   #   }
 
   def initialize(params)
-    @organisation_id = params['organisation_id'].to_i
+    @organisations_user = OrganisationsUser.find_by_id(params['id']) if params['id']
+    @organisation_id = params["organisation_id"].presence.try(:to_i)
     @user_attributes = params['user_attributes']
     @mobile = @user_attributes['mobile']
```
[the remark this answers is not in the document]
abulsayyad123: Sure... will add presence for this as well.

spec/lib/classes/organisations_user_builder_spec.rb
```diff
@@ -47,6 +54,11 @@
       }.to change{User.count}.by(1).and change{OrganisationsUser.count}.by(1)
     end
 
+    it "edit existing organisation" do
```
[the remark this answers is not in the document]
abulsayyad123: Will change the text with proper description.

lib/classes/organisations_user_builder.rb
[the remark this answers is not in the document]
namrataukirde: Thanks

spec/controllers/api/v1/authentication_controller_spec.rb
```diff
@@ -126,45 +128,38 @@
...
-      expect(User).to receive(:find_by_mobile).with(mobile).and_return(user)
-      expect(user).to_not receive(:send_verification_pin)
-      expect(controller).to receive(:app_name).and_return(BROWSE_APP)
-      post :send_pin, mobile: mobile
-      expect(response.status).to eq(401)
-      expect(parsed_body['otp_auth_key']).to eql( nil )
-      expect(parsed_body['error']).to eql( "You are not authorized." )
-    end
+    context "register_user" do
+      it 'send pin and register if donor login into Browse', :show_in_doc do
```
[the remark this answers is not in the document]
abulsayyad123: Yes it makes sense, will add the specs according to behaviour .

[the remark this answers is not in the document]
namrataukirde: Also we should add test case if new user tries to login to browse app then we create new user. So test should cover what happens in that case. :)

app/controllers/api/v1/authentication_controller.rb
```diff
@@ -167,6 +167,23 @@ def register_device
         render nothing: true, status: 204
       end
 
+      api :POST, '/v1/auth/register_user', "Search User by mobile number and create if user not found."
```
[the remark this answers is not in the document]
abulsayyad123: Yes, `register_user` doesn't seem to be explanatory according to its working. Will change it.

app/controllers/api/v1/authentication_controller.rb
```diff
@@ -183,7 +200,7 @@ def render_error(error_message)
       end
 
       def authenticated_user
-        warden.authenticated? && @user.allowed_login?(app_name)
+        warden.authenticated? && (is_browse_app? || @user.allowed_login?(app_name))
```
steveyken: Can we remove dependency on `is_browse_app?` - is there a better way to do this?

steveyken: I think we should adjust the permissions so that users are allowed to login on the browse app... rather than adding `is_browse_app` here...

app/controllers/api/v1/authentication_controller.rb
```diff
@@ -167,6 +167,23 @@ def register_device
         render nothing: true, status: 204
       end
 
+      api :POST, '/v1/auth/register_user_and_send_pin', "Search User by mobile number and create if user not found."
+      def register_user_and_send_pin
```
steveyken: I think we can use the ```signup``` endpoint here. You have ```User.find_or_create_user``` but this is doing the same as ```User.creation_with_auth``` above. The validation for mobile is handled inside the User model.

app/controllers/api/v1/authentication_controller.rb
steveyken: See ```User#allowed_login?(app_name)``` and add BROWSE_APP to the list of apps that ```return true```

app/controllers/api/v1/authentication_controller.rb
```diff
@@ -195,9 +212,9 @@ def current_user_channels
       # to successfully authenticate. This helps prevent man-in-the-middle attacks by ensuring that only this
       # client that can authenticate the OTP code with it.
       # Note: if user is nil, we generate a fake token so as to ward off unruly hackers.
-      def otp_auth_key_for(user)
-        if user.present?
-          user.most_recent_token.otp_auth_key
+      def otp_auth_key_for_user
```
steveyken: I'm not a huge fan of this change... we've swapped something explicit for something implicit... the new change relies on there being an `@user` variable whereas the previous method allowed the calling method to define it. Minor tweak but I think the previous was more flexible.

app/models/ability.rb
```diff
@@ -244,6 +245,7 @@ def public_ability
     can :index, Timeslot
     can :index, GogovanTransport
     can :index, CrossroadsTransport
+    can [:create, :show], OrganisationsUser
```
steveyken: This is a problem. We are allowing any non-logged in user to add any user to any organisation. We will need to fix this before deploying.

app/models/ability.rb
```diff
@@ -273,6 +275,7 @@ def taxonomies
     can [:index, :show], Permission
     can [:index, :show], UserRole
     can [:index, :show], CancellationReason
+    can [:index, :show], Organisation
```
steveyken: This is also a problem. It is allowing any user to view or list any organisation details. Will need to secure this to only organisations that the user belongs to.

app/models/user.rb
```diff
@@ -31,6 +31,8 @@ class User < ActiveRecord::Base
   validates :email, uniqueness: true, allow_nil: true,
     format: { with: /\A[^@\s]+@[^@\s]+\Z/ }
 
+  validates :title, :inclusion => { :in => TITLE_OPTIONS }, :allow_nil => true
```
steveyken: Good job!

app/models/user.rb
```diff
@@ -139,6 +141,10 @@ def channels
     channels
   end
 
+  def self.find_or_create_user(mobile)
```
steveyken: See previous note in controller about merging with User.creation_with_auth

app/models/ability.rb
namrataukirde: Hi @steveyken 

On browse we have new feature where if user do not exist in system then we are directly creating user with mobile and sending pin for login. After that we redirect user to account details page where we have select list for organisations and yet user is not authenticated as per role or permission in our system. Even in this case we want to show list of organisation's to user to select organisation.

That's the case where we kept it open. Hope that makes sense. Correct me if I am wrong. 

Thanks.

app/controllers/api/v1/authentication_controller.rb
namrataukirde: Hi @steveyken , 

Yup.. We also thought about it and its good suggestion. We cancelled using it due to params difference as we were only sending mobile number in params instead of user_params but I think we can send user_params and use existing sign_up endpoint. Will collaborate with @abulsayyad123 for same.

Thanks.

app/controllers/api/v1/authentication_controller.rb
namrataukirde: Makes sense. We will make it as previous one. :)

app/models/ability.rb
namrataukirde: Same case as mentioned below :( We can discuss this to clarify confusion and find better approach for this.

app/models/ability.rb
steveyken: Yes good point. However, we should make sure that it's not all the organisation details that are exposed. Just the names in the drop-down list. The permissions above are enabling access to all details.

steveyken: One solution here would be to create a `api/v1/organisations/names` endpoint. This could use an OrganisationNamesSerializer that returns just the `{id: name}` (translated of course). The ability.rb file could be adjusted so that anonymous users can access `api/v1/organisations/names.json` but NOT `api/v1/organisations.json` or `api/v1/organisations/1.json` This would leave the 'show' action available for users who are part of the organisation. Whilst it might not make a lot of difference right now, we will implemented access control shortly and it's better to have this change in now rather than forget it later.

What do you think?

app/models/ability.rb
steveyken: There is no `show` action in the controller so we can remove it here.

namrataukirde: Agree. 👍

app/models/user.rb
```diff
@@ -63,7 +63,7 @@ def self.creation_with_auth(user_params)
   end
 
   def allowed_login?(app_name)
-    if app_name == DONOR_APP
+    if SKIP_AUTH_APP_NAMES.include?(app_name)
```
steveyken: Technically, we're not skipping authentication but we are short-circuiting authorization (both AUTH prefixes so hard to distinguish between them). To make the code a bit clearer, could you remove `SKIP_AUTH_APP_NAMES` and just change to ```[DONOR_APP, BROWSE_APP].include?(app_name)``` - just so things are ultra-clear. The approach you've taken is otherwise fine.

This is a really minor point but I think it will help us going forward to see exactly what this method is doing each time we look at it rather than having to look up a constant definition in another file. Does that make sense?

app/controllers/api/v1/gc_organisations_controller.rb
```diff
@@ -24,10 +22,25 @@ def show
...
-    def serializer
+    def organisation_serializer
       Api::V1::OrganisationSerializer
     end
+
+    def organisation_name_serializer
+      Api::V1::OrganisationNamesSerializer
+    end
+
+    def find_record_and_render_json(serializer)
```
steveyken: Good work!

app/serializers/api/v1/organisation_names_serializer.rb
```diff
@@ -0,0 +1,7 @@
+module Api::V1
+  class OrganisationNamesSerializer < ApplicationSerializer
+    embed :ids, include: true
+
+    attributes :id, :name_en, :name_zh_tw
```
steveyken: Do we need to send both the English AND the Chinese name each time? Would it be better to just send the name of the organisation in the language that is required? e.g. see https://github.com/crossroads/api.goodcity/blob/master/app/serializers/api/v1/territory_serializer.rb for an example.

Are there cases where we have the chinese name but not the english name? Or vice-versa? In which case we could use COALESCE to ensure we have some values

abulsayyad123: Thats a good suggestion, I used the existing component from Stock which had both english and chinese name displayed at a time in searched list. So, created that here and displayed both English and Chinese names. Will add your suggested change.

app/models/user.rb
abulsayyad123: Sure, will add that array .

app/serializers/api/v1/organisation_names_serializer.rb
abulsayyad123: On changing name on the basis of English and Chinese will affect the Organisation model on frontend because it sends both English and Chinese names and managing conditional names in model will require lots of changes. I feel its okay to send both Chinese and english names from api.

Let me know your thoughts on this.

Thanks.

PR #558, opened 2018-10-04
GCW-2139 Beneficiary endpoint

patrixr:
A beneficiary refers to the person the goods are meant to go to.

This PR includes the following changes : 

- Creation of an IdentityType model and endpoint
This is a lookup table for the different kinds of identification papers a person can have (e.g id card, passport, refugee form)
**NOTE**: A rake task has been added to populate the different types of IDs

- Creation of a Beneficiary model and endpoint
This model contains the details of a person : name, phone, id type/number

- Addition of a 'beneficiary' field to orders

### DIFF
--- a/app/controllers/api/v1/beneficiaries_controller.rb
+++ b/app/controllers/api/v1/beneficiaries_controller.rb
@@ -30,17 +30,12 @@ module Api
 
       api :GET, '/v1/beneficiaries', "List all beneficiaries"
       def index
-        result = current_user.supervisor? ? @beneficiaries : @beneficiaries.my_beneficiaries
-        render json: result, each_serializer: serializer, status: 200
+        render json: @beneficiaries, each_serializer: serializer, status: 200
       end
 
       api :GET, '/v1/beneficiaries/1', "Get one beneficiary"
       def show
-        if !current_user.supervisor? and @beneficiary.created_by != current_user
-          render json: { error: "Forbidden." }, status: 403
-        else
-          render json: @beneficiary, serializer: serializer
-        end
+        render json: @beneficiary, serializer: serializer
       end
 
       api :POST, "/v1/beneficiaries", "Create a beneficiary"
@@ -53,12 +48,8 @@ module Api
       api :PUT, '/v1/beneficiaries/1', "Update user"
       param_group :beneficiary
       def update
-        if !current_user.supervisor? and @beneficiary.created_by != current_user
-          render json: { error: "Forbidden." }, status: 403
-        else
-          @beneficiary.update_attributes(beneficiary_params)
-          render json: @beneficiary, serializer: serializer
-        end
+        @beneficiary.update_attributes(beneficiary_params)
+        render json: @beneficiary, serializer: serializer
       end
 
       def serializer
--- a/app/controllers/api/v1/beneficiaries_controller.rb
+++ b/app/controllers/api/v1/beneficiaries_controller.rb
@@ -1,0 +1,70 @@
+module Api
+  module V1
+    class BeneficiariesController < Api::V1::ApiController
+      load_and_authorize_resource :beneficiary, parent: false
+
+      resource_description do
+        short 'Manage a list of beneficiaries.'
+        formats ['json']
+        error 401, "Unauthorized"
+        error 404, "Not Found"
+        error 422, "Validation Error"
+        error 500, "Internal Server Error"
+      end
+
+      def_param_group :beneficiary do
+        param :beneficiary, Hash, required: true do
+          param :identity_type_id, :number
+          param :identity_number, String
+          param :title, String
+          param :first_name, String
+          param :last_name, String
+          param :phone_number, String
+        end
+      end
+
+      def beneficiary_params
+        attributes = [:identity_type_id, :identity_number, :title, :first_name, :last_name,:phone_number]
+        params.require(:beneficiary).permit(attributes)
+      end
+
+      api :GET, '/v1/beneficiaries', "List all beneficiaries"
+      def index
+        result = current_user.supervisor? ? @beneficiaries : @beneficiaries.my_beneficiaries
+        render json: result, each_serializer: serializer, status: 200
+      end
+
+      api :GET, '/v1/beneficiaries/1', "Get one beneficiary"
+      def show
+        if !current_user.supervisor? and @beneficiary.created_by != current_user
+          render json: { error: "Forbidden." }, status: 403
+        else
+          render json: @beneficiary, serializer: serializer
+        end
+      end
+
+      api :POST, "/v1/beneficiaries", "Create a beneficiary"
+      param_group :beneficiary
+      def create
+        @beneficiary.created_by = current_user
+        save_and_render_object(@beneficiary)
+      end
+
+      api :PUT, '/v1/beneficiaries/1', "Update user"
+      param_group :beneficiary
+      def update
+        if !current_user.supervisor? and @beneficiary.created_by != current_user
+          render json: { error: "Forbidden." }, status: 403
+        else
+          @beneficiary.update_attributes(beneficiary_params)
+          render json: @beneficiary, serializer: serializer
+        end
+      end
+
+      def serializer
+        Api::V1::BeneficiarySerializer
+      end
+
+    end
+  end
+end
--- a/app/controllers/api/v1/beneficiaries_controller.rb
+++ b/app/controllers/api/v1/beneficiaries_controller.rb
@@ -48,8 +48,11 @@ module Api
       api :PUT, '/v1/beneficiaries/1', "Update user"
       param_group :beneficiary
       def update
-        @beneficiary.update_attributes(beneficiary_params)
-        render json: @beneficiary, serializer: serializer
+        if @beneficiary.update_attributes(beneficiary_params)
+          render json: @beneficiary, serializer: serializer
+        else
+          render json: @beneficiary.errors, status: 422
+        end
       end
 
       def serializer
--- a/app/controllers/api/v1/beneficiaries_controller.rb
+++ b/app/controllers/api/v1/beneficiaries_controller.rb
@@ -24,7 +24,7 @@ module Api
       end
 
       def beneficiary_params
-        attributes = [:identity_type_id, :identity_number, :title, :first_name, :last_name,:phone_number]
+        attributes = [:identity_type_id, :identity_number, :title, :first_name, :last_name, :phone_number]
         params.require(:beneficiary).permit(attributes)
       end
 
@@ -55,7 +55,6 @@ module Api
       def serializer
         Api::V1::BeneficiarySerializer
       end
-
     end
   end
 end
--- a/app/controllers/api/v1/orders_controller.rb
+++ b/app/controllers/api/v1/orders_controller.rb
@@ -23,6 +23,7 @@ module Api
           param :people_helped, :number
           param :detail_id, String
           param :stockit_id, String, desc: "stockit designation record id"
+          param :beneficiary_id, String
         end
       end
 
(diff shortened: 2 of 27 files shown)


app/controllers/api/v1/orders_controller.rb
```diff
@@ -119,14 +120,21 @@ def order_record
           @order.created_by = current_user
           @order.detail_type = "GoodCity"
         end
+
+        if order_params['beneficiary_attributes'] and  @order.beneficiary and @order.beneficiary.created_by == nil
```
Darshan-Chauhan: You can use `@order.beneficiary.created_by.nil?`

lib/tasks/identity_types.rake
```diff
@@ -0,0 +1,12 @@
+namespace :goodcity do
+  desc 'Initialize the different identity types'
+  task create_identity_types: :environment do
+    id_types = YAML.load_file("#{Rails.root}/db/identity_types.yml")
+    id_types.each do |index, record|
```
Darshan-Chauhan: If you don't need to user `index` argument than you can use `each_value` instead of each

patrixr:
> @patrixr , I see a lot of code-climate warnings. Can you resolve them??

@Darshan-Chauhan Fixed most code climate issues. I left some of them because they would break consistency across the code, and some, the %I arrays, I feel add little value and create confusion (probably because I'm not used to Ruby yet)

app/controllers/api/v1/beneficiaries_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+      def create
+        @beneficiary.created_by = current_user
+        save_and_render_object(@beneficiary)
+      end
+
+      api :PUT, '/v1/beneficiaries/1', "Update user"
+      param_group :beneficiary
+      def update
+        @beneficiary.update_attributes(beneficiary_params)
+        render json: @beneficiary, serializer: serializer
```
steveyken: Good to handle validation errors here.

```
if @beneficiary.update_attributes(beneficiary_params)
      render json: @beneficiary, serializer: serializer
else
      render json: @beneficiary.errors, status: 422
end
```

app/controllers/api/v1/orders_controller.rb
```diff
@@ -119,14 +120,33 @@ def order_record
...
+
         @order
       end
 
       def order_params
-        params.require(:order).permit(:stockit_id, :code, :status, :created_at,
-          :organisation_id, :stockit_contact_id, :detail_id, :detail_type, :description,
-          :state, :state_event, :stockit_organisation_id, :stockit_activity_id, :people_helped,
-          :purpose_description, purpose_ids: [], cart_package_ids: [])
+        params.require(:order).permit(
```
steveyken: Much neater, thanks!

app/models/ability.rb
```diff
@@ -79,6 +80,14 @@ def address_abilities
     can [:create, :show, :destroy], Address, addressable_type: "Contact" if can_manage_delivery_address?
   end
 
+  def beneficiary_abilities
+    can :create, Beneficiary
+    can [:create, :index, :show, :update], Beneficiary, created_by_id: @user_id
+    if can_manage_orders? || @api_user
+      can [:create, :index, :show, :update], Beneficiary
```
steveyken: For `can_manage_orders` and `api_user`, we can probably allow `:destroy` here as well.

app/models/ability.rb
```diff
@@ -79,6 +80,14 @@ def address_abilities
     can [:create, :show, :destroy], Address, addressable_type: "Contact" if can_manage_delivery_address?
   end
 
+  def beneficiary_abilities
```
steveyken: Nailed it!

app/models/beneficiary.rb
```diff
@@ -0,0 +1,6 @@
+class Beneficiary < ActiveRecord::Base
+  belongs_to :identity_type
+  belongs_to :created_by, class_name: 'User'
+
+  scope :my_beneficiaries, -> { where("created_by_id = (?)", User.current_user.try(:id)) }
```
steveyken: Do we need this anymore?

app/models/order.rb
```diff
@@ -8,6 +8,7 @@ class Order < ActiveRecord::Base
   belongs_to :stockit_contact
   belongs_to :stockit_organisation
   belongs_to :organisation
+  belongs_to :beneficiary, class_name: 'Beneficiary'
```
steveyken: No need for class name here unless it is different from the attribute name. I.e. Rails infers the `Beneficiary` class from the `beneficiary` arg to belongs_to. For the `created_by` below, class_name is set to `User` as there is no `created_by` model.

db/migrate/20180928031017_create_beneficiaries.rb
```diff
@@ -0,0 +1,17 @@
+class CreateBeneficiaries < ActiveRecord::Migration
+  def change
+    create_table :beneficiaries do |t|
+      t.references :identity_type, index: true, foreign_key: true
```
steveyken: Good use of `foreign_key`

spec/controllers/api/v1/beneficiaries_controller_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+        5.times { FactoryBot.create :beneficiary }        
+        get :index
+        expect(parsed_body['beneficiaries'].count).to eq(Beneficiary.count)
+      end
+    end
+
+    context 'When logged in as Charity user' do
+      before { generate_and_set_token(charity_user) }
+
+      it 'returns only beneficiaries created by logged in user' do
```
steveyken: good test!

spec/controllers/api/v1/beneficiaries_controller_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+        new_beneficiary = FactoryBot.create :beneficiary, created_by: charity_user
+        get :show, id: new_beneficiary.id
+        expect(response.status).to eq(200)
+        expect(parsed_body['beneficiary']['id']).to eq(new_beneficiary.id)
+        expect(parsed_body['beneficiary']['created_by_id']).to eq(charity_user.id)
+      end
+
+      it 'denies access to beneficiaries created by someone else' do
+        get :show, id: beneficiary.id
+        expect(response.status).to eq(403)
```
steveyken: :+1:

spec/controllers/api/v1/beneficiaries_controller_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+
+    end
+
+  end
+
+  describe "POST /beneficiaries" do
+
+    context 'When not logged in' do
+      it "denies creation of a beneficiary" do
+        payload = FactoryBot.build(:beneficiary).attributes.except('id', 'updated_at', 'created_at', 'created_by_id')
```
steveyken: Quick tip to DRY up the code - you could extract `payload` here up into a `let` command at the top of the file or relevant parent `context`

`let(:payload) { FactoryBot.build()... }`

and then remove all the `payload =` lines.

spec/controllers/api/v1/beneficiaries_controller_spec.rb
```diff
@@ -0,0 +1,150 @@
+require 'rails_helper'
+
+RSpec.describe Api::V1::BeneficiariesController, type: :controller do
+  let!(:beneficiary) { create :beneficiary }
```
steveyken: What happens if you change this from `let!(:beneficiary)` to `let(:beneficiary)`? Might save object creation in a few tests, then again, it might cause a test that depends on it's pre-existence, to fail. Worth a check.

spec/controllers/api/v1/beneficiaries_controller_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+      before { generate_and_set_token(supervisor) }
+
+      it "allows a supervisor to modify any beneficiary" do
+        new_beneficiary = FactoryBot.create(:beneficiary, created_by: charity_user)
+        put :update, id: new_beneficiary.id, beneficiary: { first_name: 'elvis' }
+        expect(response.status).to eq(200)
+        expect(parsed_body['beneficiary']['created_by_id']).to eq(charity_user.id)
+        expect(parsed_body['beneficiary']['id']).to eq(new_beneficiary.id)
+        expect(parsed_body['beneficiary']['first_name']).to eq('elvis')
+      end
```
steveyken: This is a really comprehensive set of tests. Great work!

steveyken:
(I don't mind if you add this to the current PR or create a new one once this is merged)

One thing I forgot to mention during the development process is related to translations for the IdentityType model. Since this is a lookup table, the standard process is to include a column for each language. There is a rails plugin we are using that returns the correct column depending on the language the request is set to.

See Territory model for reference. Our locales are 'en' and 'zh_tw'
For IdentityType model

1. Add columns for `name_en` and `name_zh_tw`
2. Remove `name` column
3. In model, add line `translates :name`

Step 3 creates a method called `name` that will return the value of `name_en` column if language is set to English and ditto for `zh_tw`. It also creates `name=` mutator methods that save to the correct column etc.

In serializers and the rest of the API code, continue to use `identity_type.name`  as before.

app/models/beneficiary.rb
patrixr: Good point

spec/controllers/api/v1/beneficiaries_controller_spec.rb
patrixr: No apparent changes when switching it

patrixr:
@steveyken Updated per your comments !

app/controllers/api/v1/orders_controller.rb
namrataukirde: Yeah, Its clear and readable. Thanks @patrixr :)

app/models/order.rb
namrataukirde: Agree with @steveyken here. We can remove `class_name` here.

app/serializers/api/v1/beneficiary_serializer.rb
```diff
@@ -0,0 +1,16 @@
...
+                :identity_type_id,
+                :identity_number,
+                :created_by_id,
+                :title,
+                :first_name,
+                :last_name,
+                :phone_number
+
+    has_one  :created_by, serializer: UserSerializer, root: :user
+    has_one  :identity_type, serializer: IdentityTypeSerializer, root: :identity_type
```
namrataukirde: Do we really need `root: :identity_type` here ? I think it will work without it as serializer name is same as what we have written in `root`.

`has_one  :created_by, serializer: UserSerializer, root: :user`

here we require root as we do not have created_by serializer  and we have UserSerializer thus `root: :user` is needed.

But for identity_type that's not case.

app/serializers/api/v1/identity_type_serializer.rb
```diff
@@ -0,0 +1,10 @@
+module Api::V1
+  class IdentityTypeSerializer < ApplicationSerializer
+    attributes :id, :name
+
+    def name__sql
+      "name_#{current_language}"
```
namrataukirde: :+1:

spec/controllers/api/v1/beneficiaries_controller_spec.rb
namrataukirde: 👌

spec/controllers/api/v1/beneficiaries_controller_spec.rb
namrataukirde: 👌

app/controllers/api/v1/orders_controller.rb
namrataukirde: How about using code: `order_params['beneficiary_attributes'] and  @order.beneficiary.try(:created_by).nil?`

CC: @Darshan-Chauhan @steveyken

namrataukirde:
PR looks good now @patrixr :+1: I think we can merge it now.

CC: @steveyken  @Darshan-Chauhan

PR #559, opened 2018-10-06
When deleting some records, delete their related records.

steveyken:
When deleting orders, we should:
* delete the related detail object
* delete any related goodcity_request records
* delete any related orders_packages

Can you see any problems with this approach?


namrataukirde:
Rebasing this branch will fix ci failure. We have already fixed that vulnerability issue and that fix is available on master.

PR #563, opened 2018-10-11
GCW-2155 filtered responses

patrixr:
Add an option to fetch orders without any associations. This is used to generate the list of orders when designating an item (performance issues)

### DIFF
--- a/app/controllers/api/v1/orders_controller.rb
+++ b/app/controllers/api/v1/orders_controller.rb
@@ -158,8 +158,7 @@ module Api
       end
 
       def select_serializer
-        is_shallow_render = params[:shallow] == 'true'
-        is_shallow_render ? shallow_serializer : serializer
+        params[:shallow] == 'true' ? shallow_serializer : serializer
       end
 
       def stockit_activity
--- a/app/controllers/api/v1/orders_controller.rb
+++ b/app/controllers/api/v1/orders_controller.rb
@@ -158,7 +158,7 @@ module Api
       end
 
       def select_serializer
-        is_shallow_render = params[:shallow] == true || params[:shallow] == 'true'
+        is_shallow_render = params[:shallow] == 'true'
         is_shallow_render ? shallow_serializer : serializer
       end
 
--- a/app/controllers/api/v1/orders_controller.rb
+++ b/app/controllers/api/v1/orders_controller.rb
@@ -98,8 +98,10 @@ module Api
       private
 
       def order_response(records)
+        is_shallow_render = params[:shallow] == true
+        selected_serializer = is_shallow_render ? shallow_serializer : serializer
         ActiveModel::ArraySerializer.new(records,
-          each_serializer: serializer,
+          each_serializer: selected_serializer,
           root: "designations",
           include_packages: true,
           include_order: false,
@@ -151,6 +153,10 @@ module Api
 
       def serializer
         Api::V1::OrderSerializer
+      end
+
+      def shallow_serializer
+        Api::V1::OrderShallowSerializer
       end
 
       def stockit_activity
--- a/app/controllers/api/v1/orders_controller.rb
+++ b/app/controllers/api/v1/orders_controller.rb
@@ -86,7 +86,7 @@ module Api
       end
 
       def my_orders
-        render json: @orders.my_orders.goodcity_orders, each_serializer: serializer,
+        render json: @orders.my_orders.goodcity_orders, each_serializer: select_serializer,
           root: "orders", include_packages: false, browse_order: true
       end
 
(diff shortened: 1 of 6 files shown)


app/controllers/api/v1/orders_controller.rb
```diff
@@ -133,6 +133,15 @@ def serializer
         Api::V1::OrderSerializer
       end
 
+      def shallow_serializer
+        Api::V1::OrderShallowSerializer
+      end
+
+      def select_serializer
+        is_shallow_render = params[:shallow] == true || params[:shallow] == 'true'
```
steveyken: Rails should ensure all params are strings. So no need to test for the true bool here. For the rspec test below, send the string 'true' and all should pass.

config/database.yml
```diff
@@ -7,7 +7,7 @@ default: &default
 
 development:
   <<: *default
-  database: goodcity_server_development
+  database: from_staging
```
steveyken: oops... can revert this!

spec/controllers/api/v1/orders_controller_spec.rb
```diff
@@ -154,6 +154,13 @@
         expect(parsed_body['designations'].count).to eq(1)
         expect(parsed_body['meta']['total_pages']).to eql(1)
       end
+
+      it "should be able to fetch designations without their associations" do
+        get :index, shallow: true
```
steveyken: Try this - ```get :index, params: { shallow: 'true' }```

spec/controllers/api/v1/orders_controller_spec.rb
```diff
@@ -154,6 +154,13 @@
...
+        expect(response.status).to eq(200)
+        expect(parsed_body.keys).to eq(['meta', 'designations'])
```
steveyken: Not sure if we need to sort the keys here or not... can't remember if this is deterministic. If you get random failures, then probably the ordering is important.

config/database.yml
patrixr: whoops

patrixr: Should not have used a git gui. Sticking to console from now on

patrixr:
@steveyken I've corrected the points mentioned in your comments

app/controllers/api/v1/orders_controller.rb
```diff
@@ -153,6 +153,15 @@ def serializer
         Api::V1::OrderSerializer
       end
 
+      def shallow_serializer
+        Api::V1::OrderShallowSerializer
+      end
+
+      def select_serializer
+        is_shallow_render = params[:shallow] == 'true'
```
namrataukirde: How about using `params[:shallow] == 'true' ? shallow_serializer : serializer` ? 🤔 💭

app/serializers/api/v1/order_shallow_serializer.rb
```diff
@@ -0,0 +1,62 @@
...
+    embed :ids, include: true
+    attributes :status, :created_at, :code, :detail_type, :id, :detail_id,
+      :contact_id, :local_order_id, :organisation_id, :description, :activity,
+      :country_name, :state, :purpose_description, :created_by_id,
+      :gc_organisation_id, :processed_at, :processed_by_id, :cancelled_at, :cancelled_by_id,
+      :process_completed_at, :process_completed_by_id, :closed_at, :closed_by_id, :dispatch_started_at,
+      :dispatch_started_by_id, :submitted_at, :submitted_by_id, :people_helped, :beneficiary_id
+
+    def local_order_id
+      (object.detail_type == "LocalOrder" || object.detail_type == "StockitLocalOrder") ? object.detail_id : nil
```
namrataukirde: How about using `["LocalOrder",  "StockitLocalOrder"].include?(object.detail_type )  ? object.detail_id : nil`.

Just a thought.

CC: @steveyken @Darshan-Chauhan @abulsayyad123

app/serializers/api/v1/order_shallow_serializer.rb
```diff
@@ -0,0 +1,62 @@
...
+    def gc_organisation_id__sql
+      "organisation_id"
+    end
+
+    def organisation_id
+      object.stockit_organisation_id
+    end
+
+    def organisation_id__sql
+    "stockit_organisation_id"
```
namrataukirde: identation. Add 2 spaces.

app/controllers/api/v1/orders_controller.rb
patrixr: Personal preference. Creating the variable improves readability.

app/serializers/api/v1/order_shallow_serializer.rb
patrixr: A bit better looking. We can make cosmetic changes later since this code has not been modified in this PR and we want to roll this out ASAP for testing

app/controllers/api/v1/orders_controller.rb
namrataukirde: I feel its unnecessary variable assignment.

PR #564, opened 2018-10-15
GCW-2148 Items can be searched on multiple columns

patrixr:


spec/controllers/api/v1/packages_controller_spec.rb
```diff
@@ -519,6 +519,55 @@ def test_packages_location_changes(package)
...
+      expect(subject['meta']['search']).to eql("peP")
+      expect(subject['items'].length).to eql(2)
+      expect(subject['items'][0]['designation_name']).to eql("pepper")
+      expect(subject['items'][1]['designation_name']).to eql("peppermint")
+    end
+
+    it 'should find items by state' do
+      pkg = create :package, received_quantity: 1, state: 'missing'
+      create :package, received_quantity: 1, state: 'expecting'
+      get :search_stockit_items, searchText: "missing", showQuantityItems: 'true'
```
namrataukirde: Can you remove `showQuantityItems: 'true'` from here and let me know if it works or not.

Thanks.

patrixr: All the items created for the test have a quantity of 1
I tested to make sure, working fine  :)

app/models/package.rb
```diff
@@ -71,12 +71,30 @@ def self.search(search_text, item_id, show_quantity_item = false)
       if item_id.presence
         where("item_id = ?", item_id)
       else
-        where("inventory_number ILIKE :query", query: "%#{search_text}%")
+        allowed_search_columns = [
```
steveyken: nice work!

PR #571, opened 2018-10-26
ADD: Item search filters for state and inventory_number presence

patrixr:
Matt noticed that we could search items on Stock which had not yet been received, or items which had no inventory numbers.
This PR is adding query options to filter those out in the search. (Will be added to the stock app soon)

### DIFF
--- a/app/models/package.rb
+++ b/app/models/package.rb
@@ -74,7 +74,7 @@ class Package < ActiveRecord::Base
         state = options[:state]
         queries = [search_query]
         queries.push "inventory_number IS NOT NULL" if options[:with_inventory_no]
-        queries.push "state = :state" if  !state.nil? && !state.empty?
+        queries.push "state = :state" unless state.blank?
 
         with_associations.where(
           queries.map { |q| "(#{q})" }.join(" AND "),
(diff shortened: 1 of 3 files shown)


Darshan-Chauhan:
@patrixr , can you fix the code climate warnings?

app/models/package.rb
```diff
@@ -66,30 +66,39 @@ class Package < ActiveRecord::Base
...
-          'designation_name',
-          'notes',
-          'state',
-          'locations.building',
-          'locations.area'
-        ]
+        state = options[:state]
+        queries = [ search_query ]
+        queries.push "inventory_number IS NOT NULL" if options[:with_inventory_no]
+        queries.push "state = :state" if not state.nil? and not state.empty?
```
steveyken:
```suggestion
        queries.push "state = :state" unless state.blank?
```

app/models/package.rb
```diff
@@ -66,30 +66,39 @@ class Package < ActiveRecord::Base
 
   attr_accessor :skip_set_relation_update, :request_from_admin
 
-  def self.search(search_text, item_id, show_quantity_item = false)
+  def self.search(search_text, item_id, options = {})
```
steveyken: good use of options.

app/models/package.rb
patrixr: cool, didn't know about blank

PR #573, opened 2018-11-01
Fix: charity_role added to user with no roles

abulsayyad123:
Hi Team,
This is PR for permission issue for user with no roles. Organisation User update action was not allowed for user with no roles. Issue is fixed.

https://jira.crossroads.org.hk/browse/GCW-2132?focusedCommentId=31692&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-31692

Thanks

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -214,7 +214,7 @@ class Ability
     if can_manage_organisations_users? || @api_user
       can [:create, :show, :index, :update], OrganisationsUser
     end
-
+    can [:update], OrganisationsUser, user_id: @user_id
   end
 
   def package_abilities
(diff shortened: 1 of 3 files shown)


app/models/ability.rb
```diff
@@ -214,7 +214,7 @@ def organisations_users_abilities
     if can_manage_organisations_users? || @api_user
       can [:create, :show, :index, :update], OrganisationsUser
     end
-
+    can [:update], OrganisationsUser, user_id: @user_id
```
steveyken: Good to see this scoped to @user_id. Good work @abulsayyad123

PR #579, opened 2018-11-16
Split quantity feature

Darshan-Chauhan:
Hi Team,

This PR contains following changes
- Adds package quantity splitting features
- Adds Q inventory number feature for copying items
- Turns on ability to view quantity items

Jira ticket:- https://jira.crossroads.org.hk/browse/GCW-2190

Please review and let me know if any other changes are required.
Thanks

**Note**: Test cases will be covered by @abhinav098 .

### DIFF
--- a/app/controllers/api/v1/packages_controller.rb
+++ b/app/controllers/api/v1/packages_controller.rb
@@ -131,7 +131,6 @@ module Api
           records = records.search(
             params['searchText'],
             params['itemId'],
-            :show_quantity_items => params['showQuantityItems'] == 'true',
             :state => params['state'],
             :with_inventory_no => params['withInventoryNumber'] == 'true'
           ).page(params["page"]).per(params["per_page"])
@@ -168,6 +167,16 @@ module Api
           send_stock_item_response
         else
           render json: { errors: result.errors.full_messages }, status: 422
+        end
+      end
+
+      def split_package
+        qty_to_split = package_params[:quantity].to_i
+        if @package.splittable?(qty_to_split)
+          @package.deduct_qty_and_make_copies(qty_to_split)
+          send_stock_item_response
+        else
+          render json: { errors: I18n.t('package.split_qty_error', qty: @package.quantity) }, status: 422
         end
       end
 
--- a/app/models/package.rb
+++ b/app/models/package.rb
@@ -66,7 +66,7 @@ class Package < ActiveRecord::Base
 
   attr_accessor :skip_set_relation_update, :request_from_admin
 
-  def self.search(search_text, item_id, options = {}) 
+  def self.search(search_text, item_id, options = {})
     records =
       if item_id.presence
         where("item_id = ?", item_id)
@@ -82,7 +82,6 @@ class Package < ActiveRecord::Base
           state: state
         )
       end
-    records = records.where(received_quantity: 1) unless options[:show_quantity_items]
     records
   end
 
@@ -171,6 +170,10 @@ class Package < ActiveRecord::Base
     Location.dispatch_location
   end
 
+  def splittable?(qty_to_split)
+    qty_to_split.to_i > 0 || qty_to_split.to_i < quantity
+  end
+
   def destroy_stale_packages_locations(new_quantity)
     if (singleton_package? || total_quantity_move_without_dispatch_location?(new_quantity))
       delete_associated_packages_locations
@@ -198,6 +201,13 @@ class Package < ActiveRecord::Base
     else
       create_associated_packages_location(dispatched_location.id, quantity, orders_package_id)
     end
+  end
+
+  def deduct_qty_and_make_copies(qty_to_split)
+    1..qty_to_split.times do
+      create_and_save_copy
+    end
+    update(quantity: quantity - qty_to_split)
   end
 
   def create_associated_packages_location(location_id, quantity, reference_to_orders_package = nil)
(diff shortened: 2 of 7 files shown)


app/controllers/api/v1/packages_controller.rb
```diff
@@ -251,6 +260,10 @@ def stock_serializer
         Api::V1::StockitItemSerializer
       end
 
+      def not_splittable?(qty_to_split)
+        @package.quantity.zero? || @package.singleton_package? || qty_to_split.to_i.zero? || qty_to_split >= @package.quantity || qty_to_split < 1
```
abhinav098: @Darshan-Chauhan , could you please make sure code doesnt exceeds 80 chars per line ? Doing that way, its readable and much more understandable.
Also to many conditions, can we split these conditions to seperate methods ?
Thoughts ? 

cc - @namrataukirde

Darshan-Chauhan: @abhinav098 , conditions are already separated out in different method. :)

Darshan-Chauhan: I can split it if it was written in if condition but cant split it here as it's the single line which is returning boolean

app/controllers/api/v1/packages_controller.rb
```diff
@@ -251,6 +260,10 @@ def stock_serializer
         Api::V1::StockitItemSerializer
       end
 
+      def not_splittable?(qty_to_split)
```
steveyken: I like the idea of having this method on the model so that a package can tell us if it is splittable or not. Also useful method to have outside of the controller. Maybe move to `@package.splitable?`

app/models/package.rb
```diff
@@ -200,6 +199,13 @@ def create_or_update_location_for_dispatch_from_stockit(dispatched_location, ord
     end
   end
 
+  def deduct_qty_and_make_copies(qty_to_split)
+    1..qty_to_split.times do
```
steveyken: Is it worth wrapping this whole method in an Active Record transaction? If we split the package but something goes wrong along the way, we want to ensure we are in a consistent state (i.e. creating more items but haven't deducted)

app/models/package.rb
```diff
@@ -602,6 +608,30 @@ def donor_condition_name
...
+  def create_and_save_copy
+    copy = dup
+    copy.quantity = 1
+    copy.received_quantity = 1
+    copy.inventory_number = generate_q_inventory_number
+    copy_and_save_images(copy) if copy.save
+  end
+
+  def generate_q_inventory_number
+    existing_q_inventory_number_count = Package.where("inventory_number LIKE ?", "#{inventory_number}Q%").count
```
steveyken: ordinarily this would be fine but in Stockit, we allow people to enter Q inventory numbers manually. It's possible they could add a number that isn't in sequential order and therefore we would get a naming clash.

Example:
F00001Q1
F00001Q3

If we now generate an inventory number here then it will try to make F00001Q3.

Solution: I think it would be fine to list inventory Q numbers, sorted in desc order, limit 1 and increment.

```
largest_q_number = Package.where("inventory_number LIKE ?", "#{inventory_number}Q%").order(desc).limit(1)
return largest_q_number.split('Q').last.to_i + 1
```

app/controllers/api/v1/packages_controller.rb
namrataukirde: I agree with it 👍

app/controllers/api/v1/packages_controller.rb
```diff
@@ -171,6 +170,16 @@ def designate_partial_item
         end
       end
 
+      def split_package
+        qty_to_split = package_params[:quantity].to_i
+        if not_splittable?(qty_to_split)
+          render json: { errors: I18n.t('package.split_qty_error', qty: @package.quantity) }, status: 422
```
namrataukirde: This can be done at model level. instead of adding error from here.

app/models/package.rb
```diff
@@ -82,7 +82,6 @@ def self.search(search_text, item_id, options = {})
           state: state
         )
       end
-    records = records.where(received_quantity: 1) unless options[:show_quantity_items]
```
namrataukirde: Why we removed it ?

app/models/package.rb
```diff
@@ -602,6 +608,30 @@ def donor_condition_name
 
   private
 
+  def create_and_save_copy
+    copy = dup
+    copy.quantity = 1
+    copy.received_quantity = 1
+    copy.inventory_number = generate_q_inventory_number
+    copy_and_save_images(copy) if copy.save
```
namrataukirde: How about calling `copy. copy_and_save_images    if copy.save` ?

app/models/package.rb
Darshan-Chauhan: To turn on ability to view quantity items

app/controllers/api/v1/packages_controller.rb
Darshan-Chauhan: @namrataukirde , That's how it is done for `custom error messages` on other methods so I just followed the convention.

app/models/package.rb
Darshan-Chauhan: Then it would change the `self` object in `copy_and_save_images`. So I would have to pass `self` as an argument so either way it would turn out the same.

app/controllers/api/v1/packages_controller.rb
namrataukirde: Okay, May be we can do this later when we will move code to service class. :)

app/models/package.rb
namrataukirde: Ohh Yes. 👍

Darshan-Chauhan:
@steveyken addressed your queries. 
I would like to continue with Donor build and other tickets so if you want to pick up the task of separating `split` into module than I'm okay with it. 
Can you review the changes and let me know if anything else is required, if not then I would like to merge this by the end of today.

Thanks

steveyken:
Excellent, yes go for merge and I'll refactor once it's in master.

steveyken:
Good to add some more specs too but we can do that when we modularize it

PR #580, opened 2018-11-19
Moving package split methods into their own module to encapsulate the…

steveyken:
… functionality.

@Darshan-Chauhan this isn't finished yet but wanted to give some initial first impressions from you before finalising the specs and architecture.

### DIFF
--- a/app/controllers/api/v1/packages_controller.rb
+++ b/app/controllers/api/v1/packages_controller.rb
@@ -172,8 +172,9 @@ module Api
 
       def split_package
         qty_to_split = package_params[:quantity].to_i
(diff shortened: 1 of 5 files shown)


Darshan-Chauhan:
@steveyken , this looks good. But is it possible to make it more general?
I mean this can be used to only split packages. 
May be we can generalize it so that it can be used to split any model object. 
Although doing that would be tricky as we'll have to pass fields to modify and we'll have to think about associations also...

So may be we should stick with only Package for now. 
👍

steveyken:
Yeah that's a good point. We can abstract that when we next need it.

PR #586, opened 2018-11-22
Migration to convert 'timestamp without time zone' to 'timestamp with time zone'

steveyken:
When PostgreSQL timestamp columns are defined as "timestamp without timezone" then native SQL queries such as `row_to_json` will return timestamps of the form `2018-11-17 07:45:47.431566` which do not include the timezone detail.     

Under normal Rails circumstances, this is fine. When Rails is handed the timestamp in this format, it will assume UTC and convert to the timezone as set in the `config.application.rb` file. Everything is correct and life is good.

However, we are using `postgres_ext_serializers` which returns data straight from PostgreSQL and bypasses Rails serialization. In this case, it is important that PostgreSQL returns the correct `iso8601` format with timezone information which our front-end apps can serialize correctly. Without the timezone information, JavaScript assumes browser timezone rather than UTC (which is what PostgreSQL assumes).

So `2018-11-17 07:45:47.431566` is returned by the API and Javascript serializes this as:

    new Date("2018-11-17 07:45:47.431566")
    > Sat Nov 17 2018 07:45:47 GMT+0530 (India Standard Time)` (or whatever timezone your browser is on)

Switching PostgreSQL columns to 'timestamp with time zone' fixes this. The native json functions return `2018-11-17 15:45:47.431566+08` which leads to

    new Date("2018-11-17 15:45:47.431566+08")
    > Sat Nov 17 2018 13:15:47 GMT+0530 (India Standard Time)

The change_column migration is lossless and timestamps are automatically converted. As a result, the migration is also reversible and can be run up and down without any loss of information.

After this PR is merged, timestamps will be returned with timezone information and the PostgreSQL will also return timestamp information.

**Note** as a result of this change, we need to include the `db/structure.sql` file as the default rails `schema.rb` does not capture the nuance of time zone within the `datetime` column type.

### DIFF
--- a/db/migrate/20181122111014_convert_timestamp_without_tz_to_timestamp_with_tz.rb
+++ b/db/migrate/20181122111014_convert_timestamp_without_tz_to_timestamp_with_tz.rb
@@ -1,7 +1,6 @@
 class ConvertTimestampWithoutTzToTimestampWithTz < ActiveRecord::Migration
-  def up
 
-    model_names = ['address', 'auth_token', 'beneficiary', 'booking_type', 'box',
+  MODEL_NAMES = ['address', 'auth_token', 'beneficiary', 'booking_type', 'box',
     'braintree_transaction', 'cancellation_reason', 'contact', 'country',
     'crossroads_transport', 'delivery', 'district', 'donor_condition', 'gogovan_order',
     'gogovan_transport', 'goodcity_request', 'holiday', 'identity_type', 'image',
@@ -13,7 +12,9 @@ class ConvertTimestampWithoutTzToTimestampWithTz < ActiveRecord::Migration
     'stockit_contact', 'stockit_local_order', 'stockit_organisation', 'subpackage_type',
     'subscription', 'territory', 'timeslot', 'user', 'user_role', 'version'].freeze
 
-    model_names.each do |model_name|
+  def up
+
+    MODEL_NAMES.each do |model_name|
       model = model_name.camelize.constantize
       model.columns.each do |column|
         if column.sql_type == "timestamp without time zone"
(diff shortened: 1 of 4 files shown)


db/migrate/20181122111014_convert_timestamp_without_tz_to_timestamp_with_tz.rb
```diff
@@ -0,0 +1,51 @@
...
+    'crossroads_transport', 'delivery', 'district', 'donor_condition', 'gogovan_order',
+    'gogovan_transport', 'goodcity_request', 'holiday', 'identity_type', 'image',
+    'inventory_number', 'item', 'location', 'message', 'offer', 'order', 'orders_package',
+    'orders_purpose', 'order_transport', 'organisation', 'organisations_user',
+    'organisation_type', 'package_categories_package_type', 'package_category',
+    'package', 'packages_location', 'package_type', 'pallet', 'permission', 'purpose',
+    'rejection_reason', 'role_permission', 'role', 'schedule', 'stockit_activity',
+    'stockit_contact', 'stockit_local_order', 'stockit_organisation', 'subpackage_type',
+    'subscription', 'territory', 'timeslot', 'user', 'user_role', 'version'].freeze
+
```
abhinav098: If these model_names are same, is there any way to write this array just once? 
thoughts @patrixr ?

patrixr: True, could do that, or extract them from postgres. But it's a data migration code which will run just once, so I'm okay with this :)

abhinav098: cool 👍

steveyken: Good thinking Abhinav, I had the same thought but hadn't followed through with it. I'll create a constant at the top of the migration file.

steveyken:
Merged

PR #588, opened 2018-11-26
GCW-2193 Order priority rules

patrixr:
The rules we're trying to match are described here : https://jira.crossroads.org.hk/browse/GCW-2193

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -94,17 +94,17 @@ class Order < ActiveRecord::Base
     last_6pm = now.change(hour: 18, min: 0, sec: 0)
     last_6pm -= 24.hours if now < last_6pm
 
-    case self.state
-      when "submitted" then
-        self.submitted_at && self.submitted_at <= now - 24.hours
-      when "processing" then
-        self.processed_at < last_6pm
-      when "awaiting_dispatch" then
-        self.order_transport.present? && self.order_transport.scheduled_at < DateTime.now
-      when "dispatching" then
-        self.dispatch_started_at && self.dispatch_started_at < last_6pm
-      else 
-        false
+    case state
+    when "submitted" then
+      submitted_at && submitted_at <= now - 24.hours
+    when "processing" then
+      processed_at < last_6pm
+    when "awaiting_dispatch" then
+      order_transport.present? && order_transport.scheduled_at < DateTime.now
+    when "dispatching" then
+      dispatch_started_at && dispatch_started_at < last_6pm
+    else 
+      false
     end
   end
 
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -90,7 +90,7 @@ class Order < ActiveRecord::Base
   end
 
   def is_priority?
-    now = DateTime.now.in_time_zone
+    now = Time.now.in_time_zone
     last_6pm = now.change(hour: 18, min: 0, sec: 0)
     last_6pm -= 24.hours if now < last_6pm
 
@@ -100,7 +100,7 @@ class Order < ActiveRecord::Base
     when "processing" then
       processed_at < last_6pm
     when "awaiting_dispatch" then
-      order_transport.present? && order_transport.scheduled_at < DateTime.now
+      order_transport.present? && order_transport.scheduled_at < now
     when "dispatching" then
       dispatch_started_at && dispatch_started_at < last_6pm
     else 
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -90,18 +90,16 @@ class Order < ActiveRecord::Base
   end
 
   def is_priority?
-    now = Time.now.in_time_zone
-    last_6pm = now.change(hour: 18, min: 0, sec: 0)
-    last_6pm -= 24.hours if now < last_6pm
+    last_6pm = last_end_of_work_day
 
     case state
-    when "submitted" then
-      submitted_at && submitted_at <= now - 24.hours
-    when "processing" then
+    when "submitted"
+      submitted_at && submitted_at <= Time.now - 24.hours
+    when "processing"
       processed_at < last_6pm
-    when "awaiting_dispatch" then
-      order_transport.present? && order_transport.scheduled_at < now
-    when "dispatching" then
+    when "awaiting_dispatch"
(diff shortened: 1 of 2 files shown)


app/models/order.rb
```diff
@@ -89,6 +89,25 @@ def set_initial_state
     self.state ||= :draft
   end
 
+  def is_priority?
+    now = Time.now.in_time_zone
+    last_6pm = now.change(hour: 18, min: 0, sec: 0)
+    last_6pm -= 24.hours if now < last_6pm
+
+    case state
+    when "submitted" then
```
namrataukirde: then is not needed here. we can remove all `then`.

app/models/order.rb
```diff
@@ -89,6 +89,25 @@ def set_initial_state
     self.state ||= :draft
   end
 
+  def is_priority?
+    now = Time.now.in_time_zone
+    last_6pm = now.change(hour: 18, min: 0, sec: 0)
+    last_6pm -= 24.hours if now < last_6pm
```
namrataukirde: Another way for this.

`last_pm = last_6pm.advance(days: -1) if now < last_6pm`

not necessary to implement.

app/models/order.rb
```diff
@@ -89,6 +89,25 @@ def set_initial_state
...
+
+    case state
+    when "submitted" then
+      submitted_at && submitted_at <= now - 24.hours
```
namrataukirde: Another way : 

`submitted_at <= now.advance(days: -1)`

spec/models/order_spec.rb
```diff
@@ -63,6 +63,96 @@
...
+        old_order = create :order, state: "submitted", submitted_at: Time.now - 25.hours
+        order = create :order, state: "submitted", submitted_at: Time.now - 23.hours
+        expect(old_order.is_priority?).to eq(true)
+        expect(order.is_priority?).to eq(false)
+      end
+    end
+
+    context 'An order under review (aka processing)' do
+      
+      after { allow(Time).to receive(:now).and_call_original } # un-stub
```
namrataukirde: We have `timecop` gem added. That can be used here.

https://github.com/crossroads/api.goodcity/blob/master/Gemfile#L91

Documentation: https://github.com/travisjeffery/timecop

app/models/order.rb
```diff
@@ -89,6 +89,25 @@ def set_initial_state
     self.state ||= :draft
   end
 
+  def is_priority?
```
namrataukirde: Just suggestion: This method name can be changed to `priority?` instead of `is_priority?`

namrataukirde: Cyclomatic compexity of this method is high ? Can we do something ?

spec/models/order_spec.rb
patrixr: that's pretty cool

spec/models/order_spec.rb
```diff
@@ -63,6 +63,96 @@
...
+
+      context 'If we\'re past 6pm' do
+        before { allow(Time).to receive(:now).and_return(after_6pm_today) }
+
+        it 'should be prioritised if process was started before 6pm and hasn\'t finished' do
+          order_started_before_6 = create :order, state: "processing", processed_at: before_6pm_today
```
namrataukirde: Again just a suggestion you can add trait here in order factory.

create :order, state: "processing", processed_at: before_6pm_today

As this is repeated at multiple places.

app/models/order.rb
patrixr: @namrataukirde 
It's high complexity from a code processor point of view, for the human eye it isn't. 
Trying to please code climate would just make it worse for us.
I could make a small rule engine, pseudo-code :
```javascript
const priority_rules = {
'submitted': (order) => { /*write rule */ }
};

// in the order
is_priority() {
  return priority_rules[order.state](order);
}
```
I don't think the switch case is that much of a pain, but I'm okay with changing it

patrixr: I kinda like is_priority bettrer 😢

namrataukirde: Yeah, Switch case is good here. Dont want to remove switch case here. 

If we can move last_6pm logic to some method then that can work.

But whatever we have now is more readable. So not in completely favour of moving code somewhere else.

Just wanted to know everyone's thoughts.

namrataukirde: ohh Okay, Then keep it. No worries.

app/models/order.rb
patrixr: Done

spec/models/order_spec.rb
```diff
@@ -63,6 +63,96 @@
...
+    context 'An order awaiting dispatch' do
+      let(:transport_before_6) { create :order_transport, scheduled_at: before_6pm_today, timeslot: "3PM" }
+      let(:transport_after_6) { create :order_transport, scheduled_at: after_6pm_today, timeslot: "19PM" }
+
+      before { Timecop.freeze(at_6pm_today) }
+
+      it 'should be prioritised if we\'re past it\'s planned dispatch schedule' do
+        priority_order = create :order, state: "awaiting_dispatch", order_transport: transport_before_6
+        non_priority_order = create :order, state: "awaiting_dispatch", order_transport: transport_after_6
+        expect(priority_order.is_priority?).to eq(true)
```
namrataukirde: expect(priority_order.is_priority?).to be_truthy  acn also be used.

patrixr: I like it 👍

patrixr: Although on second thought, truthy does not enforce a type. I'd rather make sure a boolean is returned.
@namrataukirde does the '?' prefix of the method enforce the type to boolean ? or is it just a convention ?

app/models/order.rb
```diff
@@ -336,4 +353,11 @@ def order
   def offer
     nil
   end
+
+  def last_end_of_work_day
+    now = Time.now.in_time_zone
+    last_6pm = now.change(hour: 18, min: 0, sec: 0)
```
steveyken: I think you can do Time.now.yesterday and use that as a base. But this works too.

spec/models/order_spec.rb
namrataukirde: It's just and convention.

PR #589, opened 2018-11-26
Charity user can :show an image from a published package

patrixr:


spec/models/abilities/image_abilities_spec.rb
```diff
@@ -54,6 +54,36 @@
     end
   end
 
+  context "As a Charity user" do
```
namrataukirde: I think this will be for all users not only Charity user so we might want to change description of spec. 🤔 💭

namrataukirde:
Agree with @steveyken , We need to address allow_web_publish flag on designation.

I think it gets altered on designation.

https://github.com/crossroads/api.goodcity/blob/master/app/models/orders_package.rb#L74

Specs are enough but was thinking spec context can be better something like `all users` as its not specific to charity role.

PR looks good. Thanks. :)

patrixr:
> Agree with @steveyken , We need to address allow_web_publish flag on designation.
> 
> I think it gets altered on designation.
> 
> https://github.com/crossroads/api.goodcity/blob/master/app/models/orders_package.rb#L74
> 
> Specs are enough but was thinking spec context can be better something like `all users` as its not specific to charity role.
> 
> PR looks good. Thanks. :)

If `allow_web_publish` does get set back to false, then this won't fix the actual issue we have.

patrixr:
@steveyken
I've modified the rule so that a user can see an image if it belongs to an order created by himself/herself (via package).

@namrataukirde I've kept the name of the spec after all, because it only tests for Charity users. We can't name "for all other users" unless we actually test for all those other users

PR #590, opened 2018-11-28
Created a locale switcher to ensure I18n.locale state doesn't pollute…

steveyken:
… the global spec space. There were a few cases, where running specs in random order caused locale to be switched to 'zh-tw' and kept.

TLDR;

In our rspec tests, we sometimes write `I18n.locale = 'zh-tw'` to switch into chinese and test some translations. This is fine except that when the spec is finished and the next one starts, the locale isn't reset back to 'en'. Most of our specs assume we want 'en' unless otherwise specified.

This PR ensures that we reset the locale to 'en' after every spec that switches it to another language. It does this by wrapping the locale change and expectation inside a block and resetting it afterwards.

### DIFF
--- a/spec/controllers/application_controller_spec.rb
+++ b/spec/controllers/application_controller_spec.rb
@@ -22,6 +22,10 @@ RSpec.describe ApplicationController, type: :controller do
         expect(I18n.locale).to eql(:en)
       end
 
+      after do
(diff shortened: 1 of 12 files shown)


steveyken:
This can be merged after main release has gone out

PR #592, opened 2018-11-29
GCW-2193 Order filter api

patrixr:


patrixr:
Note: Still some code-cleaning/code climate changes to apply

app/models/concerns/order_filtering.rb
```diff
@@ -0,0 +1,96 @@
...
+
+    def last_6pm
+      now = Time.now.in_time_zone
+      t = now.change(hour: 18, min: 0, sec: 0)
+      t -= 24.hours if now < t
+      t
+    end
+
+    def one_day_ago
+      Time.now - 24.hours
```
namrataukirde: It will be good to use Time.zone.now here.

CC: @steveyken

app/models/booking_type.rb
```diff
@@ -1,9 +1,12 @@
 
 class BookingType < ActiveRecord::Base
-	has_many :order_transports
+  has_many :order_transports
 
-	def self.appointment
-		BookingType.find_by(identifier: "appointment")
-	end
+  def self.appointment
+    BookingType.find_by(identifier: "appointment")
```
namrataukirde: Can be written as `find_by(identifier: "appointment")` instead of ` BookingType.find_by(identifier: "appointment")`. It is class method thus self will be class here and we can directly use find_by. :)

CC: @steveyken

app/models/booking_type.rb
```diff
@@ -1,9 +1,12 @@
...
+  end
 
-end
\ No newline at end of file
+  def self.online_order
+    BookingType.find_by(identifier: "online-order")
```
namrataukirde: same here.

app/models/concerns/order_filtering.rb
```diff
@@ -0,0 +1,96 @@
...
+    end
+
+    def other_sql
+      "lower(detail_type) NOT IN ('goodcity', 'shipment', 'stockitlocalorder')"
+    end
+
+    # HELPERS
+
+    def last_6pm
+      now = Time.now.in_time_zone
```
namrataukirde: We can use `Time.zone.now` here.

app/models/concerns/order_filtering.rb
```diff
@@ -0,0 +1,96 @@
...
+      t = now.change(hour: 18, min: 0, sec: 0)
+      t -= 24.hours if now < t
+      t
```
namrataukirde: Can be refactored as `(now < t) ? t.days_ago(1) : t`

app/models/order.rb
```diff
@@ -41,7 +42,7 @@ class Order < ActiveRecord::Base
 
   INACTIVE_STATES = ['cancelled', 'closed', 'draft'].freeze
 
-  scope :non_draft_orders, -> { where('state NOT IN (?)', 'draft') }
+  scope :non_draft_orders, -> { where.not("state = 'draft' AND detail_type = 'GoodCity'") }
```
namrataukirde: Can be renamed as `non_draft_goodcity_orders`

app/controllers/api/v1/orders_controller.rb
```diff
@@ -135,6 +130,22 @@ def order_record
...
+        records.filter(states: states, types: types, priority: priority)
+      end
+
+      def array_param(key)
+        params.fetch(key, "").strip.split(',')
+      end
+
+      def bool_param(key, default)
+        return default if params[key].nil?
+        params[key].to_s == "true"
```
namrataukirde: How about writing `params[key].nil? ? default : params[key].to_s == "true"`

app/models/concerns/order_filtering.rb
```diff
@@ -0,0 +1,96 @@
...
+    #   - other
+    #
+    # To add a custom type, add an sql returning method formatted as {type_name}_sql
+    # e.g :
+    #   def shipment_sql
+    #     "detail_type = 'shipment'"
+    #   end
+    #
+    def filter(states: [], types: [], priority: false)
+      res = all
```
namrataukirde: I feel this will load all  records in memory and then it will fire where query. I feel we should remove this all. 

If this is class method then we can directly write `where("state IN (?)", states)` as it will be used by rails as `Order.where("state IN (?)", states)`

This way all records will not be loaded in memory first.

CC: @steveyken @Darshan-Chauhan @abulsayyad123

app/models/concerns/order_filtering.rb
```diff
@@ -0,0 +1,96 @@
...
+      res = res.where("state IN (?)", states) unless states.empty?
+      res = res.where_types(types) unless types.empty?
+      res = res.priority if priority.present?
```
namrataukirde: This can be written as res.priority.presence. This way be can avoid 1 extra check.

app/models/concerns/order_filtering.rb
```diff
@@ -0,0 +1,96 @@
...
+  def is_priority?
+    self.class.priority.exists?(self)
+  end
+
+  module ClassMethods
+    def priority
+      join_order_transports.where <<-SQL
+        (state = 'submitted' AND submitted_at::timestamptz <= timestamptz '#{one_day_ago}') OR
+        (state = 'processing' AND processed_at::timestamptz < timestamptz '#{last_6pm}') OR
+        (state = 'awaiting_dispatch' AND order_transports.scheduled_at::timestamptz < timestamptz '#{Time.now}') OR
```
namrataukirde: Time.zone.now will be good here.

app/models/concerns/order_filtering.rb
patrixr: @namrataukirde 
The point here is to do conditional chaining, we just need `res` to be a reference to the current relation on which we can apply the the 3 possible sql queries below.
basically it is possible that we call this method with no arguments, in which case it returns the unmodified relation

If `all` really does run the query then we should change it, maybe to an empty `where()` ?
The question becomes : How to get the current relation ?

namrataukirde: https://github.com/crossroads/api.goodcity/commit/6cbcac939f2cf451576df626d60a0f958c9c45bf

PR #593, opened 2018-11-30
fetch booking types and calender date and timeslots for book appointment

namrataukirde:


db/permissions_roles.yml
```diff
@@ -60,6 +60,7 @@ api-write: []
 Charity:
 - can_login_to_browse
 - can_manage_orders_packages
+- can_manage_settings
```
patrixr: @namrataukirde I don't think this is right. I think we should **only** make the /calendar endpoint available to charity users.
Good that you caught that permissions were wrong, but that allows any charity user to modify the presets for slots !

namrataukirde: Yes, Correct. I added it to make it work and later forgot to change it to only calendar. Modifying it. :+1:

PR #601, opened 2018-12-10
Api side changes for updation of district through account_details page

abhinav098:
Hi team, this PR contains the API side changes for the district updation to user's account through account details page.
Please review and comment.
Thanks.

### DIFF
--- a/lib/classes/organisations_user_builder.rb
+++ b/lib/classes/organisations_user_builder.rb
@@ -16,7 +16,7 @@ class OrganisationsUserBuilder
     @organisation_id = params["organisation_id"].presence.try(:to_i)
     @user_attributes = params['user_attributes']
     @user_address_attributes = params['user_address_attributes']
-    @district = District.find(@user_address_attributes['district_id'].presence&.to_i) rescue nil
+    @district = District.find_by_id(@user_address_attributes['district_id'].presence&.to_i)
     @mobile = @user_attributes['mobile'].presence.try(:to_s)
     @position = params['position']
     fail_with_error(I18n.t('organisations_user_builder.organisation.blank')) unless @organisation_id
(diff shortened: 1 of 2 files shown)


lib/classes/organisations_user_builder.rb
```diff
@@ -15,6 +15,8 @@ def initialize(params)
     @user = @organisations_user.user if @organisations_user
     @organisation_id = params["organisation_id"].presence.try(:to_i)
     @user_attributes = params['user_attributes']
+    @user_address_attributes = params['user_address_attributes']
+    @district = District.find(@user_address_attributes['district_id'].presence&.to_i) rescue nil
```
namrataukirde: We can use find_by_id here and remove rescue nil.

lib/classes/organisations_user_builder.rb
```diff
@@ -48,10 +50,14 @@ def organisation
   end
 
   def update_user
+    update_user_address if @user_address_attributes['district_id'] && @district
     @user.roles << charity_role unless @user.roles.include?(charity_role)
     @user.update(@user_attributes)
   end
 
+  def update_user_address
+    @user.address.update(district_id: @user_address_attributes['district_id'])
```
namrataukirde: We need to verify scenario where  user do not have existing address record.

lib/classes/organisations_user_builder.rb
abhinav098: done already

lib/classes/organisations_user_builder.rb
abhinav098: Hi @namrataukirde , Address gets created as soon as user is created, but however, i have added an edge case if user doesn't have address, in that case we'll create the address with the same params, coming from Client.

PR #606, opened 2018-12-12
GCW 2216- Stock Dashboard orders count

abulsayyad123:
Hi Team, 
This is PR for jira ticket GCW-2216 - Stock Dashboard orders count and latest orders. https://jira.crossroads.org.hk/browse/GCW-2216. 
This PR includes:
* Orders count of Priority and Non-priority type.
* Specs for the scenarios.
Thanks


app/models/order.rb
```diff
@@ -297,15 +309,11 @@ def self.join_order_associations
...
-    active_orders
-    .select("DISTINCT ON (orders.id) orders.id AS key,  versions.created_at AS recently_used_at").
-    joins("INNER JOIN versions ON ((object_changes -> 'order_id' ->> 1) = CAST(orders.id AS TEXT))").
-    joins("INNER JOIN packages ON (packages.id = versions.item_id AND versions.item_type = 'Package')").
-    where(" versions.event = 'update' AND
-      (object_changes ->> 'order_id') IS NOT NULL AND
-      CAST(whodunnit AS integer) = ? AND
-      versions.created_at >= ? ", user_id, 15.days.ago).
-    order("key, recently_used_at DESC")
+    active_orders.goodcity_orders.joins(:versions).where(versions: {id: Version.select(:id).where(whodunnit: user_id.to_s, item_type: "Order")}).distinct.order(updated_at: :desc).limit(5)
```
steveyken: this line is too long and hard to read... can you shorten. also, what is the performance of this compare to the previous code?

app/models/order.rb
```diff
@@ -69,6 +69,17 @@ class Order < ActiveRecord::Base
 
   scope :goodcity_orders, -> { where(detail_type: 'GoodCity') }
 
+  def filter(opts = {})
+    query = FilterQueryBuilder.new
```
steveyken: @patrixr has moved the FilterQueryBuilder to another class file - can you rebase and update here?

app/controllers/api/v1/orders_controller.rb
```diff
@@ -83,7 +83,7 @@ def update
 
       def recent_designations
         records = Order.recently_used(User.current_user.id)
-        render json: order_response(records)
+        render json: order_response(records.to_a)
```
steveyken: do we really need `to_a` here?

app/controllers/api/v1/orders_controller.rb
```diff
@@ -96,8 +96,23 @@ def destroy
         render json: {}
       end
 
+      def summary
+        active_orders = Order.orders_group_by_state(false)
```
steveyken: I'm not a huge fan of passing boolean arguments as it's not clear what they relate to.

Could something like `Order.orders_group_by_state(priority: false)` work better? Makes it clearer what the true and false are relating too.

app/models/order.rb
abulsayyad123: Previous query was not giving expected result and it even was not working(I suspect). I tried to add change in that query itself but it was difficult to debug. That is why I rewrite the query again. I'll look into improve and shorten the query.

steveyken:
Ok, best to write tests to prove the failure and then that your solution
fixes it. TDD ftw!

abulsayyad123:
Recent designation will be another PR. As it has some edge case scenarios so will be available in latter PR along with all specs covered.

app/controllers/api/v1/orders_controller.rb
```diff
@@ -96,8 +96,23 @@ def destroy
         render json: {}
       end
 
+      def summary
+        active_orders = Order.orders_group_by_state(is_priority: false)
+        active_orders_count = orders_count(active_orders)
+
+        priority_orders = Order.orders_group_by_state(is_priority: true).transform_keys{ |key| "priority_".concat(key) }
```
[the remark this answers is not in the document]
namrataukirde: @abulsayyad123 it will be good to keep these 2 methods different. It's difficult to understand. 

e.g first method `active_orders_count` and other is `priority_active_orders_count ` . Then summary method will be like: 

`active_orders_count.merge!(priority_active_orders_count)
        render json: active_orders_count`

app/models/concerns/order_filtering.rb
```diff
@@ -3,7 +3,7 @@ module OrderFiltering
   extend ActiveSupport::Concern
 
   def is_priority?
-    self.class.priority.exists?(self)
+    self.class.priority.exists?(self.id)
```
namrataukirde: we can remove `self` here.

app/models/order.rb
```diff
@@ -308,6 +308,10 @@ def self.recently_used(user_id)
     order("key, recently_used_at DESC")
   end
 
+  def self.orders_group_by_state(is_priority: false)
```
namrataukirde: It will be good to add specs for this.

app/controllers/api/v1/orders_controller.rb
```diff
@@ -96,8 +96,23 @@ def destroy
         render json: {}
       end
 
+      def summary
```
namrataukirde: Do we already have specs for this endpoint ?

abulsayyad123: Yes, I have added specs for this endpoint.

app/models/order.rb
abulsayyad123: Added.

PR #618, opened 2019-01-07
GCW-2245 items filter api

abhinav098:
Hi team, 
This PR contains the API changes for filtering out items based on states, images, publish status. 
Please review and comment.
Thanks

### DIFF
--- a/app/models/concerns/package_filtering.rb
+++ b/app/models/concerns/package_filtering.rb
@@ -7,7 +7,7 @@ module PackageFiltering
 
     def filter(states: [], location: nil)
       res = where(nil)
-      package_state = states & %w[in_stock expecting received designated dispatched]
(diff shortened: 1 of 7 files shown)


app/models/concerns/package_filtering.rb
```diff
@@ -28,7 +28,7 @@ def where_states(states)
         "(#{send(method)})"
       end
       states_sql = queries.compact.join(" OR ")
-      join_order_packages.where(states_sql)
+      join_order_packages.where(states_sql).uniq
```
Darshan-Chauhan: 👍

namrataukirde: 👍

abulsayyad123: 👍

PR #632, opened 2019-01-24
Gcw 2348 remove order type column from order

namrataukirde:
Hi Team,

This PR removes order_type column from order. 

Please review.

Thanks.


steveyken:
I think I missed some of the conversation around this. We are moving order_type somewhere else?

namrataukirde:
Hi @steveyken,

Nope.. This is currently on staging only and we moved booking_type_id from `order_transport` to `order`.

Basically we added order_type column as per ticket : 
https://jira.crossroads.org.hk/browse/GCW-2247

Which we don't need anymore as we have booking_type_id on order table now to recognise order type.

Let me know in case of any confusion.

Thanks.

PR #638, opened 2019-02-01
Added gin trigram indexes for free-text search. 

steveyken:
Queries perform x1000 faster.

Notes:
- this speeds up free text search and the paginiation (COUNT) queries
- we've removed location free text search in order to ensure the query planner always chooses to use indexes
- switching rspec test suite to use `rake db:migrate` instead of `rake db:schema:load` - the migrations contain postgresql specific SQL that isn't captured inside schema.db


steveyken:
Fast tracking this to master

steveyken:
Note: Need to fix locations search before merging to live

PR #640, opened 2019-02-01
Gcw 2370 api performance

steveyken:
Are your changes @Darshan-Chauhan is this safe to deploy to live now?


steveyken:
OK, I'll merge to live without those changes for now.

PR #644, opened 2019-02-07
Converted object and object_changes in versions table to JSONB column

steveyken:
This is the next in a series of API speed improvements.

- JSONB columns (faster and natively indexed)
- Adjusted Location.recently_used method to use new partial index.

Please note, on a versions table with 1.2million rows, this migration can take upwards of 20 seconds.

### DIFF
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -27,14 +27,27 @@ class Location < ActiveRecord::Base
     where("building || area ILIKE ?", "%#{key}%")
   end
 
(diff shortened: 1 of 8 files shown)


steveyken:
I plan to keep the main performance branch open and merge in incremental improvements to master as they are ready. That way, we can test them out and all can benefit. Next stop is speeding up the `recent_designations` call in the same manner as `recent_locations`.

PR #647, opened 2019-02-13
GCW-2335 channel setup

abhinav098:
Hello team this is a child branch for GCW-2335-basic-setup-required-for-chat, contains changes for channel creation and updating store.


app/services/push_service.rb
```diff
@@ -1,13 +1,17 @@
 class PushService
   def send_update_store(channel, app_name, data)
-    channel = Channel.add_app_name_suffix(channel, ADMIN_APP) if app_name == ADMIN_APP
+    if [ADMIN_APP, BROWSE_APP, STOCK_APP].include?(app_name)
```
steveyken: Check the code in Channel.rb but I don't think you need the `if` statement here. It's already taken care of.

app/services/push_service.rb
```diff
@@ -1,13 +1,17 @@
...
+      channel = Channel.add_app_name_suffix(channel, app_name)
+    end
     SocketioSendJob.perform_later(channel, "update_store", data.to_json, true)
   end
 
   def send_notification(channel, app_name, data)
     data[:message] = ActionView::Base.full_sanitizer.sanitize(data[:message])
     data[:date] = Time.now.to_json.tr('"','')
-    channel = Channel.add_app_name_suffix(channel, ADMIN_APP) if app_name == ADMIN_APP
+    if [ADMIN_APP, BROWSE_APP, STOCK_APP].include?(app_name)
```
steveyken: Same as above, you should just be able to send `Channel.add_app_name_suffix(channel, app_name)` as it takes care of not adding a prefix for donor app.

abhinav098: sure @steveyken thanks for pointing out, I'll check it and get back to you.

app/services/push_service.rb
abhinav098: Yeah, I've made the changes. Thanks @steveyken 
here's the commit link https://github.com/crossroads/api.goodcity/pull/647/commits/868e249bd2344523814c6429b7e46d6dc1714da8

app/services/push_service.rb
abhinav098: I've made the changes. Here's the commit link [868e249](https://github.com/crossroads/api.goodcity/commit/868e249bd2344523814c6429b7e46d6dc1714da8)

PR #649, opened 2019-02-14
Moved User#channels to Channels#user_channels

steveyken:
This improves separation of concerns as Channel names should live inside Channel and not in the User class.


abhinav098:
looks good @steveyken

PR #650, opened 2019-02-14
Refactoring channel definitions and push update modules

steveyken:
- `Channel` : refactor channel class methods to constants that can be used in the app
- `current_user_rooms`: add 'user_x_browse' channel
- `PushUpdatesFor...` - added modules to manage specific push updates for each model. This removes the logic from the model class and also away from the more generic `PushUpdates` module which was trying to be a 'one-size-fits-all' solution and is now too hard to maintain.


app/controllers/api/v1/authentication_controller.rb
```diff
@@ -186,14 +189,6 @@ def authenticated_user
         warden.authenticated? && (is_browse_app? || @user.allowed_login?(app_name))
       end
 
-      def current_user_channels
-        Channel.channels_for_user_with_app_context(current_user, app_name)
-      end
-
-      def browse_channels
-        current_user_channels << "browse"
-      end
```
steveyken: @abhinav098 a brief note here about this change. You've correctly identified the problem - that we need to ensure the browse app is subscribing to the `browse` channel. However, I've removed the `browse_channels` method because it contains implementation detail about that specific channel. Instead, that should go inside the `Channel` class and not in a controller. (Principle of Separation of Concerns.) A controller cares about receiving a request and sending a response but it shouldn't need to know anything about channels or the details of the data it is sending.

Here we move the logic that decides whether to include the `browse` channel into the Channel.rb alongside the other logic that decides when to include the `reviewer` channel or the `order_fulfilment` channel. This groups the logic together and cleans up the code.

I'm sure all of this is already clear but felt it worth highlighting in more detail as we can apply this principle in many places to our API code.

app/models/user.rb
```diff
@@ -143,14 +143,6 @@ def send_verification_pin(app_name)
     TwilioService.new(self).sms_verification_pin(app_name)
   end
 
-  def channels
-    channels = Channel.private(self)
-    channels += Channel.reviewer if reviewer?
-    channels += Channel.supervisor if supervisor?
-    channels += Channel.order_fulfilment if order_fulfilment?
-    channels
-  end
```
steveyken: @abhinav098 This is another example of separation of concerns principle. I originally wrote this code a few years ago and incorrectly placed the channel detail inside the `User` object (because we needed access to user roles). There are no other uses of channel inside the User object so it is better and cleaner to extract them out into the `Channel` class  and call the publicly exposed `user.roleX?` methods from the `Channel` class.

This ultimately consolidates the logic for defining what channels a user subscribes too into just the `Channel` class instead of being spread out between `Channel` and `User` classes. It's a small change but makes it much easier to see what is going on now.

app/services/push_service.rb
```diff
@@ -1,12 +1,17 @@
...
-  def send_update_store(channel, app_name, data)
-    channel = Channel.add_app_name_suffix(channel, app_name)
-    SocketioSendJob.perform_later(channel, "update_store", data.to_json, true)
+  def send_update_store(channels, app_name, data)
+    # TODO remove add_app_name_suffix, decide correct channels further up
+    channels = [channels].flatten.uniq
+    channels = Channel.add_app_name_suffix(channels, app_name)
+    if channels.any?
+      SocketioSendJob.perform_later(channels, "update_store", data.to_json, true)
+    end
```
steveyken: @abhinav098 just wanted to note a subtle change here. We will allow `send_update_store` to receive `channel` argument as either a string (single channel) or an array of channels. It doesn't matter which it receives as it will flatten and uniq the channels before queuing the job. Doing it here means we won't need to worry about it everywhere else we call the code.

app/models/concerns/push_updates.rb
```diff
@@ -38,52 +46,16 @@ def update_client_store(operation)
...
   end
 
+  private
+
   def browse_updates(operation)
     json = Api::V1::BrowsePackageSerializer.new(self).as_json
     data = { item: { package: json[:browse_package], items: json[:items], images: json[:images] }, operation: operation }
-    service.send_update_store(Channel.browse, DONOR_APP, data)
-  end
-
```
steveyken: @abhinav098  I've moved the methods below to a new `PushUpdatesBase` class

app/models/concerns/push_updates_for_message.rb
```diff
@@ -1,18 +1,16 @@
 #
-# Updation of client store logic for messages is extracted here
```
steveyken: I've renamed this module from `UpdateClientStoreAndSendNotification` to `PushUpdatesForMessage` - it still includes the notification methods and I think that is fine too.

app/controllers/api/v1/authentication_controller.rb
abhinav098: yeah, that makes sense @steveyken . Thanks

app/models/concerns/push_updates.rb
```diff
@@ -22,14 +30,14 @@ def update_client_store(operation)
...
     end
 
     data  = { item: data_updates(type, operation), sender: user, operation: operation }
 
-    unless offer.nil?
-      donor_channel = Channel.private(offer.created_by_id)
+    if offer.present?
+      donor_channel = Channel.private_channels_for(offer.created_by_id, DONOR_APP)
       # update donor on his offer-cancellation
       if offer.try(:cancelled?) && self == offer
```
abhinav098: I think `try` is not needed here because we're checking the presence for the offer in the 39th line. 

Thoughts?

app/models/concerns/push_updates_base.rb
```diff
@@ -0,0 +1,43 @@
...
+    serializer_name = "Api::V1::#{self.class}Serializer".constantize
+    object_key = object.keys[0].downcase.to_sym
+    serializer_name.new(serializer.object).as_json[object_key] || {}
+  end
+
+  def service
+    PushService.new
+  end
+
+end
```
abhinav098: please add an empty line here.

app/models/concerns/push_updates.rb
steveyken: yep, that's true. will update

app/models/concerns/push_updates_base.rb
steveyken: good spot

PR #651, opened 2019-02-16
WIP: Message push update sent via subscription

steveyken:
## Why?

It turns out the logic for who to send a message push update to is already outlined in `app/models/concerns/message_subscription.rb`

This logic subscribes each user to a message depending on:
- who they are (donor, admin, charity)
- who sent the message
- who owns the object the message references (offer creator, order creator)

For each message that is written, `message_subscription` creates a record linking each intended recipient. This is exactly who we want to send the message push updates too.

This PR replaces `PushUpdatesForMessage` with `PushUpdateForSubscription`. For each subscription that is created, the `message` (not subscription) data is sent to the browser. This vastly simplifies the logic of working out who receives what.


app/models/concerns/push_updates_for_subscription.rb
```diff
@@ -0,0 +1,46 @@
...
+    associations = Message.reflections.keys.map(&:to_sym)
+    Api::V1::MessageSerializer.new(self.message, { exclude: associations })
+  end
+
+  # Determine which app to send it push update to.
+  # If recipient of message is also owner of its related_object, then send to donor/browse
+  # otherwise they are performing admin actions so send to Admin/Stock.
+  # This should work for admins logged in to DONOR app who create messages on their own offers.
+  def app_name
+    klass_name = self.related_object.class.name
```
abhinav098: `related_object` is this a rails method ? or you forgot to define this method ?

steveyken: Good catch. Haven't written that method yet! I was thinking it could help us get round the lack of polymorphism in Message model. So...

```
def related_object
  offer || order
end
```

abhinav098: Sure, that's a great approach, 
But as we have included this concern in Subscription class. `self` here would be the subscription's instance. So I think this method will be needed in `subscription.rb`
Please correct me if I'm wrong @steveyken .

app/services/channel.rb
```diff
@@ -42,30 +42,22 @@ def channels_for(user, app_name)
...
     # note that donor app channel is just user_1
-
     def add_app_name_suffix(channel_name, app_name)
       [channel_name].flatten.compact.map do |channel|
-        if add_suffix?(app_name, channel)
+        if (!app_name.blank? &&
+             app_name != DONOR_APP &&
+             channel.starts_with?('user_') &&
+             ![STOCK_APP, ADMIN_APP, BROWSE_APP].any?{ |app_name| channel.ends_with?(app_name) })
           "#{channel}_#{app_name}"
```
steveyken: @abhinav098 I've merged this all in to one method because we want to delete it soon anyway. Will protect us from starting to use `add_suffix?` anywhere else in the meantime!

abhinav098: :+1:

app/models/concerns/push_updates_for_subscription.rb
steveyken: Well spotted. I switched to `self.message.related_object`

steveyken:
I'm read for this to be merged into GCW-2335-basic-setup-required-for-chat we can then do full end-to-end testing.

PR #652, opened 2019-02-18
GCW-2366 Add staff note to order

patrixr:
This is field is meant to be used for the sticky note on the logisitics tab

As seen here: https://xd.adobe.com/view/224f7314-f1c0-40b6-7d7a-929e616be6bb-547e/screen/2f427e12-2183-4a74-869b-eb3d62c8d5e8/iPhone-6-7-8-97/

### DIFF
--- a/app/controllers/api/v1/orders_controller.rb
+++ b/app/controllers/api/v1/orders_controller.rb
@@ -170,7 +170,7 @@ module Api
           :address_id,:submitted_by_id, :staff_note,
           purpose_ids: [], cart_package_ids: [],
           beneficiary_attributes: beneficiary_attributes,
-          address_attributes: address_attributes,
+          address_attributes: address_attributes
         )
       end
 
--- a/app/controllers/api/v1/orders_controller.rb
+++ b/app/controllers/api/v1/orders_controller.rb
@@ -26,6 +26,7 @@ module Api
           param :beneficiary_id, String
           param :address_id, String
           param :booking_type_id, String, desc: 'Booking type.(Online order or appointment)'
+          param :staff_note, String, desc: 'Notes for internal use'
         end
       end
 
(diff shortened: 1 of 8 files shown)


app/controllers/api/v1/orders_controller.rb
```diff
@@ -160,15 +161,16 @@ def bool_param(key, default)
...
           :detail_id, :detail_type, :description,
           :state, :cancellation_reason, :state_event,
           :stockit_organisation_id, :stockit_activity_id,
           :people_helped, :beneficiary_id, :booking_type_id, :purpose_description,
-          :address_id,:submitted_by_id, purpose_ids: [], cart_package_ids: [],
+          :address_id,:submitted_by_id, :staff_note,
+          purpose_ids: [], cart_package_ids: [],
           beneficiary_attributes: beneficiary_attributes,
-          address_attributes: address_attributes
+          address_attributes: address_attributes,
```
abhinav098: @patrixr , unnecessary comma here.

patrixr: fixed

spec/controllers/api/v1/holidays_controller_spec.rb
```diff
@@ -34,7 +34,6 @@
       it "Should not include current day if it is a holiday", :show_in_doc do
         create(:holiday, holiday: Time.now.beginning_of_day)
         get :available_dates, schedule_days: 6
-        pp response.body
```
steveyken: yay! thanks for finding that.. I'd been searching for 'puts' and 'print' etc etc but couldn't locate it!

patrixr: I could take credit for removing it, but i'm also the one who added it 🤣

PR #656, opened 2019-02-25
Processing checklist

patrixr:
There are 2 new tables:

* process_checklists  - This is just a lookup table on the different things to check which we'll configure via a settings page

* orders_process_checklists - This is a join table, if a record exist linking an order to a process_checklist row, that means it was checked. Unchecking something is simply a delete

Note: will probably modify this to forbid transition if the checklist isn't complete (need to check with Matt), there's already a `can_transition` method on the Order for that purpose.

### DIFF
--- a/app/models/process_checklist.rb
+++ b/app/models/process_checklist.rb
@@ -1,0 +1,9 @@
+class ProcessChecklist < ActiveRecord::Base
+  translates :text
+  belongs_to :booking_type
+  has_many :orders
+
+  def self.for_booking_type(bt)
+    where({ booking_type: bt })
+  end
+end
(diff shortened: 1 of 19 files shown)


app/models/process_checklist.rb
```diff
@@ -0,0 +1,9 @@
+class ProcessChecklist < ActiveRecord::Base
+  translates :text
+  belongs_to :booking_type
+  has_many :orders
```
steveyken: Do we need a `through` here in order to use the join table?

patrixr: Unsure, seems to be working 😱

PR #657, opened 2019-02-27
Gcw 2433 make mobile field optional for stock

namrataukirde:
Hi Team,

This PR contains changes for making mobile field optional while creating a user from the stock app.

Please review.

Thanks.

### DIFF
--- a/app/controllers/api/v1/organisations_users_controller.rb
+++ b/app/controllers/api/v1/organisations_users_controller.rb
@@ -29,7 +29,7 @@ module Api
       api :POST, "/v1/organisations_user", "Create a package"
       param_group :organisations_user
       def create
-        builder = OrganisationsUserBuilder.new(params['organisations_user'].to_hash).build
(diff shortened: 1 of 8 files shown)


app/controllers/api/v1/organisations_users_controller.rb
```diff
@@ -29,7 +29,7 @@ class OrganisationsUsersController < Api::V1::ApiController
       api :POST, "/v1/organisations_user", "Create a package"
       param_group :organisations_user
       def create
-        builder = OrganisationsUserBuilder.new(params['organisations_user'].to_hash).build
+        builder = OrganisationsUserBuilder.new(params['organisations_user'].to_hash).build(is_stock_app?)
```
steveyken: Rather than passing `is_stock_app?` as an argument to the `build` method, I think it would be better to pass the `app_name` to the `initializer` to make it more generic and adaptable to future needs. For example, we might need app_name in the future for a different reason.

e.g. `OrganisationsUserBuilder.new(params: params['organisations_user'].to_hash, app_name: app_name).build`

PR #658, opened 2019-02-28
Improve performance of search_stockit_items method

steveyken:
* Removed total_pages from search_stockit_it to improve performance (COUNTS are expensive).

Refactored PackageFiltering methods:
* to remove unneccessary SQL
* fixed a bug in state selections - modules do not like respond_to? methods.
* refactored specs

### DIFF
--- a/app/controllers/api/v1/packages_controller.rb
+++ b/app/controllers/api/v1/packages_controller.rb
@@ -127,16 +127,11 @@ module Api
       def search_stockit_items
         records = @packages # security
         records = records.undispatched if params["orderId"].present?
-        if params['searchText'].present?
-          records = records.search(
-            params['searchText'],
-            params['itemId'].presence,
-            with_inventory_no: params['withInventoryNumber'] == 'true'
-          )
-        end
-        # apply selected filters
-        records = apply_filters(records).page(params["page"]).per(params["per_page"])
-        pages = records.total_pages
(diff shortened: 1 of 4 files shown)


steveyken:
I better fix those specs first!

steveyken:
spec's fixed. ready for review. i'll merge tomorrow if ok

app/controllers/api/v1/packages_controller.rb
```diff
@@ -127,16 +127,11 @@ def print_barcode
...
-            params['itemId'].presence,
-            with_inventory_no: params['withInventoryNumber'] == 'true'
-          )
-        end
-        # apply selected filters
-        records = apply_filters(records).page(params["page"]).per(params["per_page"])
-        pages = records.total_pages
+        records = records.search(search_text: params['searchText'], item_id: params['itemId'],
+          with_inventory_no: params['withInventoryNumber'] == 'true') if params['searchText'].present?
+        params_for_filter = ['state', 'location'].inject({}){|h, k| h[k] = params[k] if params[k].present?; h}
```
namrataukirde: Can we use each_with_object here so that we don't need to return `h` at the end of inject.

https://blog.arkency.com/inject-vs-each-with-object/

steveyken: Thanks, I'm really pleased to learn about this method... much nicer.

PR #659, opened 2019-02-28
send sms only if to mobile number is present

namrataukirde:
Hi Team,

This PR will avoid sending SMS if the user does not have a mobile number. As we are now allowing user creation without mobile number from stock app. If we create any order on behalf of such user then it's not valid to try sending SMS to such user.

Please review.

Thanks.


abhinav098:
:+1:

PR #665, opened 2019-03-06
Speeding up serialization by swapping out PES for AMS

steveyken:
In some cases (e.g. many associations), it is faster to do serialization using AMS (active-model-serialization) instead of PES (postgres-ext-serializers). This PR switches methods to AMS that see a significant performance boost > x5.


steveyken:
If review passes, I'll merge this to master as we can reap the benefits immediately. Will add more incrementally.

namrataukirde:
Nice.. We have already used this same approach at some places. :) Thanks for change. 👍

PR #666, opened 2019-03-06
Don't format TZ strings in the serializers

steveyken:
Given work in PR #586 and `config/initializers/active_record.rb` we are now storing TZ data in the database. This means that we don't need to instruct serializers to format TZ strings anymore.

Examples:

Before
`"created_at": "2018-11-28T03:18:08.542238+00:00"`

After
`"created_at": "2018-11-28T11:18:08.542+08:00"`

Note that in both cases, javascript is able to interpret the value correctly.
```
> new Date("2018-11-28T03:18:08.542238+00:00")
   Wed Nov 28 2018 11:18:08 GMT+0800 (Hong Kong Standard Time)

> new Date("2018-11-28T11:18:08.542+08:00")
   Wed Nov 28 2018 11:18:08 GMT+0800 (Hong Kong Standard Time)
```


namrataukirde:
Nice 👌

PR #667, opened 2019-03-06
Don't send message sender's mobile or email etc in push updates

steveyken:


app/serializers/api/v1/user_serializer.rb
```diff
@@ -33,6 +33,7 @@ def include_attribute?
     end
     alias_method :include_address?, :include_attribute?
     alias_method :include_mobile?, :include_attribute?
+    alias_method :include_email?, :include_attribute?
```
namrataukirde: I was thinking of using `alias` instead of `alias_method`. But using alias_method is better than `alias`.

https://blog.bigbinary.com/2012/01/08/alias-vs-alias-method.html

Just wanted to highlight if someone thinks same way like me in future. 😄

namrataukirde:
👍

PR #676, opened 2019-03-09
Gcw 2379 remove authorised by id column from order

namrataukirde:
Hi Team,

This PR removed an authorised_by_id column from order.

Please review.

Thanks.


spec/models/order_spec.rb
```diff
@@ -100,7 +100,7 @@
       end
     end
 
-    context 'with authorised_by_id' do
+    context 'with submitted_by_id' do
```
abulsayyad123: You missed the change on Line 147 i.e ` context "without authorised_by_id" ` it should be ` context "without submitted_by_id" `.

namrataukirde: Thanks @abulsayyad123. I will correct it and will send for review to you again. :)

PR #677, opened 2019-03-11
Fetch messages by state

patrixr:
- Allow the endpoint to support multiple order_ids / item_ids / order_ids e.g `GET /messages?order_id=123,654,789`
- Allow fetching the user's message by state e.g `GET /messages?order_id=123,654,789&state=unread`

### DIFF
--- a/app/controllers/api/v1/messages_controller.rb
+++ b/app/controllers/api/v1/messages_controller.rb
@@ -29,15 +29,17 @@ module Api
       param :offer_id, String, desc: "Return messages for offer id."
       param :item_id, String, desc: "Return messages for item id."
       param :order_id, String, desc: "Return messages for order id"
+      param :state, String, desc: "Message state (unread|read) to filter on"
       def index
         @messages = @messages.where(id: params[:ids].split(",")) if params[:ids].present?
         @messages = @messages.where(offer_id: params[:offer_id].split(",")) if params[:offer_id].present?
         @messages = @messages.where(order_id: params[:order_id].split(",")) if params[:order_id].present?
         @messages = @messages.where(item_id: params[:item_id].split(",")) if params[:item_id].present?
+        @messages = @messages.with_user_read_state(User.current_user, params[:state]) if params[:state].present?
         render json: @messages, each_serializer: serializer, root: 'messages'
       end
 
-      api :GET, "/v1/messages/1", "List a message"
+      api :GET, "/v1/messages/1", "Get a message"
       def show
         render json: @message, serializer: serializer, root: 'message'
       end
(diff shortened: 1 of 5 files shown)


steveyken:
If you send params as an array then rails will automatically parse it for you. E.g. order_id[]=1&order_id[]=2.

Not sure if that is better but is standard for checkboxes etc.

patrixr:
> If you send params as an array then rails will automatically parse it for you. E.g. order_id[]=1&order_id[]=2.
> 
> Not sure if that is better but is standard for checkboxes etc.

oh nice, so I might not need this at all

patrixr:
> > If you send params as an array then rails will automatically parse it for you. E.g. order_id[]=1&order_id[]=2.
> > Not sure if that is better but is standard for checkboxes etc.
> 
> oh nice, so I might not need this at all

@steveyken  I tried that but it didn't seem to work, I'll keep the `split` for the time being and I'll try to adjust later

spec/controllers/api/v1/messages_controller_spec.rb
```diff
@@ -28,20 +32,55 @@
...
+        3.times { create :message, order: order }
+        3.times { create :message, order: order2 }
+        get :index, order_id: "#{order.id},#{order2.id}"
+        expect(subject['messages'].length).to eq(6)
+      end
+
+      it "for a certain state" do
+        3.times { create :message, offer: offer }
+        3.times { create :message, offer: offer, sender_id: user.id }
+        3.times { create :message, offer: offer2 }
```
namrataukirde: Which messages have read status here?

app/controllers/api/v1/messages_controller.rb
```diff
@@ -29,15 +29,17 @@ class MessagesController < Api::V1::ApiController
...
+      param :state, String, desc: "Message state (unread|read) to filter on"
       def index
         @messages = @messages.where(id: params[:ids].split(",")) if params[:ids].present?
-        @messages = @messages.where(offer_id: params[:offer_id]) if params[:offer_id].present?
-        @messages = @messages.where(order_id: params[:order_id]) if params[:order_id].present?
-        @messages = @messages.where(item_id: params[:item_id]) if params[:item_id].present?
+        @messages = @messages.where(offer_id: params[:offer_id].split(",")) if params[:offer_id].present?
+        @messages = @messages.where(order_id: params[:order_id].split(",")) if params[:order_id].present?
+        @messages = @messages.where(item_id: params[:item_id].split(",")) if params[:item_id].present?
+        @messages = @messages.with_user_read_state(User.current_user, params[:state]) if params[:state].present?
```
abhinav098: @patrixr, could you think of any other name please ? 
I think `with_user_read_state` is ambiguous. With user read status is implying that the `read` messages should come.
But instead we are fetching messages here on the based of state which is coming in params.

Thanks

cc- @namrataukirde

spec/controllers/api/v1/messages_controller_spec.rb
patrixr: the ones created by the user are marked as read by default

patrixr:
> left a review comment.

@abhinav098 i've renamed it

abhinav098:
👍

PR #680, opened 2019-03-12
Added delivery/schedule payload for scheduled offers.

steveyken:
If an offer is in scheduled state, then include the delivery and schedule information in the summary serializer payload.


steveyken:
@patrixr Merging to test. Ping me if any problems post-merge.

PR #681, opened 2019-03-12
March Release #2 Recent work

steveyken:


steveyken:
everyone, putting together an API release to get the benefit of recent changes. I have reviewed an think everything is backwards compatible i.e. we can release without needing to release new apps. Can you also confirm this?

@namrataukirde is it true that the switch from `authorized_by` to `submitted_by` will work fine on existing apps if we deploy this API. I think so but want to check

abhinav098:
Could you please add the description @steveyken, all the stuff which is there in this PR ?
Thanks

namrataukirde:
@steveyken , For braintree, We need donor release. Else everything will work fine with this.

CC: @abulsayyad123

PR #685, opened 2019-03-14
GCW-2480 Search user improvements

abhinav098:
Hi team, in this PR. 
For user-search, search includes email and mobile

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -2,6 +2,7 @@ class User < ActiveRecord::Base
   has_paper_trail class_name: 'Version'
   include PushUpdates
   include RollbarSpecification
+  include UserSearch
 
   has_one :address, as: :addressable, dependent: :destroy
   has_many :auth_tokens, dependent: :destroy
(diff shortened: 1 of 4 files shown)


app/models/user.rb
```diff
@@ -73,7 +73,10 @@ def self.recent_orders_created_for(user_id)
   end
 
   def self.search(searchText, role)
-    joins(:roles).where("roles.name = ?  AND first_name ILIKE ? OR last_name ILIKE ?", role, "%#{searchText}%", "%#{searchText}%").uniq
```
steveyken: I've also written a very similar method (still in progress) for `offer.search` on another branch. https://github.com/crossroads/api.goodcity/blob/offers-shallow-load/app/models/concerns/offer_search.rb

 Can you do a similar approach, particularly:
* handling case where searchText is empty
* moving into a model concern
* use LEFT OUTER JOIN (I think this avoids `uniq`?)
* Fix bug for AND and OR (unless I'm missing something here?)

abhinav098:
sure, i'll resolve the conflicts and then merge.

abhinav098:
merging this to master

PR #686, opened 2019-03-14
upgrade rails version to 4.2.11.1

namrataukirde:


steveyken:
I've released this to master so we can benefit immediately

namrataukirde:
Great.. Thanks :)

PR #692, opened 2019-03-25
Live updates for Appointment slots and Packages

patrixr:
This pull request adds live updates to : 
- appointment slots
- appointment slot presets
- packages 

#### Prevent appointment preset duplicates

Modifying the time of a preset will be rejected if that time is alredy used in a different preset

#### New push class

You'll notice I created a push class named `PushUpdatesMinimal` 
It's meant to be a much simpler version of the existing push updates, the idea is that it is a _configurable concern_ , instead of having one concern per model

Currently the only thing configurable is the channels we broadcast to.
Hopefully it'll grow to one day be robust enough to replace some of the current push update classes which are growing in complexity.

Notes:  
- The name of the class might change, I just needed to make the distinction with the current PushUpdate
- Any architectural suggestions are most welcome.

How to use :

```
class MyModel < ActiveRecord::Base
  include PushUpdatesMinimal

  after_save :push_changes # trigger callback here ...
  after_destroy :push_changes # ... and here. The concern will detect what kind of operation took place
  push_targets [ Channel::STOCK_CHANNEL ] # Here we set the target channels to send to
end
```

For more advanced cases :

```
class MyOtherModel < ActiveRecord::Base
  include PushUpdatesMinimal

  after_save :push_changes
  after_destroy :push_changes
  push_targets do |record| 
      chans  = [ Channel::STOCK_CHANNEL ]
      chans << 'foo' if record.bar
      chans
  end
end
```

#### Device ID

You'll also notice the addition of the **X-GOODCITY-DEVICE-ID** header.
We'll need to send that from the application side (already added to stock)

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -97,9 +97,12 @@ class Ability
 
   def goodcity_request_abilitites
     can :create, GoodcityRequest if can_create_goodcity_requests?
-    can [:index, :show, :update, :destroy], GoodcityRequest, created_by_id: @user_id
     if can_manage_goodcity_requests?
-      can [:create, :destroy, :update], GoodcityRequest
+      can [:index, :show, :create, :destroy, :update], GoodcityRequest
+    else
+      can [:index, :show, :update, :destroy], GoodcityRequest, GoodcityRequest.of_user(@user_id) do |r|
+        r.created_by_id == @user_id || r.order.created_by_id == @user_id
+      end
     end
   end
 
--- a/app/models/appointment_slot.rb
+++ b/app/models/appointment_slot.rb
@@ -1,7 +1,13 @@
 class AppointmentSlot < ActiveRecord::Base
+  include PushUpdatesMinimal
+
   validates :quota, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
   validate :no_duplicate, on: [:create, :update]
   before_save :clean_timestamp
+
+  after_save :push_changes
+  after_destroy :push_changes
+  push_targets [ Channel::STOCK_CHANNEL ]
 
   scope :upcoming, -> { where("timestamp >= ?", DateTime.now.beginning_of_day.utc.to_s(:db)) }
 
--- a/app/models/appointment_slot_preset.rb
+++ b/app/models/appointment_slot_preset.rb
@@ -23,7 +23,7 @@ class AppointmentSlotPreset < ActiveRecord::Base
       .count
 
     if conflicts.positive?
-      errors.add(:base, "An appointment slot already exists for this time")
+      errors.add(:base, I18n.t("appointment_slots.already_exists"))
     end
   end
 end
--- a/app/models/appointment_slot_preset.rb
+++ b/app/models/appointment_slot_preset.rb
@@ -1,8 +1,29 @@
 class AppointmentSlotPreset < ActiveRecord::Base
+  include PushUpdatesMinimal
+
+  validate :no_conflict, on: :update
+
   validates :day, numericality: { only_integer: true, greater_than: 0, less_than: 8  }
   validates :hours, numericality: { only_integer: true, greater_than_or_equal_to: 0, less_than: 24 }
   validates :minutes, numericality: { only_integer: true, greater_than_or_equal_to: 0, less_than: 60 }
   validates :quota, numericality: { only_integer: true, greater_than_or_equal_to: 1 }
 
   scope :ascending, -> { order('hours ASC').order('minutes ASC') }
+
+  after_save :push_changes
+  after_destroy :push_changes
+  push_targets [ Channel::STOCK_CHANNEL ]
+
+  private
+
+  def no_conflict
+    conflicts = AppointmentSlotPreset
+      .where({ day: self.day, hours: self.hours, minutes: self.minutes })
+      .where.not({ id: self.id })
+      .count
+
+    if conflicts > 0
+      errors.add(:base, "An appointment slot already exists for this time")
+    end
+  end
 end
--- a/app/models/appointment_slot_preset.rb
+++ b/app/models/appointment_slot_preset.rb
@@ -18,11 +18,11 @@ class AppointmentSlotPreset < ActiveRecord::Base
 
   def no_conflict
     conflicts = AppointmentSlotPreset
-      .where({ day: self.day, hours: self.hours, minutes: self.minutes })
+      .where({ day: day, hours: hours, minutes: minutes })
       .where.not({ id: self.id })
       .count
 
-    if conflicts > 0
+    if conflicts.positive?
       errors.add(:base, "An appointment slot already exists for this time")
     end
   end
--- a/app/models/concerns/push_updates_minimal.rb
+++ b/app/models/concerns/push_updates_minimal.rb
@@ -34,7 +34,7 @@ module PushUpdatesMinimal
 
   def serialize(record)
     associations = record.class.reflections.keys.map(&:to_sym)
-    "Api::V1::#{record.class}Serializer".constantize.new(record, { exclude: associations })
+    "Api::V1::#{record.class}Serializer".safe_constantize.new(record, { exclude: associations })
   end
 
   def user
--- a/app/models/concerns/push_updates_minimal.rb
+++ b/app/models/concerns/push_updates_minimal.rb
@@ -1,0 +1,42 @@
+module PushUpdatesMinimal
+  extend ActiveSupport::Concern
+
+  module ClassMethods
+    attr_reader :target_channels
+
+    private
+
+    def push_targets(channels = [])
+      @target_channels = channels
+    end
+  end
+
+  def push_changes
+    record = self
+    operation = read_operation record
+    data = { item: serialize(record), sender: user, operation: operation }
+    PushService.new.send_update_store target_channels, data
+  end
+
+  def target_channels
+    self.class.target_channels
+  end
+
+  private
+
+  def serialize(record)
+    associations = record.class.reflections.keys.map(&:to_sym)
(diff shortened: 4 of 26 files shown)


app/models/appointment_slot_preset.rb
```diff
@@ -1,8 +1,29 @@
...
+
+  after_save :push_changes
+  after_destroy :push_changes
+  push_targets [ Channel::STOCK_CHANNEL ]
+
+  private
+
+  def no_conflict
+    conflicts = AppointmentSlotPreset
+      .where({ day: self.day, hours: self.hours, minutes: self.minutes })
```
namrataukirde: Do we need to write `self` here? I think writing `day`, `hours` and `minutes` will work on object. Let me know if it's something different.

app/models/appointment_slot_preset.rb
```diff
@@ -1,8 +1,29 @@
...
+      .where.not({ id: self.id })
+      .count
+
+    if conflicts > 0
```
namrataukirde: We can use method `.positive?` here instead of `> 0`

app/models/concerns/push_updates_minimal.rb
```diff
@@ -0,0 +1,50 @@
...
+      return self.class.target_channels_func.call(self)
+    end
+    self.class.target_channels
+  end
+
+  private
+
+  def serialize(record)
+    associations = record.class.reflections.keys.map(&:to_sym)
+    "Api::V1::#{record.class}Serializer".constantize.new(record, { exclude: associations })
```
namrataukirde: Will suggest using`safe_constantize` instead of `constantize`.

https://apidock.com/rails/String/safe_constantize

app/models/ability.rb
```diff
@@ -97,9 +97,12 @@ def beneficiary_abilities
 
   def goodcity_request_abilitites
     can :create, GoodcityRequest if can_create_goodcity_requests?
-    can [:index, :show, :update, :destroy], GoodcityRequest, created_by_id: @user_id
     if can_manage_goodcity_requests?
-      can [:create, :destroy, :update], GoodcityRequest
+      can [:index, :show, :create, :destroy, :update], GoodcityRequest
+    else
+      can [:index, :show, :update, :destroy], GoodcityRequest, GoodcityRequest.of_user(@user_id) do |r|
+        r.created_by_id == @user_id || r.order.created_by_id == @user_id
```
namrataukirde: 👍

app/models/appointment_slot_preset.rb
patrixr: we can, but I'm curious what's the value 🤔

app/models/concerns/push_updates_minimal.rb
patrixr: sure, I just copied that from the other files

namrataukirde:
Good Work 👏 

@patrixr, I would like to know how device_id will be helpful here. I am asking just for my understanding :)

app/models/appointment_slot_preset.rb
patrixr: It's just a JS habit, of typing `this.` 
I do value codes that are explicit generally though, I really don't mind a few extra characters for more clarity

app/models/concerns/push_updates_minimal.rb
patrixr: I've updated it 👍

patrixr:
> Good Work 👏
> 
> @patrixr, I would like to know how device_id will be helpful here. I am asking just for my understanding :)

Basically, when an app received a push update, it can know whether the change was triggered by itself
e.g
```
if (newChanges.deviceId === this.deviceId) {
   return; // do nothing
}
```

You can check the `subscription.js` file of Stock in the PR 😄

patrixr:
> > Good Work 👏
> > @patrixr, I would like to know how device_id will be helpful here. I am asking just for my understanding :)
> 
> Basically, when an app received a push update, it can know whether the change was triggered by itself
> e.g
> 
> ```
> if (newChanges.deviceId === this.deviceId) {
>    return; // do nothing
> }
> ```
> You can check the `subscription.js` file of Stock in the PR 😄

patrixr:
whoops, clicked on `close pr` by mistake :D -- reopening it

app/models/appointment_slot_preset.rb
namrataukirde: https://www.rubydoc.info/gems/rubocop/0.16.0/Rubocop/Cop/Style/RedundantSelf

Actually writing such self is not good practice in the ruby world so just thought to suggest this change.

patrixr: Interesting, I'm not sure I agree with Rubocop though, different mindsets I guess.
I'll change it to be consistent
Thanks for sharing 😄

app/models/concerns/push_updates_minimal.rb
namrataukirde: 👍

app/models/appointment_slot_preset.rb
namrataukirde: it's just a method of numeric class.

https://ruby-doc.org/core-2.3.0/Numeric.html#method-i-positive-3F

I feel it more readable on why not to use if ruby provides it. 

Again depends on personal choice. :) 

Just thought to mention this method.

app/models/appointment_slot_preset.rb
```diff
@@ -18,11 +18,11 @@ class AppointmentSlotPreset < ActiveRecord::Base
 
   def no_conflict
     conflicts = AppointmentSlotPreset
-      .where({ day: self.day, hours: self.hours, minutes: self.minutes })
+      .where({ day: day, hours: hours, minutes: minutes })
```
namrataukirde: How about writing single where query for it?

where(''day = (?) and hours = (?) minutes= (?) and id != (?)" )

Correct me if it will be different apart from syntax?  :)

patrixr: should result in the same thing 👍 

just a preference thing, I find `where(''day = (?) and hours = (?) minutes= (?) and id != (?)" , day, hours, minutes, id)` to be a bit too long

patrixr: long live nosql 😄

app/models/appointment_slot.rb
```diff
@@ -1,8 +1,14 @@
 class AppointmentSlot < ActiveRecord::Base
+  include PushUpdatesMinimal
+
   validates :quota, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
   validate :no_duplicate, on: [:create, :update]
   before_save :clean_timestamp
 
+  after_save :push_changes
+  after_destroy :push_changes
+  push_targets [ Channel::STOCK_CHANNEL ]
```
steveyken: Do these also appear on Browse app?

app/models/appointment_slot_preset.rb
```diff
@@ -1,8 +1,29 @@
 class AppointmentSlotPreset < ActiveRecord::Base
+  include PushUpdatesMinimal
+
+  validate :no_conflict, on: :update
```
steveyken: Not on `create` as well?

app/models/appointment_slot.rb
steveyken: also, I wonder if this could be ORDER_FULFILMENT_CHANNEL as it is only related to the order part of the stock app.

app/models/appointment_slot_preset.rb
```diff
@@ -1,8 +1,29 @@
...
   validates :day, numericality: { only_integer: true, greater_than: 0, less_than: 8  }
   validates :hours, numericality: { only_integer: true, greater_than_or_equal_to: 0, less_than: 24 }
   validates :minutes, numericality: { only_integer: true, greater_than_or_equal_to: 0, less_than: 60 }
   validates :quota, numericality: { only_integer: true, greater_than_or_equal_to: 1 }
 
   scope :ascending, -> { order('hours ASC').order('minutes ASC') }
+
+  after_save :push_changes
+  after_destroy :push_changes
+  push_targets [ Channel::STOCK_CHANNEL ]
```
steveyken: I wonder if this could be ORDER_FULFILMENT_CHANNEL as it is only related to the order part of the stock app.

app/models/appointment_slot_preset.rb
patrixr: on create the slots actually get merged together, the edge case was on update only

app/models/appointment_slot.rb
patrixr: They do appear on browse, but they are fetched every time we open the schedule page, so it's not a persisted resource in the app.

patrixr: for reference: Will keep `STOCK_CHANNEL`  after a chat, as non fulfilment roles with `can_manage_settings` are allowed to modify slots

app/models/concerns/push_updates_minimal.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def user
+    Api::V1::UserSerializer.new(User.current_user, { user_summary: true })
+  end
+
+  def read_operation(record)
+    if record.destroyed?
+      return :delete
+    end
+    record.id_changed? ? :create : :update
```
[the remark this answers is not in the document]
patrixr: will update

app/models/appointment_slot_preset.rb
```diff
@@ -1,8 +1,29 @@
...
+  private
+
+  def no_conflict
+    conflicts = AppointmentSlotPreset
+      .where({ day: day, hours: hours, minutes: minutes })
+      .where.not({ id: self.id })
+      .count
+
+    if conflicts.positive?
+      errors.add(:base, "An appointment slot already exists for this time")
```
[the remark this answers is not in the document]
patrixr: 👍

PR #696, opened 2019-03-26
GCW-2432 user sign up using email

namrataukirde:


app/controllers/api/v1/authentication_controller.rb
```diff
@@ -67,20 +67,20 @@ class AuthenticationController < Api::V1::ApiController
       def send_pin
         @mobile = Mobile.new(params[:mobile])
         unless @mobile.valid?
-          return render_error(@mobile.errors.full_messages.join('. '))
+          return render_error(@mobile.errors.full_messages.join(". "))
         end
 
         @user = User.find_by_mobile(@mobile.mobile)
```
patrixr: shouldn't this now be `find_user_by_mobile_or_email` instead ?

app/models/user.rb
```diff
@@ -22,54 +22,79 @@ class User < ActiveRecord::Base
...
+  def send_pin_email
+    begin
+      SendgridService.new(self).send_pin_email
+    rescue => e
+      Rollbar.error(e, error_class: "Sendgrid Error", error_message: "Sendgrid pin email")
     end
-    user
+  end
+
+  def send_verification_pin(app_name, mobile, email)
```
patrixr: @namrataukirde question :
 
If I sign-up with my email, and later add my mobile information. Will all the pin codes switch to mobile ?
Is that an expected behaviour ? I don't really know what's the best experience in that case, just thinking out loud. Maybe the social workers will still prefer the email 🤔

app/services/sendgrid_service.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def sendgrid_email_formation(email) #base
+    SendGrid::Email.new(email: email)
+  end
+
+  private
+
+  def send_to_sendgrid?
+    Rails.env.production?
```
patrixr: Maybe enable it for staging as well ? to test

spec/models/user_spec.rb
```diff
@@ -157,40 +167,67 @@
...
         user = create(:user, mobile: mobile)
-        expect(User).to receive(:find_by_mobile).with(mobile).and_return(user)
+        user_attributes["email"] = nil
+        expect(User).to receive(:find_user_by_mobile_or_email).with(mobile, nil).and_return(user)
         expect(user).to receive(:send_verification_pin)
         User.creation_with_auth(user_attributes, DONOR_APP)
       end
     end
 
+    context "when email does exist" do
```
patrixr: Would be good to have specs to validate that the pin codes are sent through the correct medium (email or sms)

e.g
```
test "users without mobile should receive the pin code by email" do
   ...
  expect(user).to receive(:send_email_pin)
  expect(user).not_to receive(:send_sms_pin)
end
```

Three cases :

- user only has an email
- user only has a mobile
- user has both (and therefore, we decide what the expected behaviour)

I know that time is of the essence, so we can add those later  : )

namrataukirde: Nope that's not a use case. If user enters mobile number then he should receive sms and if enters email_id then will receive email

app/services/sendgrid_service.rb
namrataukirde: Yeah.. Sure.. Thanks 👍

app/controllers/api/v1/authentication_controller.rb
namrataukirde: Nope... send_pin end point is used for stock, admin and donor and not for broswe.

app/models/user.rb
namrataukirde: Whaver user uses at the time of login then mode will be considered for sending pin. Like if user has both mobile and email_id but uses mobile number for login then sms will be send. If user uses email_id for login then we will send email for pin.

app/controllers/api/v1/authentication_controller.rb
patrixr: didn't know there were 2 endpoints, which one is the browse endpoint ?

app/models/user.rb
patrixr: oh I see, you pass in the payload data to the function. got it

app/controllers/api/v1/authentication_controller.rb
namrataukirde: Browse uses sign_up as we send pin even if user is not already available in db and after that we allow user to fill account details. So that pin always needs to go through process thus sign_up endpoint.

patrixr: I see, it's a bit weird that there's more sign-in functionalities in the sign_up method than in the sign_in method. Not worth making any changes now though. It'll probably clear itself up as our scope expands : )

spec/models/user_spec.rb
patrixr: Got it 👍 

In that case we can verify the new implementation is working as expected by replacing `expect(user).to receive(:send_verification_pin)` with `expect(user).to receive(:send_email_pin)`

app/controllers/api/v1/authentication_controller.rb
namrataukirde: Yup 👍

spec/models/user_spec.rb
namrataukirde: Added this spec. :)

PR #698, opened 2019-03-29
Validate emails to block harmful domains

namrataukirde:


app/models/user.rb
```diff
@@ -34,6 +34,8 @@ class User < ActiveRecord::Base
   validates :email, uniqueness: true, allow_blank: true,
                     format: {with: /\A[^@\s]+@[^@\s]+\Z/}
 
+  validates :email, fake_email: true, :if => lambda { Rails.env.staging? }
```
patrixr: only on staging ??

namrataukirde: Yup.. Only for testing .. Right now.

we were thinking to not block fake emails for staging and we will block only for prod.

Any thoughts? Is it okay to block for both environments? 🤔

abhinav098: I think we should do it for both. As we have tested emails and we are getting them properly.

namrataukirde: Okay.. Great. adding it for both environments.

app/models/user.rb
```diff
@@ -34,6 +34,8 @@ class User < ActiveRecord::Base
   validates :email, uniqueness: true, allow_blank: true,
                     format: {with: /\A[^@\s]+@[^@\s]+\Z/}
 
+  validates :email, fake_email: true, :if => lambda { Rails.env.staging? || Rails.env.production? }
```
steveyken: I'm always wary of code that doesn't run in certain environments. In this case, we'll never catch it running in development so hard to be sure things haven't broken before pushing to staging or live etc.

I checked the blacklist. Emails ending with 'example.org' or 'goodcity.hk' would pass for testing so why don't we use those and remove the restriction.

patrixr: @steveyken We're actually seeing issues with the fake_email_validator (check slack), so that makes it a very fair point : )

abhinav098: @steveyken  @patrixr , also we can fork this gem, and add our own blacklist.

cc - @namrataukirde

abhinav098: Ideally it should only run only on production and for Staging and PROD we should have separate mailing services, so that testing doesn's spoil the reputation for PROD mailing service.
Thoughts ?
cc - @steveyken  @patrixr  @namrataukirde

PR #700, opened 2019-04-01
#March2019Release4

patrixr:
Hi team,

This PR aims at adding Appointment confirmation emails. Those are sent once the staff finishes processing an order.

Please do give it a try on Staging if you can


abhinav098:
Hi @patrixr , is it backwards compatible? 
I mean do we need to release ember app with this change?

patrixr:
> Hi @patrixr , is it backwards compatible?
> I mean do we need to release ember app with this change?

No app release is needed, this just adds a hook to the event transition, and sends a mail to the user

abhinav098:
Ohh that's great.

PR #701, opened 2019-04-01
GCW-2438 allow redesignate designated item

abulsayyad123:
Hi Team, 
This is PR for ticket GCW-2468- Allow Redesignated for designated and dispatch items. Please review the PR and let me know if any change is required.

Thanks.

### DIFF
--- a/app/controllers/api/v1/packages_controller.rb
+++ b/app/controllers/api/v1/packages_controller.rb
@@ -169,6 +169,7 @@ module Api
         designator = Designator.new(@package, params[:package])
         result = designator.designate
         if result.errors.blank?
+          designate_stockit_item(params[:package][:order_id])
           send_stock_item_response
         else
           render json: { errors: result.errors.full_messages }, status: 422
@@ -239,6 +240,7 @@ module Api
       end
 
       def send_stock_item_response
+        @package.reload
         if @package.errors.blank? && @package.valid? && @package.save
           render json: @package,
             serializer: stock_serializer,
--- a/app/controllers/api/v1/packages_controller.rb
+++ b/app/controllers/api/v1/packages_controller.rb
@@ -166,14 +166,16 @@ module Api
       end
 
       def designate_partial_item
-        if params[:package][:quantity].to_i.zero?
-          check_and_undesignate = Package.package_already_used?(params[:package], @package)
-          if check_and_undesignate
-            render json: { errors: check_and_undesignate[:errors] }, status: 422
+        designator = Designator.new(@package, params[:package])
+        if designator.designated?
+          if designator.already_designated_to_same_order?
+            render json: { errors: "Already designated to this Order" }, status: 422
             return
+          else
+            designator.undesignate
           end
         end
-        result = add_partially_designated_item
+        result = designator.designate
         if result.errors.blank?
           designate_stockit_item(params[:package][:order_id])
           send_stock_item_response
--- a/app/controllers/api/v1/packages_controller.rb
+++ b/app/controllers/api/v1/packages_controller.rb
@@ -167,7 +167,7 @@ module Api
 
       def designate_partial_item
         designator = Designator.new(@package, params[:package])
-        result = designator.designate_or_redesignate
+        result = designator.designate
         if result.errors.blank?
           send_stock_item_response
         else
--- a/app/controllers/api/v1/packages_controller.rb
+++ b/app/controllers/api/v1/packages_controller.rb
@@ -167,7 +167,7 @@ module Api
 
       def designate_partial_item
         designator = Designator.new(@package, params[:package])
-        result = designator.check_designated_and_designate_item
+        result = designator.designate_or_redesignate
         if result.errors.blank?
           send_stock_item_response
         else
@@ -265,13 +265,6 @@ module Api
 
       def stock_serializer
         Api::V1::StockitItemSerializer
-      end
-
-      def add_partially_designated_item
-        OrdersPackage.add_partially_designated_item(
-          order_id: params[:package][:order_id],
-          package_id: params[:package][:package_id],
-          quantity: params[:package][:quantity].to_i.zero? ? params[:package][:received_quantity] : params[:package][:quantity])
       end
 
       def remove_stockit_prefix(stockit_inventory_number)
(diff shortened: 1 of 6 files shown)


lib/classes/designator.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  # undesignate_package params is passed from 'undesignate_designated_package'
+  def undesignate(undesignate_package = nil)
+    packages = undesignate_package ? undesignate_package : @params
+    OrdersPackage.undesignate_partially_designated_item(packages)
+    @package.undesignate_from_stockit_order
+  end
+
+  def undesignate_designated_package
+    undesignate_package = {}
```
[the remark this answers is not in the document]
namrataukirde: We can just change method name and all will be good. :)

app/controllers/api/v1/packages_controller.rb
```diff
@@ -161,16 +160,14 @@ def designate_stockit_item(order_id)
       end
 
       def undesignate_partial_item
-        OrdersPackage.undesignate_partially_designated_item(params[:package])
-        @package.undesignate_from_stockit_order
+        designator = Designator.new(@package, params[:package])
```
abhinav098: instead of assigning it to a variable, directly use `Designator.new(@package, params[:package]).undesignate`

app/controllers/api/v1/packages_controller.rb
```diff
@@ -161,16 +160,14 @@ def designate_stockit_item(order_id)
...
         send_stock_item_response
       end
 
       def designate_partial_item
-        result = OrdersPackage.add_partially_designated_item(
-          order_id: params[:package][:order_id],
-          package_id: params[:package][:package_id],
-          quantity: params[:package][:quantity])
+        designator = Designator.new(@package, params[:package])
+        result = designator.designate
```
abhinav098: same here, this can be written as `result = Designator.new(@package, params[:package]).designate`

lib/classes/designator.rb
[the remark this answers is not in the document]
abulsayyad123: Sure.. will change it.

app/controllers/api/v1/packages_controller.rb
namrataukirde: Agree :+1:

lib/classes/designator.rb
```diff
@@ -0,0 +1,54 @@
+class Designator
+  attr_accessor :package, :params, :order_id, :orders_package
+
+  def initialize(package, package_params)
+    @package = package
+    @params = package_params
+    @order_id = package_params[:order_id].to_i
+    @orders_package = @package.orders_packages.new
+    @existing_designation =  OrdersPackage.find_by_id(@params[:orders_package_id])
```
namrataukirde: Operator `=` should be surrounded by a single space.  Just from code climate. 😉

lib/classes/designator.rb
```diff
@@ -0,0 +1,54 @@
...
+    @existing_designation
+  end
+
+  def designating_to_existing_designation?
+    @existing_designation&.order_id == @params[:order_id].to_i
+  end
+
+  def designate_item
+    @orders_package.order_id = @order_id
+    @orders_package.quantity = @params[:quantity_to_designate] ? @params[:quantity_to_designate] :  @params[:quantity]
```
namrataukirde: Operator `:` should be surrounded by a single space. 

Again from code climate and not me. 😜

patrixr: `npm install --save namrata-climate` 😄

namrataukirde: 😆

app/controllers/api/v1/packages_controller.rb
patrixr: My personal preference in those scenarios is usually to have it as a Service, and not needing an instance at all
e.g `Designator.designate(@package)`

not a big deal though, just worth sharing 😄 
the important part is that the logic was extracted to a single place, the rest is just nitpicking

app/controllers/api/v1/packages_controller.rb
abulsayyad123: Sure will do it.

lib/classes/designator.rb
```diff
@@ -0,0 +1,54 @@
...
+    @existing_designation
+  end
+
+  def designating_to_existing_designation?
+    @existing_designation&.order_id == @params[:order_id].to_i
+  end
+
+  def designate_item
+    @orders_package.order_id = @order_id
+    @orders_package.quantity = @params[:quantity_to_designate] ? @params[:quantity_to_designate] : @params[:quantity]
```
namrataukirde: whats difference in params[:quantity_to_designate] and params[:quantity] ?

We should always get only on param in this case. I don't think we need to send 2 different params from ember just for same action which is designate.

Not a blocker but will be good to change.

abulsayyad123: `quantity_to_designate` will come in params if package is designated else it will not be present in params and in that case `quantity` will value be assigned.

lib/classes/designator.rb
```diff
@@ -0,0 +1,54 @@
...
+  # undesignate_package params is passed from 'form_nested_params_for_undesignate'
+  def undesignate(undesignate_package = nil)
+    packages = undesignate_package ? undesignate_package : @params
+    OrdersPackage.undesignate_partially_designated_item(packages)
+    @package.undesignate_from_stockit_order
+  end
+
+  def form_nested_params_for_undesignate
+    undesignate_package = {}
+    @params[:quantity] = @package.received_quantity
```
namrataukirde: Can we use quantity came from ember here to form undesignate params ? Instead of using record's received quantity?

lib/classes/designator.rb
namrataukirde: But it's not good to send 2 different params for 2 scenario's for same designate action.

abulsayyad123: Adding change of only using `quantity_to_designate`.

lib/classes/designator.rb
abulsayyad123: I am sending `received_quantity` from Ember side , is it fine to use that ?

namrataukirde: Nope.

PR #704, opened 2019-04-02
GCW-2394 dont allow item dispatch if order not processed

abhinav098:
Hi team,
This PR add api side restriction for item dispatch(partial)
Thanks

### DIFF
--- a/app/controllers/api/v1/packages_controller.rb
+++ b/app/controllers/api/v1/packages_controller.rb
@@ -199,6 +199,9 @@ module Api
 
       def dispatch_stockit_item
         @orders_package = OrdersPackage.find_by(id: params[:package][:order_package_id])
+        order = @orders_package.order
+        render_order_status_error and return if order.prohibit_item_dispatch?
+
         if @orders_package.dispatch_orders_package
           @package.dispatch_stockit_item(@orders_package, params["packages_location_and_qty"], true)
           send_stock_item_response
@@ -261,6 +264,10 @@ module Api
       end
 
       private
+
+      def render_order_status_error
+        render json: { errors: I18n.t('orders_package.order_status_error') }, status: 403
+      end
 
       def stock_serializer
         Api::V1::StockitItemSerializer
--- a/app/controllers/api/v1/packages_controller.rb
+++ b/app/controllers/api/v1/packages_controller.rb
@@ -199,8 +199,7 @@ module Api
 
       def dispatch_stockit_item
         @orders_package = OrdersPackage.find_by(id: params[:package][:order_package_id])
-        order = @orders_package.order
-        render_order_status_error and return if order.prohibit_item_dispatch?
+        return render_order_status_error if @orders_package.order.can_dispatch_item?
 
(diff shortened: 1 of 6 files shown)


spec/controllers/api/v1/packages_controller_spec.rb
```diff
@@ -38,6 +38,18 @@ def test_packages_location_changes(package)
     expect(package.locations.first.building).to eq 'Dispatched'
   end
 
+  describe "Dispatching item" do
+    before { generate_and_set_token(user) }
+    let(:order) { create :order, state: Order::ORDER_NOT_PROCESSED_STATES.sample}
+    let(:orders_package) { create :orders_package, package: package, order: order }
+
+    it "throws error if order is not processed" do
```
namrataukirde: Can we check response error message here?

abhinav098: yeah sure, it can be done. I'll update changes here.

app/controllers/api/v1/packages_controller.rb
```diff
@@ -199,6 +199,9 @@ def undesignate_stockit_item
 
       def dispatch_stockit_item
         @orders_package = OrdersPackage.find_by(id: params[:package][:order_package_id])
+        order = @orders_package.order
+        render_order_status_error and return if order.prohibit_item_dispatch?
```
namrataukirde: Can't we directly use @orders_package.order.prohibit_item_dispatch? 🤔

Darshan-Chauhan: Hmmmmmmmmm

config/locales/en.yml
```diff
@@ -70,6 +70,7 @@ en:
     split_qty_error: "Quantity to split should be at least 1 and less than %{qty}"
   orders_package:
     already_dispatched: "This has been already dispatched."
+    order_status_error: "You need to complete processing Order first before dispatching."
```
steveyken: Just refining the english a bit "You need to finish processing the order before dispatching."

config/locales/zh-tw.yml
```diff
@@ -65,6 +65,7 @@ zh-tw:
     split_qty_error: "Quantity to split should be at least 1 and less than %{qty}"
   orders_package:
     already_dispatched: "This has been already dispatched."
+    order_status_error: "You need to complete processing Order first before dispatching."
```
steveyken: same here

app/controllers/api/v1/packages_controller.rb
```diff
@@ -199,6 +199,8 @@ def undesignate_stockit_item
 
       def dispatch_stockit_item
         @orders_package = OrdersPackage.find_by(id: params[:package][:order_package_id])
+        return render_order_status_error if @orders_package.order.can_dispatch_item?
```
steveyken: i think the logic is the wrong way round here `can_dispatch_item` suggests the action can proceed but instead we are throwing an error?

also, please call it `package`. We want to remove the usage of the word item as much as possible.

PR #705, opened 2019-04-03
Fixing email and mobile validations check uniqueness if values present

abhinav098:
Hello team this PR fixes validation issues for user mobile and email.
Thanks

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -29,10 +29,12 @@ class User < ActiveRecord::Base
 
   validates :mobile, format: {with: Mobile::HONGKONGMOBILEREGEXP}, unless: :request_from_stock_without_mobile?
 
-  validates :mobile, presence: true, uniqueness: true, unless: :request_from_stock_without_mobile?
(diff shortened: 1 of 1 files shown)


app/models/user.rb
```diff
@@ -29,10 +29,12 @@ class User < ActiveRecord::Base
 
   validates :mobile, format: {with: Mobile::HONGKONGMOBILEREGEXP}, unless: :request_from_stock_without_mobile?
 
-  validates :mobile, presence: true, uniqueness: true, unless: :request_from_stock_without_mobile?
+  validates :mobile, presence: true, unless: :request_from_stock_without_mobile?
+  validates :mobile, uniqueness: true, if: 'mobile.present?'
```
namrataukirde: Isn't it work without quotes around it? 🤔

patrixr: good point @namrataukirde

abhinav098: It works both ways @namrataukirde. With and without quotes too.
=> https://stackoverflow.com/a/20915135

PR #707, opened 2019-04-04
GCW-2517 add 4,7,8 in the regex for mobile

abhinav098:
This PR adds 4,7,8 to the regex for mobile validations.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -154,6 +154,10 @@ unless ENV['LIVE'] == "true"
     { mobile: "+85251111112", first_name: "Daniel", last_name: "Dell52" },
     { mobile: "+85251111113", first_name: "Dakota", last_name: "Deryn53" },
     { mobile: "+85251111114", first_name: "Delia", last_name: "Devon54" },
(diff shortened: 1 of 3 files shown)


namrataukirde:
Great.. Thanks @abhinav098 . Will be good to add some specs.

namrataukirde:
If you can add some existing users with these numbers then it will be good. You can check seeds.rb. That's not urgent though. It's just a thought.

CC: @steveyken

namrataukirde:
Foor seed :+1:

db/seeds.rb
```diff
@@ -154,6 +154,10 @@
     { mobile: "+85251111112", first_name: "Daniel", last_name: "Dell52" },
     { mobile: "+85251111113", first_name: "Dakota", last_name: "Deryn53" },
     { mobile: "+85251111114", first_name: "Delia", last_name: "Devon54" },
+    { mobile: "+85241111111", first_name: "Daemon", last_name: "Dara55" },
```
steveyken: with these changes, please make the last_name suffix match the number variant.

E.g. for `41111111` the last_name should be Dara41
for `41111112` the last_name should be Dell42

and so on for all the new names listed below.

PR #708, opened 2019-04-04
Fix distinct issue by sorting after sql

patrixr:
The serializers exploded, so I'm moving to a sorting post-query
This will be replaced once we implement Matt's sorting design later on

No changes on the tests as the expected behaviour did not change

will merge for Matt to be able to test on Staging. Please review post-merge. Thanks :D

### DIFF
--- a/app/controllers/api/v1/orders_controller.rb
+++ b/app/controllers/api/v1/orders_controller.rb
@@ -48,8 +48,18 @@ module Api
         records = apply_filters(@orders).with_eager_load
           .search(params['searchText'], params['toDesignateItem'].presence).descending
           .page(params["page"]).per(params["per_page"])
-        orders = order_response(records)
+        orders = order_response sort_by_urgency(records)
         render json: {meta: {total_pages: records.total_pages, search: params['searchText']}}.merge(JSON.parse(orders))
+      end
+
+      def sort_by_urgency(records)
+        states = array_param(:state)
+        if (states & Order::ACTIVE_STATES).present?
+          records = records.sort_by do |r|
+            r.order_transport.nil? ? -1 : r.order_transport.scheduled_at.to_i
+          end
+        end
+        records
       end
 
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/orders_controller.rb
```diff
@@ -48,10 +48,20 @@ def index
...
           .page(params["page"]).per(params["per_page"])
-        orders = order_response(records)
+        orders = order_response sort_by_urgency(records)
         render json: {meta: {total_pages: records.total_pages, search: params['searchText']}}.merge(JSON.parse(orders))
       end
 
+      def sort_by_urgency(records)
+        states = array_param(:state)
+        if (states & Order::ACTIVE_STATES).present?
+          records = records.sort_by do |r|
```
namrataukirde: Hey @patrixr , Is there any specific reason to use `sort_by` instead of `order`?

namrataukirde: https://apidock.com/rails/ActiveRecord/QueryMethods/order

namrataukirde: as per my knowledge sort_by is slower than order

namrataukirde: And how about we sort orders at search result itself instead of doing it here later on.

https://github.com/crossroads/api.goodcity/blob/master/app/models/order.rb#L302

CC: @steveyken

patrixr: @namrataukirde it's done here as a temporary workaround, because there was an SQL conflict with the serializers. I've temporarily disabled the sql `ORDER BY` and added this in the meantime

As for sort_by vs order, there's no issue in switching if order is more performant

namrataukirde: Yeah.. I feel order is more performant.. But we can discuss it on Monday. Steve told me reason behind why you did this way so no worries. :)

PR #709, opened 2019-04-04
Fix blank user name in slack message

abhinav098:
The issue was, twilio job was getting send before user name was updated. 
So moved it below so that user name is there. Also in the twilio service, it was written as User.current_user.full_name, but it wont be available in the service. So used `@user`.
Please review 
Thanks

### DIFF
--- a/app/services/sendgrid_service.rb
+++ b/app/services/sendgrid_service.rb
@@ -17,6 +17,9 @@ class SendgridService
   def send_email
     if send_to_sendgrid?
       sendgrid_instance.client.mail._("send").post(request_body: mail.to_json)
+    else
+      message = "SlackSMS ('#{user.email}') #{message_body}"
+      SlackMessageJob.perform_later(message, ENV["SLACK_PIN_CHANNEL"])
     end
   end
 
@@ -59,6 +62,11 @@ class SendgridService
   end
 
   private
+
+  def message_body
+    pin = user.most_recent_token.otp_code
+    I18n.t('twilio.browse_sms_verification_pin', pin: pin)
+  end
 
   def send_to_sendgrid?
     Rails.env.production? || Rails.env.staging?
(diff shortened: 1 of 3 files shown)


namrataukirde:
@abhinav098 , In brackets we get mobile number. So will it handle email or phone number in brackets scenario?

namrataukirde:
Means what if user using email id to login?

abhinav098:
No the message was the `send_welcome_message` contains the name of the user.

abhinav098:
ohh i see.. sorry to misundertand.

abhinav098:
Bracket should contain the mobile/email for the user which has been used to sign-up. That's what you're saying right?

namrataukirde:
Yes @abhinav098 👍

namrataukirde:
e.g considering mobile number : `SlackSMS (+85291111111) Single-use pin is 3217. GoodCity.HK welcomes you! Enjoy browsing quality goods.(If you didn't request this message, please ignore)`

app/services/twilio_service.rb
```diff
@@ -30,10 +30,11 @@ def send_unread_message_reminder(url)
   # options[:to] = "+85261111111"
   # options[:body] = "SMS body"
   def send_sms(options)
-    options = {to: @user.mobile}.merge(options)
     if send_to_twilio?
+      options = { to: @user.mobile }.merge(options)
       TwilioJob.perform_later(options)
     else
+      options = { to: @user.mobile.presence || @user.email.presence } .merge(options)
```
namrataukirde: It should be based on what user is using for login and not based on what is present.

Thoughts @abhinav098 ?

abhinav098: Yeah i think you're right.

namrataukirde: 🙂

app/services/sendgrid_service.rb
```diff
@@ -17,6 +17,9 @@ def set_personalizaton_variables
   def send_email
     if send_to_sendgrid?
       sendgrid_instance.client.mail._("send").post(request_body: mail.to_json)
```
namrataukirde: We have slack notifications for prod environment as well right now. So I don't think this is good approach to fix this.

PR #710, opened 2019-04-05
Fix blank sms issue for slack sms.

abhinav098:
This PR fixes the slack sms.
Thanks


app/services/sendgrid_service.rb
```diff
@@ -17,6 +17,9 @@ def set_personalizaton_variables
   def send_email
     if send_to_sendgrid?
```
namrataukirde: We have slack notifications for prod environment as well right now. So I don't think this is good approach to fix this.

PR #711, opened 2019-04-05
Stop blank message if there's no user 

abhinav098:
Fix For this message
SlackSMS ()  has added you to the GoodCity for Charities platform. Please download the app and log in using this mobile number.
Thanks.

### DIFF
--- a/app/services/twilio_service.rb
+++ b/app/services/twilio_service.rb
@@ -12,7 +12,7 @@ class TwilioService
   end
 
   def send_welcome_msg
-    user_name = User.current_user&.full_name
+    user_name = User.current_user&.full_name.presence
     send_sms(body: welcome_sms_text(user_name)) if user_name
   end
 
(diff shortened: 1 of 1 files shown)


app/services/twilio_service.rb
```diff
@@ -12,7 +12,7 @@ def sms_verification_pin(app_name)
   end
 
   def send_welcome_msg
-    user_name = User.current_user&.full_name
+    user_name = User.current_user&.full_name.presence
```
namrataukirde: 👍

steveyken: interesting concept. not seen this before. like it

abhinav098: Yep, `presence` method checks for the value and returns that value. and if there no value or its blank, it returns nil :)
Rails is amazing it aways excites me @steveyken ;)

PR #713, opened 2019-04-06
Gcw 2370 improve offer search

steveyken:
Added the following improvements to the offer search:

- Search for chinese package name
- Search for gogovan driver name
- Search for gogovan driver mobile
- Search for gogovan driver license
- Don't return duplicate results


spec/models/concerns/push_updates_for_offer_spec.rb
```diff
@@ -0,0 +1,43 @@
+# require 'rails_helper'
```
patrixr: This looks odd. Was it a debug that got pushed by mistake ? : )

app/models/concerns/offer_search.rb
```diff
@@ -25,7 +26,10 @@ module OfferSearch
           joins("LEFT OUTER JOIN items ON offers.id = items.offer_id").
           joins("LEFT OUTER JOIN messages ON offers.id = messages.offer_id OR items.id = messages.item_id").
           joins("LEFT OUTER JOIN packages ON packages.item_id = items.id").
-          joins("LEFT OUTER JOIN package_types ON package_types.id = packages.package_type_id")
+          joins("LEFT OUTER JOIN package_types ON package_types.id = packages.package_type_id").
```
namrataukirde: Place the `.` on the next line, together with the method name.

From code climate.

spec/models/concerns/push_updates_for_offer_spec.rb
steveyken: oops... didn't mean to include this file!

app/models/concerns/offer_search.rb
patrixr: don't do it @steveyken  code climate is evil 😆

PR #715, opened 2019-04-09
Gcw 2421 add total request count

abulsayyad123:
Hi Team,
This PR is regarding jira ticket GCW-2421- add total request count on summary tab(https://jira.crossroads.org.hk/browse/GCW-2421). Please review the PR and let me know if any change is required.

Thanks.

### DIFF
--- a/app/controllers/api/v1/orders_controller.rb
+++ b/app/controllers/api/v1/orders_controller.rb
@@ -55,15 +55,7 @@ module Api
       api :GET, '/v1/designations/1', "Get a order"
       def show
         root = is_browse_app? ? "order" : "designation"
-        render json: @order,
-          serializer: serializer,
-          root: root,
-          exclude_code_details: true,
-          include_packages: bool_param(:include_packages, true),
-          include_order: false,
-          include_territory: true,
-          include_images: true,
-          exclude_stockit_set_item: true
+        render json: Api::V1::OrderSerializer.new(order_with_orders_count, root: root).to_json
       end
 
       def transition
@@ -119,6 +111,13 @@ module Api
           include_order: false,
           include_images: true,
           exclude_stockit_set_item: true).as_json
+      end
+
+      def order_with_orders_count
+        Order.users_order_count(@order.created_by_id).each do |key, value|
+          @order.send("#{key}_count=", value)
+        end
+        @order
       end
 
       def order_record
--- a/app/controllers/api/v1/orders_controller.rb
+++ b/app/controllers/api/v1/orders_controller.rb
@@ -55,7 +55,8 @@ module Api
       api :GET, '/v1/designations/1', "Get a order"
       def show
         root = is_browse_app? ? "order" : "designation"
-        render json: Api::V1::OrderSerializer.new(order_with_orders_count, root: root).to_json
+        orders_count = Order.counts_for(@order.created_by_id)
+        render json: { meta: { counts: orders_count } }.merge(Api::V1::OrderSerializer.new(@order, root: root).as_json)
       end
 
       def transition
@@ -111,13 +112,6 @@ module Api
           include_order: false,
           include_images: true,
           exclude_stockit_set_item: true).as_json
-      end
-
-      def order_with_orders_count
-        Order.users_order_count(@order.created_by_id).each do |key, value|
-          @order.send("#{key}_count=", value)
-        end
-        @order
       end
 
       def order_record
--- a/app/serializers/api/v1/order_serializer.rb
+++ b/app/serializers/api/v1/order_serializer.rb
@@ -24,34 +24,6 @@ module Api::V1
     has_one  :address, serializer: AddressSerializer
     has_one  :district, serializer: DistrictSerializer
 
-    def unread_messages_count
-      object.subscriptions.where(state: 'unread', user_id: object.created_by_id).count
-    end
-
-    def unread_messages_count__sql
-      "(select count(*) from subscriptions s where s.order_id = orders.id and s.state = 'unread' and s.user_id = orders.created_by_id)"
-    end
-
-    def submitted_count
-      object.submitted_count
-    end
-
-    def cancelled_count
-      object.cancelled_count
-    end
-
-    def closed_count
-      object.closed_count
-    end
-
-    def draft_count
-      object.draft_count
-    end
-
-    def dispatching_count
-      object.dispatching_count
-    end
-
     def item_ids
     end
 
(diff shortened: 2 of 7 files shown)


app/serializers/api/v1/order_serializer.rb
```diff
@@ -24,6 +26,50 @@ class OrderSerializer < OrderShallowSerializer
...
+      "(select count(*) from subscriptions s where s.order_id = orders.id and s.state = 'unread' and s.user_id = orders.created_by_id)"
+    end
+
+    def user_submitted_order_count
+      Order.users_order_count(object.created_by_id, 'submitted')
+    end
+
+    def user_submitted_order_count__sql
+      "(select count(*) from orders where orders.state = 'submitted'
+        and orders.created_by_id = orders.created_by_id)"
```
steveyken: this doesn't look right... `orders.created_by_id = orders.created_by_id`

config/initializers/constants.rb
```diff
@@ -8,6 +8,7 @@
 STOCK_APP = "stock".freeze
 BROWSE_APP = "browse".freeze
 STOCKIT_APP = "stockit".freeze
+TOTAL_REQUESTS_STATES = ["submitted", "awaiting_dispatch", "closed", "cancelled"].freeze
```
steveyken: This should be in `app/models/orders.rb` as it is specific to orders.
Only general constants that are used elsewhere in the app should be here.

spec/models/order_spec.rb
```diff
@@ -77,6 +77,51 @@
...
+      expect(Order.users_order_count(nil, 'cancelled')).to eq(0)
+      expect(Order.users_order_count(nil, 'closed')).to eq(0)
+    end
+
+    it "will not return orders count if order_state is not provided" do
+      expect(Order.users_order_count(user.id, '')).to eq(0)
+      expect(Order.users_order_count(user.id, '')).to eq(0)
+      expect(Order.users_order_count(user.id, '')).to eq(0)
+      expect(Order.users_order_count(user.id, '')).to eq(0)
+    end
```
steveyken: Great spec coverage

app/controllers/api/v1/orders_controller.rb
```diff
@@ -65,15 +65,7 @@ def sort_by_urgency(records)
...
-        render json: @order,
-          serializer: serializer,
-          root: root,
-          exclude_code_details: true,
-          include_packages: bool_param(:include_packages, true),
-          include_order: false,
-          include_territory: true,
-          include_images: true,
-          exclude_stockit_set_item: true
+        render json: Api::V1::OrderSerializer.new(order_with_orders_count, root: root).to_json
```
steveyken: can you change `to_json` to `as_json` - `as_json` returns a hash and then render will convert it correctly

app/serializers/api/v1/order_serializer.rb
```diff
@@ -34,40 +34,24 @@ def unread_messages_count__sql
...
-      Order.users_order_count(object.created_by_id, 'submitted')
+    def submitted_count
+      object.submitted_count
     end
 
-    def user_submitted_order_count__sql
-      "(select count(*) from orders where orders.state = 'submitted'
-        and orders.created_by_id = orders.created_by_id)"
+    def cancelled_count
+      object.cancelled_count
```
steveyken: given these are now attributes on the order, do you need to have a method defined or is `attributes :cancelled_count` enough?

app/controllers/api/v1/orders_controller.rb
```diff
@@ -131,6 +123,13 @@ def order_response(records)
           exclude_stockit_set_item: true).to_json
       end
 
+      def order_with_orders_count
+        Order.users_order_count(@order.created_by_id).each do |key, value|
```
steveyken: i think a better approach would be to replace the class scope with attribute readers on the order model. You're halfway there already.

Try this:
- delete `users_order_count`
- add the following on `order.rb`

```
def submitted_count
  Order.where(created_by_id: self.created_by_id, state: 'submitted').size
end
```

then reference `submitted_count` in the serializer. 
Once this is done, we should check the speed, it could still be quite slow for the index methods.

Another approach would be to send this data down as `meta` outside of the object. This would be much faster and not require calculating for every order.

So in the controller, `render json: { meta: { counts: {submitted: 2, cancelled: 4} } }.merge(OrderSerializer.new(@results).as_json)...`

app/serializers/api/v1/order_serializer.rb
steveyken: ignore this comment in light of my above suggestion. if it works, we can remove the counts from the serializer altogether

app/controllers/api/v1/orders_controller.rb
abulsayyad123: I am trying it out.

config/initializers/constants.rb
abulsayyad123: Moved it to order_spec file.

spec/models/order_spec.rb
```diff
@@ -77,6 +78,34 @@
     it{ is_expected.to have_db_column(:staff_note).of_type(:string)}
   end
 
+  describe '.counts_for' do
```
patrixr: WARNING: TDD MONOLOGUE INCOMING 😆  (just saw an opportunity to share, the code is actually good)

I love seeing specs ! :D

However, that is actually much more important to have tests on the **controllers** than models. Here's why : 

The code that counts orders could in theory be written on both the model level and the controller level.
However --> The controller is the **entry point** of the application. Meaning that if you test it there, you're actually testing that your server returns the expected value, and therefore behaves in the expected way.

Ideally it would also have an explicit name there, e.g `context 'counting orders by state'`

An example : Imagine the controller does not call `.counts_for` , then technically our server **isn't complying** with the specifications --> even though this test passes.
Ideally it is tested on both the controller and the model, but if you have to choose, the controller is where it should be tested (a more BDD approach).

When it comes to those end to end tests, the tests should not be aware of _how_ the code is written, it should test that the app behaves correctly under a set of circumstance. Think of all the specs of the server, it should be like reading a list of capabilities, or a documentation.

You can find more information in my book `How I annoyed my team to death, by Patrick`

Hope you'll find something interesting here @abulsayyad123 @namrataukirde @Darshan-Chauhan @abhinav098 
Everyone send good vibes to @steveyken , he has to deal with my rambling every day

steveyken: lol, can you send me the ISBN, I'll order a copy for everyone from BookDepository :D

abulsayyad123:
Closing this PR due to changes in the approach. This code is not compatible with the latest changes, creating new PR for this.

PR #716, opened 2019-04-10
Rename of package_images to images

patrixr:
The name `package_images` was causing images not to be found on browse. Changing it for consistency

### DIFF
--- a/app/serializers/api/v1/package_serializer.rb
+++ b/app/serializers/api/v1/package_serializer.rb
@@ -3,7 +3,7 @@ module Api::V1
     embed :ids, include: true
 
     has_one :package_type, serializer: PackageTypeSerializer
-    has_many :images, serializer: ImageSerializer, root: :package_images
+    has_many :images, serializer: ImageSerializer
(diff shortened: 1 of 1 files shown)


app/serializers/api/v1/package_serializer.rb
```diff
@@ -3,7 +3,7 @@ class PackageSerializer < ApplicationSerializer
     embed :ids, include: true
 
     has_one :package_type, serializer: PackageTypeSerializer
-    has_many :images, serializer: ImageSerializer, root: :package_images
+    has_many :images, serializer: ImageSerializer
```
Darshan-Chauhan: Will this have any effect on other apps??

patrixr: The reason for that is that Stock mostly use `StockItemSerializer` and browse expects `BrowsePackageSerializer`

patrixr: part of the Serializer hell I'd like to fix haha

PR #726, opened 2019-04-25
Using SQL to sort order search results.

steveyken:
Use SQL to order search results by scheduled_at date


patrixr:
> WIP
> 
> * Use SQL to order search results. Question: @patrixr where should orders with no scheduled_at date be sorted to (top or bottom of list)?
> * Still need to fix specs

@steveyken Good question, I would guess top, otherwise we might never find them if the list is really long
Although, that would only apply to drafts, which we would probably filter for explicitely

patrixr:
> * Still need to fix specs

Specs should already be in place : )

PR #728, opened 2019-04-29
GCW-2555-Add pagination according to params

abulsayyad123:
Hi Team,
I have added pagination for orders index action according to passed per_page params.
Thanks

### DIFF
--- a/app/controllers/api/v1/orders_controller.rb
+++ b/app/controllers/api/v1/orders_controller.rb
@@ -47,7 +47,7 @@ module Api
         return recent_designations if params['recently_used'].present?
         records = apply_filters(@orders).with_eager_load
           .search(params['searchText'], params['toDesignateItem'].presence).descending
-          .page(params["page"]).per(150)
+          .page(params["page"]).per(params["per_page"] || 150)
         orders = order_response sort_by_urgency(records)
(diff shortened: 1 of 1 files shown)


app/controllers/api/v1/orders_controller.rb
```diff
@@ -47,7 +47,7 @@ def index
         return recent_designations if params['recently_used'].present?
         records = apply_filters(@orders).with_eager_load
           .search(params['searchText'], params['toDesignateItem'].presence).descending
-          .page(params["page"]).per(150)
+          .page(params["page"]).per(params["per_page"] || 150)
```
abhinav098: what will be the per_page param ? @abulsayyad123 ?

patrixr: Could we restore the default number back to 25 ? 150 is no longer needed, it's too agressive as a default.

@abhinav098 I believe `params["per_page"]` is what used to be there before we hardcoded the 150. The stock app sends `?per_page=25`

abulsayyad123: Sure @patrixr , I wasn't sure about the number that is why I added the default one. Will change it to 25.

PR #729, opened 2019-05-02
Fix wrong order channel definition

patrixr:
This PR fixes the issue @abhinav098 reported where popups would appear on Admin whenever an order was submitted


Darshan-Chauhan:
@patrixr , will this still happen if User with `Order fulfilment role` is logged-in to donor app?

patrixr:
> @patrixr , will this still happen if User with `Order fulfilment role` is logged-in to donor app?

@Darshan-Chauhan no, the order fulfilment channel only gets enabled on the stock app

In channel.rb :

`channels << ORDER_FULFILMENT_CHANNEL if user.order_fulfilment? and app_name == STOCK_APP`

PR #731, opened 2019-05-03
Remove unread messages count from serializers

abhinav098:
This PR removes the unread_messages_count attribute from order serialisers as its not required. 
Its calculated on the ember side.
Thanks


abhinav098:
Sure, will merge it once I get one more approval. 
cc- @namrataukirde

PR #733, opened 2019-05-03
rake task for updating existing data for quantity items

abhinav098:


lib/tasks/update_received_quantity_to_package.rake
```diff
@@ -0,0 +1,15 @@
+# rake goodcity:update_received_quantity_to_package
+namespace :goodcity do
+  desc 'Update received quantities to package'
+  task update_received_quantity_to_package: :environment do
+    counter = 0
+    quantity_packages = Package.where("packages.received_quantity > 1")
```
namrataukirde: Package.where("packages.received_quantity > 1").find_each will be better here. No need assign it to variable. :)

lib/tasks/update_received_quantity_to_package.rake
```diff
@@ -0,0 +1,15 @@
...
+    quantity_packages.each do |package|
+      all_packages = Package.where("packages.inventory_number LIKE ?",
+        "#{package.inventory_number}%")
+      child_packages = all_packages - [package] if all_packages.count > 1
+      child_packages.any? && package.update(received_quantity: package.quantity) && counter+=1
```
namrataukirde: What if that package is designated? In that scenario, quantity will be 0 and we will never have received quantity as 0.

namrataukirde:
Also please create a ticket for this issue on Jira.

abhinav098:
sure 👍

lib/tasks/update_received_quantity_to_package.rake
abhinav098: sure 👍

steveyken: yes, and definitely use `find_each` - worth a quick read to see why. hint: batch loading saves memory -  just using `find` will load 300,000+ records from SQL into ruby serialized AR objects before it's started to process the loop. See https://apidock.com/rails/ActiveRecord/Batches/ClassMethods/find_each

lib/tasks/update_received_quantity_to_package.rake
steveyken: also, if `all_packages.count == 0` then `child_packages` will not be initialized.

PR #736, opened 2019-05-10
GCW-2534- Organisation not searchable

abulsayyad123:
Hi Team,
This PR is for JIRA ticket GCW-2534- Organisation not searchable with half organisation name. Possible reason for not displaying the expected results is because we are restricting only 25 records from api. I have added pagination in the query.
Please review the PR and let me know if any changes is required.


app/controllers/api/v1/gc_organisations_controller.rb
```diff
@@ -40,12 +40,12 @@ def organisation_name_serializer
 
     def find_record_and_render_json(serializer)
       if params['ids'].present?
-        records = @organisations.where(id: params['ids'])
+        records = @organisations.where(id: params['ids']).page(params["page"]).per(params["per_page"] || 25)
```
namrataukirde: I will suggest to add 25 value in constant.

PR #738, opened 2019-05-13
FIX=> Handling multiquantity filter on api side

abulsayyad123:
Hi Team,
This PR is for handling multi quantity filter on Api side instead of stock app side. I have added params to filter for Orders page Item search. Please review the PR and let me know if any change is required. 

Thanks

### DIFF
--- a/app/controllers/api/v1/packages_controller.rb
+++ b/app/controllers/api/v1/packages_controller.rb
@@ -128,9 +128,9 @@ module Api
       def search_stockit_items
         records = @packages # security
         records = records.search(search_text: params['searchText'], item_id: params['itemId'],
+          restrict_multi_quantity: params['restrictMultiQuantity'],
           with_inventory_no: params['withInventoryNumber'] == 'true') if params['searchText'].present?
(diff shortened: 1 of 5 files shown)


abhinav098:
Could you please add some test cases for this change @abulsayyad123 ?

app/controllers/api/v1/packages_controller.rb
```diff
@@ -130,6 +130,7 @@ def search_stockit_items
         records = records.search(search_text: params['searchText'], item_id: params['itemId'],
           with_inventory_no: params['withInventoryNumber'] == 'true') if params['searchText'].present?
         params_for_filter = ['state', 'location'].each_with_object({}){|k, h| h[k] = params[k] if params[k].present?}
+        records =  records.inventorize_and_not_multiquantity if params["isOrdersSearch"]
```
patrixr: If think the param name should be more explicit as to what it does, `isOrdersSearch` does not give any information as to what it does. I'd argue towards something explicit like `?allowMultiQuantity=true|false` (example)

PR #739, opened 2019-05-15
Switch off image deletion on live API.

steveyken:
This will temporarily disable the image delete API job which assist us operationally whilst we figure out the bug in more detail.


.env.sample
```diff
@@ -9,6 +9,7 @@ OTP_CODE_VALIDITY=
 CLOUDINARY_CLOUD_NAME=
 CLOUDINARY_API_KEY=
 CLOUDINARY_API_SECRET=
+ENV['PREVENT_CLOUDINARY_IMAGE_DELETION']=
```
patrixr:
```suggestion
PREVENT_CLOUDINARY_IMAGE_DELETION=
```

PR #741, opened 2019-05-17
Make changes to the API for using fcm instead of GCM

abhinav098:
Changes to register device with fcm

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -135,6 +135,7 @@ workflows:
       - deploy:
           requires:
             - test
+            - audit
           filters:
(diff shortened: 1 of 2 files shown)


steveyken:
Do we need to change the label from gcm to fcm here? I know for completeness this is good but what happens to all the previous versions of our apps that send 'gcm' when they register? Perhaps better to keep 'gcm' for now

abhinav098:
We can put both, "gcm" and "fcm" too. for previous. For for new devices it will take fcm only.

namrataukirde:
👍

PR #743, opened 2019-05-17
New push updates for orders

patrixr:
Removed the old PushUpdates class from orders, replaced it with the configurable version.
Also added a way to disable stockit sync (env variable), useful when testing


Darshan-Chauhan:
👍

PR #749, opened 2019-05-27
FIX=> orders_packages added in orders response for browse app

abulsayyad123:
Hi Team,
This PR is for adding orders_packages in response of orders `show` action along with associated specs. Specs has been added considering both `STOCK` and `BROWSE` app.

### DIFF
--- a/app/controllers/api/v1/orders_controller.rb
+++ b/app/controllers/api/v1/orders_controller.rb
@@ -59,7 +59,7 @@ module Api
           serializer: serializer,
           root: root,
           exclude_code_details: true,
-          include_packages: bool_param(:include_packages, true),
+          include_packages: bool_param(:include_packages, false),
           include_order: false,
           include_territory: true,
           include_images: true,
--- a/spec/controllers/api/v1/orders_controller_spec.rb
+++ b/spec/controllers/api/v1/orders_controller_spec.rb
@@ -97,7 +97,7 @@ RSpec.describe Api::V1::OrdersController, type: :controller do
         expect(parsed_body['designations'].count).to eq(Order.where.not(state: 'draft').count)
         expect(parsed_body['designations'].map { |it| it['state'] }).to_not include('draft')
       end
-      
+
       # Test turned off as currently hardcoded to 150
       # it 'returns the number of items specified for the page' do
       #   5.times { FactoryBot.create :order, :with_state_submitted } # There are now 7 no-draft orders in total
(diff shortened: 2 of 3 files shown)


app/controllers/api/v1/orders_controller.rb
```diff
@@ -59,7 +59,7 @@ def show
           serializer: serializer,
           root: root,
           exclude_code_details: true,
-          include_packages: bool_param(:include_packages, true),
+          include_packages: bool_param(:include_packages, false),
```
patrixr: this worries me a bit, the orders serializer has very weird behaviours, where some `include_x` are dependant on others.

@steveyken could you advise ?

Darshan-Chauhan: Agree with @patrixr , even if you pass `true or false` to include relationship the packages gets loaded based on relationship with other models.

For example here it was not loading orders_packages because `bool_param(:include_packages, true)` : https://github.com/crossroads/api.goodcity/blob/master/app/serializers/api/v1/order_serializer.rb#L47

As per the current code, i.e `bool_param(:include_packages, true)` it should never load orders_packages through order serializer. But if you go to Stock App and load any order(`/orders/:id`), you'll see orders_packages in response, so one might wonder if we're not loading orders_packages through order serializer then how they're coming in response. 
Answer to that is `packages` gets loaded through order serializer and orders_packages gets loaded through package serializer.

So in case if you didn't want to include a model `X` in response but if that model `X` has relationship with other model `Y` which gets called through the serializer then model `X` will get included even though we've explicitly specified to not to include.

So yeah... 😄

steveyken: Yeah it's really hard to understand completely the consequences of this change, which means we really need to fix the underlying code. Perhaps now that we've all been involved in serialization for a while we can brainstorm a 'best practices' guide for which dependencies to load and how. I know it's hard because AMS `include_...` is also broken in plenty of cases too so hard to know what is working and what isn't!

patrixr: Some context on the codes going on in the serializers (which make *no* sense to me) 

```
    def include_cart_packages?
      !@options[:include_packages] && !@options[:include_order]
    end

    def include_orders_packages?
      !@options[:include_packages] && !@options[:include_order]
    end
```

I've tried fixing those in the past, but those are actually preventing things from exploding elsewhere (should not have been an acceptable fix to begin with)

Darshan-Chauhan: @patrixr , the reason behind using `!` in `include_x` method is, all your controller action is calling same serializer. Say in action `a` you're passing `include_packages: true` but in action `b` you're not passing any options, in that case `@options[:include_packages]` will be undefined and hence `!@options[:include_packages]` will be `true`. 

This could really mess up things if you forgot to pass any options or didn't use `!` in serializer.

abulsayyad123:
Closing this PR. Need discussion with Team to fix this issue.  I have created ticket for this issue https://jira.crossroads.org.hk/browse/GCW-2675

PR #750, opened 2019-05-28
[NEW] Goodcity Settings

patrixr:
This PR adds a new table to the api (named `goodcity_settings`)

Each row is a key value pair, representing a configuration entry. The idea is to have the apps read that configuration at startup, in order for us to be able to tweak behaviours without making releases.

You'll notice I added an `admin.html` page.
This loads a minimal UI to allow us to edit settings (frontend is available here: https://github.com/crossroads/settings.goodcity), it is currently being loaded via CDN as I didn't want to overload the api repo with UI code.

How to use:
* open `api.goodcity.hk/admin`
* login using mobile/pin
* edit configs

It is generally important that we move away from ssh-ing into prod, or editing the database directly. Those are (kinda) acceptable now, but we eventually want to build tools to be able to support Goodcity as it grows. This could serve as a first step in that direction (*suggestions are always welcome*)


db/schema.rb
```diff
@@ -202,10 +205,18 @@
   add_index "goodcity_requests", ["order_id"], name: "index_goodcity_requests_on_order_id", using: :btree
   add_index "goodcity_requests", ["package_type_id"], name: "index_goodcity_requests_on_package_type_id", using: :btree
 
+  create_table "goodcity_settings", force: :cascade do |t|
+    t.string "key"
+    t.string "value"
+    t.string "desc"
```
abulsayyad123: What is "desc" ?

patrixr: description

app/models/goodcity_setting.rb
```diff
@@ -0,0 +1,4 @@
+class GoodcitySetting < ActiveRecord::Base
+  validates :key, presence: true
+  validates :key, uniqueness: true
```
steveyken:
```suggestion
  validates :key, uniqueness: true, presence: true
```

You can batch validation params together as a hash

db/migrate/20190524024950_create_goodcity_setting.rb
```diff
@@ -0,0 +1,9 @@
+class CreateGoodcitySetting < ActiveRecord::Migration
+  def change
+    create_table :goodcity_settings do |t|
+      t.string :key, index: true, unique: true, foreign_key: true
```
steveyken: What's the foreign_key?

db/schema.rb
steveyken: Might be confused with 'descending' in SQL!

db/migrate/20190524024950_create_goodcity_setting.rb
patrixr: oops, bad copy pasta

PR #751, opened 2019-05-29
fix azure notifications with fcm

namrataukirde:
Hi Team,

This PR contains changes to make azure notifications compatible with fcm. Even if they have migrated to fcm tags are still gcm.

Consider doc: https://firebase.google.com/docs/cloud-messaging/send-message

Please review.

Thanks.

### DIFF
--- a/app/services/azure_notifications_service.rb
+++ b/app/services/azure_notifications_service.rb
@@ -39,6 +39,7 @@ class AzureNotificationsService
 
   def encoded_url(platform, handle)
     case platform
+    when "gcm" then URI::encode("GcmRegistrationId eq '#{handle}'")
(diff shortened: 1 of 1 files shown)


Darshan-Chauhan:
We need GCM and FCM both. As older APPs will be still using GCM.

steveyken:
> We need GCM and FCM both. As older APPs will be still using GCM.

Yes I wanted to make the same point. Keep GCM as well but redirect it to FCM for notification hub

namrataukirde:
Yes Right @Darshan-Chauhan, @steveyken. We need to keep both as old apps are running fine with notifications. I will add that change. Thanks.

namrataukirde:
@steveyken Could you please review PR again?

PR #752, opened 2019-05-30
GCW- 2421- Add total request count to order's contact summary

abulsayyad123:
Hi Team,
This PR is for ticket GCW-2421- Adding total request counts. Since, we have common routes for fetching orders data in nested routes. Which will results in unnecessary request i.e orders_count, for data which is not required in other pages. So, I have added separate ajax request for orders_count on orders_summary page instead of adding it in users or orders payload.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -332,7 +332,7 @@ class Ability
   def user_abilities
     can [:current_user_profile], User
     can [:show, :update], User, id: @user_id
-    can [:index, :show, :update, :recent_users], User if can_read_or_modify_user?
+    can [:index, :show, :update, :recent_users, :orders_count], User if can_read_or_modify_user?
   end
 
   def version_abilities
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -331,8 +331,8 @@ class Ability
 
   def user_abilities
     can [:current_user_profile], User
-    can [:show, :update], User, id: @user_id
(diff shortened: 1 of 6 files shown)


app/models/ability.rb
```diff
@@ -332,7 +332,7 @@ def stockit_abilities
   def user_abilities
     can [:current_user_profile], User
     can [:show, :update], User, id: @user_id
-    can [:index, :show, :update, :recent_users], User if can_read_or_modify_user?
+    can [:index, :show, :update, :recent_users, :orders_count], User if can_read_or_modify_user?
```
namrataukirde: Why we need to add this with `can_read_or_modify_user?` permission?

app/models/order.rb
```diff
@@ -96,6 +96,11 @@ def can_dispatch_item?
     ORDER_UNPROCESSED_STATES.include?(state)
   end
 
+  def self.counts_for(created_by_id)
+    order_state_counts = group(:state).where(created_by_id: created_by_id).count
```
namrataukirde: How about  `where(state IN (?), valid_states_array).distinct.group(:state).where(:created_by_id).count` ?

Could please check which is better one and let us know?

abulsayyad123: Sure, will try it out.

abulsayyad123: I benchmarked performance of both queries.  The second query is little slow on an average of 2-3 ms.

abulsayyad123: Added refactored query.

PR #754, opened 2019-05-31
enable sms_reminders

namrataukirde:
Hi Team,

This PR is to enable sms reminders on staging.

Thanks.


namrataukirde:
Just FYI: Before merging this PR i want to check count of users on staging for reminders.

namrataukirde:
Current valid user count for this PR is 51.

config/initializers/constants.rb
```diff
@@ -1,5 +1,5 @@
 TITLE_OPTIONS = %w(Mr Mrs Miss Ms)
-SUBSCRIPTION_REMINDER_TIME_DELTA = 4.hours
+SUBSCRIPTION_REMINDER_TIME_DELTA = 15.minutes
```
patrixr: wouldn't that variable affect prod as well ?

namrataukirde: Yup.. It will affect for prod as well but before merging to prod we will change the value. Or another thought is we can have 2 different constants for staging and prod. 


What are your thoughts @patrixr?

namrataukirde:
Guys, I want to merge it. So tim's can test it.

abhinav098:
👍

PR #755, opened 2019-06-03
Add rake for fix missing organizations from orders

abhinav098:
Earlier I was thinking of handling the scenario for users having multiple organisations as well, but looking at this data, i don't think we need that anymore.
```
Staging Data 
o = Order.where("organisation_id IS NULL and code ILIKE ?","GC-%").where.not(state: "draft").map(&:created_by).map(&:organisations).map(&:count)
 => [1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 1, 1, 1, 0, 1, 1, 1, 1, 1, 0, 1, 1, 0, 1, 1, 0, 1, 1, 1, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1] 
```
```
Prod Data => 
Order.where("organisation_id IS NULL and code ILIKE ?","GC-%").where.not(state: "draft").map(&:created_by).compact.map(&:organisations).map(&:count)
 => [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 1, 1, 1, 1, 1, 1, 1, 1, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1] 
```

### DIFF
--- a/lib/tasks/fix_missing_organisations.rake
+++ b/lib/tasks/fix_missing_organisations.rake
@@ -7,14 +7,14 @@ namespace :goodcity do
     #get all the gc orders without organisation and not draft draft orders
     Order.where("orders.state <> ? and orders.organisation_id IS NULL and orders.created_by_id IS NOT NULL and code ILIKE ?","draft", "GC-%").find_each do |order|
       creator_organisations = order.created_by&.organisations
-      next unless creator_organisations
+      next if creator_organisations.blank?
       if order.update(organisation_id: creator_organisations.first.id)
         count += 1
         print "."
       end
     end
     log.info(": #{count} orders updated")
-    puts "#{count} orders updated"
     log.close
+    puts "#{count} orders updated"
   end
 end
--- a/lib/tasks/fix_missing_organisations.rake
+++ b/lib/tasks/fix_missing_organisations.rake
@@ -1,11 +1,14 @@
-#rake goodcity:fix_missing_organisations
+# rake goodcity:fix_missing_organisations
 
 namespace :goodcity do
   task fix_missing_organisations: :environment do
     count = 0
     log = Goodcity::RakeLogger.new("fix_missing_organisations")
-    #get all the gc orders without organisation and not draft draft orders
-    Order.where("orders.state <> ? and orders.organisation_id IS NULL and orders.created_by_id IS NOT NULL and code ILIKE ?","draft", "GC-%").find_each do |order|
+    # get all the gc orders without organisation and not draft draft orders
+    Order
+      .where(organisation_id: nil)
+      .where.not(created_by_id: nil, state: 'draft')
+      .where("code ILIKE ?", "GC-%").find_each do |order|
       creator_organisations = order.created_by&.organisations
       next if creator_organisations.blank?
       if order.update(organisation_id: creator_organisations.first.id)
--- a/lib/tasks/fix_missing_organisations.rake
+++ b/lib/tasks/fix_missing_organisations.rake
@@ -1,0 +1,20 @@
+#rake goodcity:fix_missing_organisations
+
+namespace :goodcity do
+  task fix_missing_organisations: :environment do
+    count = 0
+    log = Goodcity::RakeLogger.new("fix_missing_organisations")
+    #get all the gc orders without organisation and not draft draft orders
+    Order.where("orders.state <> ? and orders.organisation_id IS NULL and orders.created_by_id IS NOT NULL and code ILIKE ?","draft", "GC-%").find_each do |order|
+      creator_organisations = order.created_by&.organisations
+      next unless creator_organisations
+      if order.update(organisation_id: creator_organisations.first.id)
+        count += 1
+        print "."
+      end
+    end
+    log.info(": #{count} orders updated")
+    puts "#{count} orders updated"
+    log.close
+  end
+end


lib/tasks/fix_missing_organisations.rake
```diff
@@ -0,0 +1,21 @@
+#rake goodcity:fix_missing_organisations
+
+namespace :goodcity do
+  task fix_missing_organisations: :environment do
+    count = 0
+    log = Goodcity::RakeLogger.new("fix_missing_organisations")
+    #get all the gc orders without organisation and not draft draft orders
+    Order.where.not(state: 'draft')
```
namrataukirde: I feel these 2 queries can be written as single query.

lib/tasks/fix_missing_organisations.rake
```diff
@@ -0,0 +1,20 @@
+#rake goodcity:fix_missing_organisations
+
+namespace :goodcity do
+  task fix_missing_organisations: :environment do
+    count = 0
+    log = Goodcity::RakeLogger.new("fix_missing_organisations")
+    #get all the gc orders without organisation and not draft draft orders
+    Order.where("orders.state <> ? and orders.organisation_id IS NULL and orders.created_by_id IS NOT NULL and code ILIKE ?","draft", "GC-%").find_each do |order|
+      creator_organisations = order.created_by&.organisations
+      next unless creator_organisations
```
namrataukirde: What if you get blank active relation like : `#<ActiveRecord::Associations::CollectionProxy []>` Will next get executed?

lib/tasks/fix_missing_organisations.rake
```diff
@@ -0,0 +1,20 @@
...
+      if order.update(organisation_id: creator_organisations.first.id)
```
namrataukirde: Could you please verify once if there is any callback which can cause an issue as we are using an update? If not then great. :)

abhinav098: I have purposely used update so that callbacks get fired and the jobs which are intended, they get fired while updating the organisation for order. I have checked push_changes callback gets fired, which will run a job for updating the order @namrataukirde .

lib/tasks/fix_missing_organisations.rake
abhinav098: I think next will not get executed, will add a `blank?` check here, so that we make sure it doesn't break.

abhinav098:
@namrataukirde, I have taken care of your comments on this Rake, Please review this again. 
Thanks

PR #759, opened 2019-06-05
Remove hardcoded donor conditions

patrixr:
The API was sending different labels for donor_conditions in some cases, which caused inconsistencies with the browse app (As reported by Nokia). This PR normalizes the texts 

Couple of general rules to note about the code that was removed : 
```
 "(CASE WHEN donor_conditions.id=1 THEN 'New'
              WHEN donor_conditions.id=2 THEN 'Mixed'
              WHEN donor_conditions.id=3 THEN 'Used'
              WHEN donor_conditions.id=4 THEN 'Broken'
        END)"
```

* We cannot hardcode IDs anywhere, since we can never guarantee that they match. e.g If the table was ever re-seeded or edited, we would likely have different IDs

* We want to avoid hardcoded strings in the code as they are hard to track and make it very hard to ensure consistency ('New', 'Mixed', etc)

Pretty basic stuff, but I thought I'd point it out anyways 😄

### DIFF
--- a/app/controllers/api/v1/packages_controller.rb
+++ b/app/controllers/api/v1/packages_controller.rb
@@ -53,8 +53,7 @@ module Api
           include_order: true,
           include_orders_packages: true,
           exclude_stockit_set_item: @package.set_item_id.blank? ? true : false,
-          include_images: @package.set_item_id.blank?,
-          include_stock_condition: is_stock_app?
+          include_images: @package.set_item_id.blank?
       end
 
       api :POST, "/v1/packages", "Create a package"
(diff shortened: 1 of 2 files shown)


namrataukirde:
Completely agree with this and now we are not using any hardcoded such thing. Maybe the code is too old 😉 But thanks for mentioning it. We all can take care of it going forward. :)

namrataukirde:
Just to 1 thing. Names in live db are different. So maybe that's why we did this. You can just check it before merging it. We can have a discussion on this.

patrixr:
@namrataukirde 

Correct. This code was originally meant to have them match the StockIt names, but that also (mistakenly) caused the names not to be consistent across GoodCity. 
Matt and Steve are aware of the impact of this change : )
If we ever want to re-add those we can talk about it, and we would re-implement it correctly. This code is a time-bomb sadly, so we're very keen to get it out 😄

namrataukirde:
Great then, Thanks for the clarification. 👍

PR #760, opened 2019-06-06
Tweaked tests and added user.orders relation

steveyken:
Just some small tweaks. If you approve, feel free to merge immediately.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -7,6 +7,7 @@ class User < ActiveRecord::Base
   has_one :address, as: :addressable, dependent: :destroy
   has_many :auth_tokens, dependent: :destroy
   has_many :offers, foreign_key: :created_by_id, inverse_of: :created_by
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -7,6 +7,7 @@ class User < ActiveRecord::Base
   has_one :address, as: :addressable, dependent: :destroy
   has_many :auth_tokens, dependent: :destroy
   has_many :offers, foreign_key: :created_by_id, inverse_of: :created_by
+  has_many :orders, foreign_key: :created_by_id, inverse_of: :created_by
```
namrataukirde: has_many :created_orders, -> { order "id DESC" }, class_name: "Order", foreign_key: :created_by_id

We do have this. Still, we want to add orders association?

namrataukirde:
Test cases related changes looks good 👍 

Added just one query.

steveyken:
This PR is no longer necessary based on @namrataukirde comments and I've also cleaned the specs in another PR

PR #762, opened 2019-06-10
GCW-2616 allow ability to search goods on browse

abhinav098:
Hello @team,
This PR adds the ability to search items on browse app.
Please review this PR.
Thanks.
Ember PR => https://github.com/crossroads/browse.goodcity/pull/404


app/controllers/api/v1/packages_controller.rb
```diff
@@ -36,8 +36,14 @@ class PackagesController < Api::V1::ApiController
 
       api :GET, "/v1/packages", "get all packages for the item"
       def index
-        @packages = @packages.find(params[:ids].split(",")) if params[:ids].present?
-        render json: @packages, each_serializer: serializer, include_orders_packages: true
+        if params["browseSearch"]
```
patrixr: I don't think this property is needed, you can just check the presence of `params["searchText"`] instead.

If we look at the api design we **want**,  I personally would expect all the following to work :

- `/packages` -> all
- `/packages?searchText=text` -> search
- `/packages?ids=1,2,3` -> specific ids

for which the implementation could *roughly* be something like : 
```
records = @packages
records = records.search({search_text: params['searchText']}) if params["searchText"].present?
records = records.find(params[:ids].split(",")) if params[:ids].present?
records = records..filter('state' => params["state"]) if params["state"].present?
```

One question for @steveyken --> I kinda feel we should consolidate the serializers, and not have `BrowsePackageSerializer`. All of those `is_browse_app` make the control-flow complexity sky-rocket.
I've already started using the normal serializer on browse (for some requests), they're pretty much the same

steveyken: yes that's good. re ```BrowsePackageSerializer``` I think @abhinav had begun work a while back to see if it was possible. Did you find any roadblocks?

abhinav098: Yes @steveyken I tried doing that once for message serializers, but it wasn't required at that time, so I stopped working on it.

app/controllers/api/v1/packages_controller.rb
```diff
@@ -36,8 +36,12 @@ class PackagesController < Api::V1::ApiController
 
       api :GET, "/v1/packages", "get all packages for the item"
       def index
-        @packages = @packages.find(params[:ids].split(",")) if params[:ids].present?
-        render json: @packages, each_serializer: serializer, include_orders_packages: true
+        records = @packages #security
+        records = records.filter("state" => params["state"]) if params["state"].present?
+        records = records.search({search_text: params["searchText"]}) if params["searchText"]
```
patrixr: do we have specs for those ?

abhinav098: I am working on specs right now. Will add those in a while.

app/controllers/api/v1/packages_controller.rb
```diff
@@ -273,6 +277,10 @@ def stock_serializer
         Api::V1::StockitItemSerializer
       end
 
+      def browse_serializer
+        Api::V1::BrowsePackageSerializer
```
patrixr: can we try not using this serializer ? I'm still not convinced it should exist in the first place :)

abhinav098: I'll check and let you know about that.

abhinav098: I've checked it, it works fine with `Api::V1::PackageSerializer`, so will use that only.

abhinav098: I've checked it, it works fine with `Api::V1::PackageSerializer`, so will use that only.

cc - @patrixr @steveyken @namrataukirde

app/controllers/api/v1/packages_controller.rb
patrixr: Cool, remember that as much as we can, specs should be written _before_ the implementation : )

app/controllers/api/v1/packages_controller.rb
patrixr: nice : )

app/controllers/api/v1/packages_controller.rb
```diff
@@ -36,8 +36,12 @@ class PackagesController < Api::V1::ApiController
 
       api :GET, "/v1/packages", "get all packages for the item"
       def index
-        @packages = @packages.find(params[:ids].split(",")) if params[:ids].present?
-        render json: @packages, each_serializer: serializer, include_orders_packages: true
+        records = @packages #security
```
steveyken: no real need for `records = ` just use `@packages = @packages...` - like we do elsewhere and above.

app/controllers/api/v1/packages_controller.rb
```diff
@@ -36,8 +36,12 @@ class PackagesController < Api::V1::ApiController
...
+        records = records.filter("state" => params["state"]) if params["state"].present?
+        records = records.search({search_text: params["searchText"]}) if params["searchText"]
+        records = records.find(params[:ids].split(",")) if params[:ids].present?
+        records = records.page(params["page"]).per(params["per_page"])
```
steveyken: what if `page` and `per_page` are omitted or are bad? See `page` and `per_page` methods down around line 150 and use those. Perhaps even move those methods to ApiController so that other controllers can use them

app/controllers/api/v1/packages_controller.rb
steveyken: great if we can remove `BrowsePackageSerializer`

app/models/ability.rb
```diff
@@ -7,7 +7,7 @@ class Ability
   attr_accessor :user, :user_id, :admin, :supervisor, :reviewer, :user_offer_ids, :user_permissions
 
   PERMISSION_NAMES = ['can_manage_items', 'can_manage_goodcity_requests', 'can_manage_packages',
-    'can_manage_offers', 'can_manage_organisations_users',
+    'can_manage_offers', 'can_manage_organisations_users', 'can_search_packages',
```
steveyken: I'm not sure we need a permission for being able to search as this is something an anonymous user can do. However, we do need to restrict what packages anonymous users can see (only `allow_web_publish == true`) which is something we do already.

app/models/concerns/package_filtering.rb
```diff
@@ -12,7 +12,7 @@ module PackageFiltering
         where("item_id = ?", item_id)
       else
         search_query = ['inventory_number', 'designation_name', 'notes', 'case_number'].
-          map { |f| "#{f} ILIKE :search_text" }.
+          map { |f| "packages.#{f} ILIKE :search_text" }.
```
steveyken: 👍

app/models/package.rb
```diff
@@ -68,6 +68,7 @@ class Package < ActiveRecord::Base
   scope :undispatched, -> { where(stockit_sent_on: nil) }
   scope :undesignated, -> { where(order_id: nil) }
   scope :not_multi_quantity, -> { where("quantity <= 1") }
+  scope :browse_in_stock_packages, -> {where("quantity = 1 AND allow_web_publish=true")}
```
steveyken: Good effort to use this in permissions but I think it's important to keep permissions focussed only on security access (in ability.rb). 

`quantity=1` is not a security measure, it's a filter - better to add that in package filters and I think `allow_web_publish=true` is already taken care of via permissions already.

abhinav098:
Thanks for reviewing @steveyken. Will incorporate these changes and send the PR again for review. 
:+1:

abhinav098:
@steveyken @namrataukirde @patrixr , Please review this PR, I have made all the required changes, and changed the approach for searching goods on browse.

app/controllers/api/v1/api_controller.rb
```diff
@@ -43,6 +43,18 @@ def render_objects_with_cache(object, pid)
       end
     end
 
+    # nil.to_i = 0
+    def page
+      @page = params['page'].to_i
+      (@page == 0) ? 1 : @page
```
namrataukirde: instead of `== 0` we can use `.zero?` method here

app/controllers/api/v1/api_controller.rb
```diff
@@ -43,6 +43,18 @@ def render_objects_with_cache(object, pid)
...
+    end
+
+    # max limit is 25
+    def per_page
+      @per_page = params['per_page'].to_i
+      (@per_page == 0 or @per_page > DEFAULT_SEARCH_COUNT) ? DEFAULT_SEARCH_COUNT : @per_page
```
namrataukirde: Instead of or can we use || ? Let me know if any specific reason to use this.

app/controllers/api/v1/api_controller.rb
abhinav098: yeah, sure @namrataukirde , will do that. Thanks :)

app/controllers/api/v1/api_controller.rb
abhinav098: No reason, will change to `||`.

steveyken:
I think this is ready to merge? @abhinav098

abhinav098:
U'mm, no, Browse side is not compatible with this PR. Once the refactor one is complete, we can merge that to this one. And then merge this PR to master, because refactor PR is a child PR of this branch. https://github.com/crossroads/api.goodcity/pull/767

abhinav098:
This PR is now ready to merge to master @steveyken 
cc - @namrataukirde @abulsayyad123 @patrixr

PR #765, opened 2019-06-14
Alter SQL to filter offers in the join query

steveyken:
This ensures the correct user and message state are related.


namrataukirde:
Looks good to me. 👍

PR #766, opened 2019-06-17
Backend cart

patrixr:
Hi team,

Here are the API changes for the cloud based cart.

## New model

I have added the `cart_item` model/controller.
When a user requests an item on browse, it is added there. Those are later transformed into `orders_packages` when we checkout/submit. 
Until the order is submitted, the cart is editable 

## New checkout endpoint

I've added an endpoint that takes care of the whole checkout process, we no longer want this to exist on the front-end side

Ths endpoint is `/cart_items/checkout`. When called, it'll create all the `orders_packages`, empty the cart, and submit the draft order.


app/controllers/api/v1/cart_items_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+      api :POST, '/v1/cart_items', "Create a cart item"
+      def create
+        save_and_render_object_with_errors(@cart_item)
+      end
+
+      api :POST, '/v1/cart_items/checkout', "Checkout and add all the packages to an order"
+      param_group :checkout
+      def checkout
+        errors = CartCheckout
+          .designate_cart_items(@cart_items, ignore_unavailable: params["ignore_unavailable"].present?)
```
steveyken: `ignore_unavailable=no` would mean yes :)

app/controllers/api/v1/cart_items_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+      def cart_item_params
+        params.require(:cart_item).permit(
+          :is_available,
+          :user_id,
+          :package_id
+        )
+      end
+
+      def checkout_order
+        Order.find_by(id: params[:order_id]) if params[:order_id].present?
```
steveyken: The following should be safe as id is primary key and must always exist (not nil). It will raise `RecordNotFound` if it doesn't exist which should result in 422 response.
```suggestion
        Order.find(params[:order_id])
```

app/models/orders_package.rb
```diff
@@ -71,7 +72,7 @@ def undispatch_orders_package
   end
 
   def update_state_to_designated
-    package.update_allow_web_publish_to_false
+    package.unpublish
```
steveyken: Much clearer!

lib/classes/cart_checkout.rb
```diff
@@ -0,0 +1,94 @@
+#
+# Helper class for the checkout process
+#
+# Process:
+#   1. designage each package to the order
```
steveyken:
```suggestion
#   1. designate each package to the order
```

lib/classes/cart_checkout.rb
```diff
@@ -0,0 +1,94 @@
...
+  end
+
+  # --- Action
+
+  def to_order(order)
+    steps = [
+      -> { validate(order) },
+      -> { add_cart_items_to_order(order) },
+      -> { submit_order(order) }
+    ]
```
steveyken: I like this!

app/controllers/api/v1/cart_items_controller.rb
patrixr: oh noes 😆

lib/classes/cart_checkout.rb
patrixr: me too, love creating steps : D

abhinav098:
@patrixr , Overall a very good implementation. 
But I feel association of cart_items with users is vague.

Ideally how it should be implemented it,
There should be a `cart`, which belongs to the user.
and cart should have many `cart_items`. 

cart_items linked to users is not making sense. 

So how I see it should be is:

```User should have many carts.
Cart belongs to user.```

```Now cart_items, 
cart should have many cart_items and cart_item should belong to a cart.
also, cart_item should belong to package. ```

Having it this way will keep it more readable. 
Also what if we have to fetch a user's purchase ?

Just do user.cart
and if we have to fetch cart_items it should be user.cart.cart_items

user.cart_items is a bit odd to me. 

Please correct me if i'm wrong. And also please let me know your thoughts on this.


Also, team, please let me know your thoughts on this.
cc - @steveyken @namrataukirde @abulsayyad123 @Darshan-Chauhan

**Comment by patrixr** [comment]
@abhinav098 

We thought about it, as it makes sense from a "human readability" perspective. And it sounds right for a user to have a "Cart" object.

But that would really only be useful if the user actually had multiple carts. Which is not the case, there's only one.

So if you added a `cart` that to the user, the `cart` model would essentially just be a useless intermediary to the `cart_items`. Adding an extra table and relationship for no good reason

**Comment by abhinav098** [comment]
No @patrixr , I meant, user will have many carts. not only one. 
Eg. user came first to the app, he ordered something. Cart is created with cart_items. When he submits/checkout the cart. We should persist cart record for a user.

Whenever user requests a new item, a new cart should get created, unless he has not checked out the cart, user should be able to add items to the same cart. When ever user submits the cart, this cart should me marked submitted. 

And whenever the same user comes for next order, new cart must be created, following the same process.

**Comment by patrixr** [comment]
@abhinav098 

There's a slight mixup in how the process works. Let's have a call when you're feeling better so that I can explain. But here's a first explanation :

> When he submits/checkout the cart. We should persist cart record for a user

That's not actually how it works, the cart is not persisted at checkout (by cart, I mean all the `cart_items`), but long before then. What happens at the checkout is that the `orders_packages` are created, and the cart is emptied - no different from what was implemented before, except it happens server-side.

Here's the flow : 

1 - The user clicks on "Request Item" on browse 
2 - A `cart_item` is created and *persisted immediately* on the server (note we're not waiting for the checkout)
3 - User repeats as many times as he/she wishes, populating his/her requested items
4 - The user does a cart checkout procedure
5 - The server creates all the `orders_packages` based on the `cart_items` , and empties the cart
6 - The user now has a submitted order with packages in it, and an empty cart which he she can start adding things to again

And for the rest of the order's lifecycle, the cart is never involved again. The order has no relation to the cart. 

You can think of it as a temporary list of desired items, exactly like we had before, except that instead of storing it in the localstorage, it's in the database.

**Comment by steveyken** [comment]
Good conversation to read.  @abhinav098 describes the traditional shopping cart scenario well. In our case, it might be a bit misleading to call it a cart (I like @patrixr suggestion of renaming to `requested_packages`) as what we have here isn't really a shopping cart but more of a list. In reality,  our order process is more like a traditional shopping cart as it contains packages and has state etc. Perhaps 'wish_list' is another concept to help explain what we have here. (Though I wouldn't recommend changing it to that!)

I'm fine for this PR to go through as is, but also fine if wanting to rename to `requested_packages` first.

**Comment by patrixr** [comment]
> Good conversation to read. @abhinav098 describes the traditional shopping cart scenario well. In our case, it might be a bit misleading to call it a cart (I like @patrixr suggestion of renaming to `requested_packages`) as what we have here isn't really a shopping cart but more of a list. In reality, our order process is more like a traditional shopping cart as it contains packages and has state etc. Perhaps 'wish_list' is another concept to help explain what we have here. (Though I wouldn't recommend changing it to that!)
> 
> I'm fine for this PR to go through as is, but also fine if wanting to rename to `requested_packages` first.

`wish_list` sounds even better ! What do you think @abhinav098  ?

**Comment by namrataukirde** [comment]
Nice discussion @abhinav098,  @patrixr and @steveyken.

Yes, What @abhinav098 is saying is correct and used on most of the e-commerce sites but we have different db structure and slight different scenario's so we might not need cart and cart_items as new db structure.

Whatever is implemented that makes sense to me as it's just for storing what we have in cart on server side. When we are submitting cart then we are actually storing all data against order which means don't need to save cart separately and I am fine with the current implementation.

If @abhinav098 wants then we all can discuss this on call whenever he fell better. :)

**Comment by patrixr** [comment]
Yeah I think the whole confusion originates from the name "cart_items"

File: app/controllers/api/v1/cart_items_controller.rb [hunk 6]
```diff
@@ -0,0 +1,82 @@
...
+      end
+
+      def bool_param(key)
+        return false if params[key].nil?
```
namrataukirde: @patrixr what we are getting with params['key']? Is it "true" in string?

Consider following:
nil.to_s == "true" => false
true.to_s == "true" => true
false.to_s == "true" => false

I don't we need to add this line.

We can make it one-liner. Let me know your thoughts.

app/models/cart_item.rb
```diff
@@ -0,0 +1,38 @@
...
+  validates_uniqueness_of :package_id, scope: :user_id
+
+  # --- Hooks
+
+  before_save :update_availability
+
+  def update_availability
+    self.is_available = package.published? &&
+      package.orders_packages.none? { |pkg| pkg.designated? || pkg.dispatched? } &&
+      package.quantity > 0
```
namrataukirde: package.quantity.positive?

Just ruby way. :p

app/controllers/api/v1/cart_items_controller.rb
patrixr: good point, I was doing that thinking of javascript, where `null.toString()` would explode
Forgot about those ruby quirks :)

app/models/orders_package.rb
namrataukirde: Yup.. its much better and clearer now. Thanks.

app/models/cart_item.rb
patrixr: I really prefer `> 0` though 🤣 
more breezy, prettier code

I'll change it to match the rest, but I don't agree with the "ruby way" for `positive?` haha

app/models/package.rb
```diff
@@ -300,10 +302,18 @@ def delete_associated_packages_locations
...
+  def unpublish
     update(allow_web_publish: false)
   end
 
+  def publish
+    update(allow_web_publish: true)
+  end
+
+  def published?
+    return allow_web_publish.present?
```
namrataukirde: I think we don't need to add return here? 🤔 💭

app/models/package.rb
```diff
@@ -571,6 +586,12 @@ def donor_condition_name
     donor_condition.try(:name_en) || item.try(:donor_condition).try(:name_en)
   end
 
+  def update_carts
+    cart_items.each do |cart_item|
```
namrataukirde: Again just FYI about ruby one-liner way:

cart_items.each(&:update_availability!)

You can keep this as well. Whatever you like :)

lib/classes/cart_checkout.rb
namrataukirde: Nice 👌

app/models/package.rb
patrixr: indeed we don't, I'm just new to ruby so I type it without thinking about it haha

app/models/package.rb
patrixr: cool stuff!

app/models/package.rb
patrixr: updated 👍

lib/classes/cart_checkout.rb
```diff
@@ -0,0 +1,94 @@
...
+      end
+      cart_item.destroy
+    end
+  end
+
+  def designate_package(package, order)
+    results = Designator.new(package, {
+      order_id: order.id,
+      package_id: package.id,
+      quantity: 1,
```
namrataukirde: Do we really want to hardcode this value? 

Because of whatever packages we are publishing on browse will have quantity as 1. Should we consider here the available quantity of package?

I know adding 1 by default is safer right now. But just thinking and would like to know thoughts. :)

lib/classes/cart_checkout.rb
```diff
@@ -0,0 +1,94 @@
...
+    steps.each do |step|
+      step.call()
+      break if errors.any?
+    end
+    errors
+  end
+
+  private
+
+  def initialize(cart_items, ignore_unavailable: false)
```
namrataukirde: I will recommend initialize method not to keep as private. Purpose of the class is we can use it from anywhere else as well. If you still want to keep it's scope limited to inheritance then will recommend making it protected. But I will suggest making it public.

Mostly initialize method written as the first method in class some typical ways which ruby people follow. 😸

patrixr: This is similar to a singleton pattern, we do *not* want the users to initialize it manually.
The api of this class is made to be used with the `designate_cart_items` entry point.

patrixr: It's by design

lib/classes/cart_checkout.rb
patrixr: I had a chat with Steve about it, I think on the long run the `cart_items` (which are being renamed to `requested_packages`) will likely have a quantity field which would apply here.
This is temporary until we support multi-qty

namrataukirde: 👍

lib/classes/cart_checkout.rb
namrataukirde: 👍

PR #768, opened 2019-06-20
Don't send SMS reminders for initial 'thank you for your offer' message

steveyken:
Start the SMS Reminder Sent At clock AFTER the initial system generated 'Thank you for your offer' message has gone out. We don't want to send user SMS about this message.

### DIFF
--- a/app/models/offer.rb
+++ b/app/models/offer.rb
@@ -130,6 +130,7 @@ class Offer < ActiveRecord::Base
 
     before_transition on: :submit do |offer, _transition|
       offer.submitted_at = Time.now
+      offer.created_by.update_attribute(:sms_reminder_sent_at, Time.now + 1.minute) # start the SMS reminder clock from here
     end
 
     before_transition on: :start_review do |offer, _transition|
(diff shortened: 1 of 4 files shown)


steveyken:
I've fast tracked this to live so that we can stop SMS's users when we send 'thank you for your offer' messages. Still need review for inclusion in master branch for this release.

steveyken:
Ref: https://github.com/crossroads/api.goodcity/commits/live and https://github.com/crossroads/api.goodcity/tree/7b4c7725e33f0c40399205edc9a6b7cea3ccf982

app/models/offer.rb
```diff
@@ -130,6 +130,7 @@ def set_initial_state
 
     before_transition on: :submit do |offer, _transition|
       offer.submitted_at = Time.now
+      offer.created_by.update_attribute(:sms_reminder_sent_at, Time.now + 1.minute) # start the SMS reminder clock from here
```
steveyken: @namrataukirde using Time.now works correctly - I've tested this on live.

namrataukirde: Great.. Thanks Steve for checking it. :)

app/services/subscriptions_reminder.rb
```diff
@@ -19,6 +19,8 @@ def generate
   #   its not order related messages
   #   are reviewers will receive SMS only on offers they have created (exclude offers they are subscribed too)
   # If sms_reminder_sent_at is NULL then use created_at so we don't SMS user immediately
+  # IMPORANT NOTE: When a user submits an offer, we set the sms_reminder_sent_at time to now + 1.minute
+  #   to avoid alerting on the system 'thank you for submitting your offer' message
```
namrataukirde: 👍

PR #769, opened 2019-06-21
Gcw 2409 fix logging

steveyken:
Log output is now in format: `key1=value1 key2=value2` which is better ingested by Splunk.

For example:

```
time='2019-06-21T17:25:41.568+08:00' level=INFO method=GET path=/api/v1/offers/252 format=json controller=Api::V1::OffersController action=show status=200 duration=110.69 view=2.49 db=25.91 request_ip=::1 user_id=11 app_name=admin app_version=0.15.6
```

Features:
* More readable by Splunk
* Now includes app_name, app_version and ip address
* Rails logger can now take a hash to make log filtering simpler `Rails.logger.info(key=value)`

e.g. `Rails.logger.info(class: self.class.name, msg: "SMS reminder sent", user_id: user.id)`

Makes it possible to track how many SMS messages we send per user to ensure we don't oversend


steveyken:
hi guys, going to push this through so we can get the logs into Splunk asap and start tracking the SMS reminders more closely. Please review and suggest any comments.

PR #771, opened 2019-06-25
Tweak to simplify permissions.

steveyken:
@abhinav098 this is what i'd propose re security permissions. If you think something it's right, could you write a failing spec to show the case you mean and commit it here? Thanks


abhinav098:
Closing this as per discussion on slack with @steveyken .

PR #774, opened 2019-07-03
Staff can create offers with no donor

patrixr:
If a staff member sends `created_by_id` in the payload, it'll be applied to the created offer.

### DIFF
--- a/app/controllers/api/v1/offers_controller.rb
+++ b/app/controllers/api/v1/offers_controller.rb
@@ -160,7 +160,10 @@ module Api
         attributes = [:language, :origin, :stairs, :parking, :estimated_size,
           :notes, :delivered_by, :state_event, :cancel_reason,
           :cancellation_reason_id, :saleable]
-        attributes << :created_by_id if User.current_user.staff?
+        attributes.concat [
+          :created_at, :created_by_id, :submitted_at, :state,
+          :reviewed_at, :reviewed_by_id
+        ] if User.current_user.staff?
         params.require(:offer).permit(attributes)
       end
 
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/offers_controller.rb
```diff
@@ -39,7 +39,7 @@ class OffersController < Api::V1::ApiController
       api :POST, '/v1/offers', "Create an offer"
       param_group :offer
       def create
-        @offer.created_by = current_user
+        @offer.created_by_id = current_user.id unless offer_params.has_key?(:created_by_id)
```
abhinav098: @patrixr, What if `created_by_id` comes as nil ? i.e. order_params: {created_by_id: nil} ? 
Screen Shot 2019-07-03 at 12 04 02 PM

abhinav098: You can use 
`unless order_params[:created_by_id].present?`
or maybe just 
`unless order_params[:created_by_id]`

steveyken: I think that's the idea. The admin is allowed to set it to nil. This is for offers that come that have no donor (it really happens!). The specs cover the scenario

patrixr: That's exactly what we want. The goal is to allow staff members to create an order with `created_by_id=nil`

patrixr: ref -> https://jira.crossroads.org.hk/browse/GCW-2679

patrixr: Note that the params will prevent donors from doing that

PR #783, opened 2019-07-12
GCW-2605 send submission email

abhinav098:
Hello team,
This PR is for sending emails to the users on order submission for pickup and delivery.
Please review.
Thanks

Will need to add template ids to the .env file for staging and prod both.

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -269,11 +269,11 @@ class Order < ActiveRecord::Base
   end
 
   def send_submission_pickup_email?
-    booking_type.eql?(BookingType.appointment) || order_transport&.pickup?
+    booking_type.appointment? || order_transport&.pickup?
   end
 
   def send_submission_delivery_email?
-    booking_type == BookingType.online_order
+    booking_type.online_order?
   end
 
   def send_order_submission_email
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -272,19 +272,11 @@ class Order < ActiveRecord::Base
     booking_type.appointment? || order_transport&.pickup?
   end
 
-  def send_submission_delivery_email?
-    booking_type.online_order?
-  end
-
   def send_order_submission_email
     return if created_by.nil? || !state.eql?("submitted")
-    sendgrid_instance = SendgridService.new(created_by)
+    type = send_submission_pickup_email? ? "submission_pickup" : "submission_delivery"
     begin
-      if send_submission_pickup_email?
-        sendgrid_instance.send_order_pickup_email self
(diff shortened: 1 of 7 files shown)


app/models/order.rb
```diff
@@ -267,6 +268,28 @@ def can_transition
     return (required_process_checks - process_checklists).empty?
   end
 
+  def send_submission_pickup_email?
+    booking_type.eql?(BookingType.appointment) || order_transport&.pickup?
```
patrixr: Could you use `booking_type.appointment?` instead ?

There is a (very very very unlikely) potential edge case. Since booking types are defined by a string (e.g "appointment), it technically opens up the possibility of having multiple booking types with the same identifier (for example if the seed task is faulty and creates duplicates).

So it's better to verify the identifier (that's what `appointment?` does), as opposed to comparing to another record.

This is really

app/models/order.rb
```diff
@@ -267,6 +268,28 @@ def can_transition
...
+  end
+
+  def send_submission_delivery_email?
+    booking_type == BookingType.online_order
```
patrixr: same here

app/models/order.rb
```diff
@@ -267,6 +268,28 @@ def can_transition
...
+  end
+
+  def send_order_submission_email
+    return if created_by.nil? || !state.eql?("submitted")
+    sendgrid_instance = SendgridService.new(created_by)
+    begin
+      if send_submission_pickup_email?
```
patrixr: Do we need two methods here ?
They seem to do the same thing apart from using a different template ID, and they take the same argument.

maybe a simple 

```
sendgrid_instance.send_order_submission_email(self, type: type_of_the_order)
```

Really not a big deal, but the whole point of the `template_id ` mapping in the the sendgrid service, is to not have an `if/else` flow here. I think the method generation, although fun, is not bringing too much value

app/models/order.rb
abhinav098: yeah, it makes sense. Please re-check, I've changed it.

patrixr: looks good !

app/models/order.rb
```diff
@@ -267,6 +268,28 @@ def can_transition
     return (required_process_checks - process_checklists).empty?
   end
 
+  def send_submission_pickup_email?
+    booking_type.appointment? || order_transport&.pickup?
```
patrixr: Appointment and pickup have the same email ? 🤔

abhinav098: yep, https://jira.crossroads.org.hk/browse/GCW-2605

PR #784, opened 2019-07-15
redirect users to packages controller if they hit `browse/fetch_packages`

abhinav098:
For old mobile apps. Tested it using postman hit the api endpoint for 
`localhost:4000/api/v1/browse/fetch_packages` and `localhost:4000/api/v1/packages`
Working fine for me.


abhinav098:
Added a comment @patrixr .

PR #786, opened 2019-07-15
GCW-2683-Allow to add items to submitted order

abulsayyad123:
Hi Team,
This PR is for JIRA ticket GCW-2683- Allow to add items to submitted orders. Please review this PR and let me know if any change is required.

### DIFF
--- a/lib/classes/cart_checkout.rb
+++ b/lib/classes/cart_checkout.rb
@@ -54,7 +54,7 @@ class CartCheckout
 
   def validate(order)
     return errors.add(:base, I18n.t('cart.bad_order')) if order.blank?
-    return errors.add(:base, I18n.t('cart.already_processed')) unless active_orders.include?(order.state)
+    return errors.add(:base, I18n.t('cart.already_processed')) unless Order::NON_PROCESSED_STATES.include?(order.state)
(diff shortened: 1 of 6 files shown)


lib/classes/cart_checkout.rb
```diff
@@ -91,4 +91,8 @@ def designate_package(package, order)
   def add_errors(errs)
     errs.full_messages.each { |m| errors.add(:base, m) }
   end
+
+  def active_orders
```
patrixr: I think the `order` model already has lists of states (by type)
Maybe you can use those directly (e.g `Order::ACTIVE_STATES`) to centralise the types of states

spec/controllers/api/v1/requested_packages_controller_spec.rb
```diff
@@ -182,6 +184,24 @@
...
+        it "returns submitted order if the order is already submitted" do
+          post :checkout, order_id: submitted_order.id
+          expect(response.status).to eq(200)
+
+          order = Order.find(submitted_order.id)
+          expect(order.state).to eq('submitted')
+          expect(order.orders_packages.length).to eq(requested_packages.length)
+        end
+
+        it "returns processing order if the order is already processed" do
```
patrixr:
```suggestion
it "doesn't modify the state of a processing order" do
```

maybe clearer ?

PR #789, opened 2019-07-16
Restrict user to view other user's order

abhinav098:
This PR is a fix for the issue which was discussed in the call. (i.e Restrict user to view other user's order)
`rake 'goodcity:remove_permission_from_role[permission_name, role_name]'`
eg: 
`rake 'goodcity:remove_permission_from_role[can_manage_orders, Charity]'`

### DIFF
--- a/lib/tasks/remove_can_manage_order_permission_from_charity_users.rake
+++ b/lib/tasks/remove_can_manage_order_permission_from_charity_users.rake
@@ -1,0 +1,7 @@
+namespace :goodcity do
+  task remove_can_manage_order_permission_from_charity_users: :environment do
+    role_permission = Permission.find_by(name:"can_manage_orders").role_permissions.where(role_id: Role.find_by(name: "Charity").id).first
+
+    role_permission.destroy if role_permission
+  end
+end
--- a/lib/tasks/remove_can_manage_order_permission_from_charity_users.rake
+++ b/lib/tasks/remove_can_manage_order_permission_from_charity_users.rake
@@ -1,7 +1,0 @@
-namespace :goodcity do
-  task remove_can_manage_order_permission_from_charity_users: :environment do
-    role_permission = Permission.find_by(name:"can_manage_orders").role_permissions.where(role_id: Role.find_by(name: "Charity").id).first
-
-    role_permission.destroy if role_permission
-  end
-end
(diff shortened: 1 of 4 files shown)


lib/tasks/remove_can_manage_order_permission_from_charity_users.rake
```diff
@@ -0,0 +1,7 @@
+namespace :goodcity do
+  task remove_can_manage_order_permission_from_charity_users: :environment do
+    role_permission = Permission.find_by(name:"can_manage_orders").role_permissions.where(role_id: Role.find_by(name: "Charity").id).first
```
namrataukirde: Space missing after colon.

lib/tasks/remove_can_manage_order_permission_from_charity_users.rake
```diff
@@ -0,0 +1,7 @@
...
+
+    role_permission.destroy if role_permission
```
namrataukirde: Use safe navigation (`&.`) instead of checking if an object exists before calling the method.

lib/tasks/remove_permission_from_role.rake
```diff
@@ -0,0 +1,10 @@
+# to run this rake task run the following command.
+# rake goodcity:remove_permission_from_role permission=permission_name role=role_name
+
+namespace :goodcity do
+  task remove_permission_from_role: :environment do
+    Permission.find_by(name: ENV["permission"])
```
patrixr: I think rake supports command line arguments, maybe more appropriate than env variables. Not a big deal though

abhinav098: >command line arguments 

@patrixr , If i run `rake goodcity:remove_permission_from_role permission=abc role=xyz`
These are available as `ENV` variables 
i.e.  
`ENV['permission']="abc"`
`ENV['role']="xyz"`

patrixr: I know, it's fine. It just that ENV variables generally tend to serve a different purpose than args. So it's more "proper" to pass in arguments. But doesn't really make a practical difference

patrixr: no need to change anything, just thought I'd point it out : )

PR #790, opened 2019-07-16
Added all search on all states if no states selected 

abulsayyad123:
Hi Team,
This PR is for removing search on active offers if no states selected in filters. If no states selected it should search on all offers. I have removed `where` clause for states in `offers_search` because there is already `states` filter in `offer_filters.rb` file (https://github.com/crossroads/api.goodcity/blob/master/app/models/concerns/offer_filtering.rb#L14).

### DIFF
--- a/app/models/concerns/offer_search.rb
+++ b/app/models/concerns/offer_search.rb
@@ -12,7 +12,6 @@ module OfferSearch
     # package_type: name_en, name_zh_tw
 
     scope :search, -> (options = {}) {
-      offer_states = options[:states].empty? ? Offer.nondraft_states : options[:states]
       search_text = options[:search_text] || ''
       search_query = ['offers.notes', 'users.first_name', 'users.last_name',
          'users.email', 'users.mobile', 'items.donor_description',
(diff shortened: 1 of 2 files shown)


app/models/concerns/offer_search.rb
```diff
@@ -12,7 +12,6 @@ module OfferSearch
     # package_type: name_en, name_zh_tw
 
     scope :search, -> (options = {}) {
-      offer_states = options[:states].empty? ? Offer.nondraft_states : options[:states]
```
patrixr: doesn't that remove the ability to filter by state altogether ? `options[:states]` seems to be ignored completely.
Maybe I'm no understanding the PR right 🤔

abulsayyad123: I have added state filter code in offer filtering (https://github.com/crossroads/api.goodcity/blob/master/app/models/concerns/offer_filtering.rb#L14) so there was duplication of state filters. In offer filtering it will be required in future for dashboard states. So, I thought it would be better to remove from it from `search_filtering.rb`

PR #791, opened 2019-07-16
send sms to only mobile numbers

namrataukirde:
Hi Team,

This PR will avoid calling Twilio job if user does not have a mobile number and will fix the issue of trying to send SMS on email id.

Please review.

Thanks.

### DIFF
--- a/app/services/twilio_service.rb
+++ b/app/services/twilio_service.rb
@@ -31,10 +31,15 @@ class TwilioService
   # options[:to] = "+85261111111"
   # options[:body] = "SMS body"
   def send_sms(options)
-    options = {to: @user.mobile}.merge(options)
+    to = @user.mobile.presence
+
+    options = { to: to }.merge(options)
+
+    return unless options[:to]
+
     if send_to_twilio?
       TwilioJob.perform_later(options)
-    elsif options[:to].presence
+    else
       message = "SlackSMS (to: #{options[:to]}, id: #{user.id}, full_name: #{user.full_name}) #{options[:body]}"
       SlackMessageJob.perform_later(message, ENV['SLACK_PIN_CHANNEL'])
     end
--- a/app/services/twilio_service.rb
+++ b/app/services/twilio_service.rb
@@ -31,11 +31,10 @@ class TwilioService
   # options[:to] = "+85261111111"
   # options[:body] = "SMS body"
   def send_sms(options)
-    to = @user.mobile.presence || @user.email.presence
-    options = {to: to}.merge(options)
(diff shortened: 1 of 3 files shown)


app/services/twilio_service.rb
```diff
@@ -31,11 +31,10 @@ def send_unread_message_reminder(url)
   # options[:to] = "+85261111111"
   # options[:body] = "SMS body"
   def send_sms(options)
-    to = @user.mobile.presence || @user.email.presence
-    options = {to: to}.merge(options)
+    options = {to: @user.mobile}.merge(options)
```
steveyken: Great, yes that's fine. might be worth failing if no 'to' field as it's not valid to continue.
What you've implement is more along the lines of defensive programming - preventing an error from occurring but in this case, i think we do want the program to crash so we can find the bug quickly as there should never be a case of sending SMS to a user with no mobile.

```suggestion
    options = {to: @user.mobile}.merge(options)
    raise SomeError.new("User mobile missing") unless options[:to].present?
    
    if send_to_twilio?
    ...
    else
      send to slack
    end
```

steveyken: just a thought

namrataukirde: Thanks for review. Let me think on this. :)

namrataukirde: Yes @steveyken. It's good idea to send error message to rollbar. I will add it. Thanks.

spec/services/twilio_service_spec.rb
```diff
@@ -57,13 +59,21 @@
...
     it "sends order submitted acknowledgement to charity who submitted order" do
       allow(twilio).to receive(:send_to_twilio?).and_return(true)
       body = "Thank you for placing order #{order.code} on GoodCity. Our team will be in touch with you soon.\n"
       expect(TwilioJob).to receive(:perform_later).with(to: user.mobile, body: body)
       twilio.order_confirmed_sms_to_charity(order)
     end
+
+    it "do not sends order submitted acknowledgement via sms to charity without mobile who submitted order" do
+      expect(twilio_with_no_mobile_user).not_to receive(:send_to_twilio?)
+      body = "Thank you for placing order #{order_for_charity_without_mobile.code} on GoodCity. Our team will be in touch with you soon.\n"
```
steveyken: no need for `body` here

namrataukirde: Sure.. :)

namrataukirde: Removed this unnecessary code.

PR #796, opened 2019-07-25
GCW-2685- Admin Dashboard

abulsayyad123:
Hi Team,
This PR is for ticket GCW-2685- Admin Dashboard. This PR includes:
1) Offers count Active States for All Offers and logged in User. I have added separate endpoint 'summary' for this.
2) Recent Offers query for fetching recent 5 Offers according state (it will be 'submitted' in this case).
3) Removed 'draft' Offers in search and filter query.
4) Test cases for Recent Offers and Offers summary (Offers Count).

### DIFF
--- a/app/controllers/api/v1/offers_controller.rb
+++ b/app/controllers/api/v1/offers_controller.rb
@@ -140,8 +140,9 @@ module Api
       end
 
       def summary
-        priority_and_non_priority_active_offers_count = Offer.non_priority_active_offers_count.merge(
-          Offer.priority_active_offers_count
+        self_reviewer = bool_param(:selfReview, false)
+        priority_and_non_priority_active_offers_count = Offer.non_priority_active_offers_count(self_reviewer=self_reviewer).merge(
+          Offer.priority_active_offers_count(self_reviewer=self_reviewer)
         )
         render json: priority_and_non_priority_active_offers_count
       end
@@ -173,6 +174,7 @@ module Api
           state_names: array_param(:state),
           priority: bool_param(:priority, false),
           self_reviewer: bool_param(:selfReview, false),
+          recent_offers: bool_param(:recent_offers, false),
           before: time_epoch_param(:before),
           after: time_epoch_param(:after)
         })
--- a/app/controllers/api/v1/offers_controller.rb
+++ b/app/controllers/api/v1/offers_controller.rb
@@ -72,7 +72,7 @@ module Api
       def search
         records = @offers.search({ search_text: params['searchText'], states: array_param(:state) })
         records = apply_filters(records)
-        records = records.page(params["page"]).per(params["per_page"] || params["recent_offer_per"]|| DEFAULT_SEARCH_COUNT)
+        records = records.page(params["page"]).per(params["per_page"] || params["recent_offer_count"] || DEFAULT_SEARCH_COUNT)
         offers = offer_response(records.with_summary_eager_load)
         render json: {meta: {total_pages: records.total_pages, search: params['searchText']}}.merge(offers)
       end
--- a/app/controllers/api/v1/offers_controller.rb
+++ b/app/controllers/api/v1/offers_controller.rb
@@ -72,7 +72,7 @@ module Api
       def search
         records = @offers.search({ search_text: params['searchText'], states: array_param(:state) })
         records = apply_filters(records)
-        records = records.page(params["page"]).per(params["per_page"] || DEFAULT_SEARCH_COUNT)
+        records = records.page(params["page"]).per(params["per_page"] || params["recent_offer_per"]|| DEFAULT_SEARCH_COUNT)
         offers = offer_response(records.with_summary_eager_load)
         render json: {meta: {total_pages: records.total_pages, search: params['searchText']}}.merge(offers)
       end
@@ -140,9 +140,8 @@ module Api
       end
 
       def summary
-        self_reviewer = bool_param(:selfReview, false)
-        priority_and_non_priority_active_offers_count = Offer.non_priority_active_offers_count(self_reviewer=self_reviewer).merge(
-          Offer.priority_active_offers_count(self_reviewer=self_reviewer)
+        priority_and_non_priority_active_offers_count = Offer.priority_and_non_priority_offers_count_for(self_reviewer: false).merge(
+          Offer.priority_and_non_priority_offers_count_for(self_reviewer: true)
         )
         render json: priority_and_non_priority_active_offers_count
       end
--- a/app/controllers/api/v1/offers_controller.rb
+++ b/app/controllers/api/v1/offers_controller.rb
@@ -140,10 +140,10 @@ module Api
       end
 
       def summary
-        priority_and_non_priority_active_offers_count = Offer.priority_and_non_priority_offers_count_for(self_reviewer: false).merge(
-          Offer.priority_and_non_priority_offers_count_for(self_reviewer: true)
+        all_offers_count = Offer.offers_count_for(self_reviewer: false).merge(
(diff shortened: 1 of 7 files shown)


app/controllers/api/v1/offers_controller.rb
```diff
@@ -139,6 +139,13 @@ def mark_inactive
         render json: @offer, serializer: offer_serializer
       end
 
+      def summary
+        priority_and_non_priority_active_offers_count = Offer.priority_and_non_priority_offers_count_for(self_reviewer: false).merge(
```
patrixr: if the function returns the priority count AND the non priority count, isn't that the same as all count ?? 😄 
Therefore shouldn't the function be named `offers_count_for` ?

app/models/concerns/offer_filtering.rb
```diff
@@ -47,5 +49,11 @@ def self.last_6pm
       t -= 24.hours if now < t
       t
     end
+
+    def self.assoicate_delivery_and_schedule
+      res = joins("LEFT OUTER JOIN deliveries ON offers.id = deliveries.offer_id")
+      res = res.joins("LEFT OUTER JOIN schedules ON deliveries.schedule_id = schedules.id")
+      res
```
patrixr:
```ruby
def self.assoicate_delivery_and_schedule
   joins <<-SQL
      LEFT OUTER JOIN deliveries ON offers.id = deliveries.offer_id
      LEFT OUTER JOIN schedules ON deliveries.schedule_id = schedules.id
   SQL
end
```
absolutely nothing wrong with this function, just thought I'd share my "esthetic" preferences : )

app/controllers/api/v1/offers_controller.rb
patrixr: actually that comment applies more to the implementation of the method below (poorly placed)

app/models/concerns/offer_filtering.rb
patrixr: (don't change anything, just for the sake of sharing)

app/models/offer.rb
```diff
@@ -207,6 +208,43 @@ def nondraft_states
     def donor_states
       valid_states - ["draft"]
     end
+
+    def priority_and_non_priority_offers_count_for(self_reviewer: false)
```
patrixr: @abulsayyad123 Could you share a sample json output for the requests, I have trouble understanding the key renaming and all

abulsayyad123: Sure,
```
{
  active_offers_total_count: 166
  priority_receiving: 2
  priority_reviewed: 127
  priority_scheduled: 13
  priority_under_review: 6
  receiving: 2
  reviewed: 127
  reviewer_active_offers_total_count: 4
  reviewer_priority_under_review: 2
  reviewer_under_review: 4
  scheduled: 13
  under_review: 24
}
```

app/controllers/api/v1/offers_controller.rb
```diff
@@ -3,7 +3,7 @@
 module Api
   module V1
     class OffersController < Api::V1::ApiController
-      before_action :eager_load_offer, except: [:index, :create, :search]
+      before_action :eager_load_offer, except: [:index, :create, :search, :summary]
```
abhinav098: Use `%i` or `%I` for an array of symbols.

app/models/offer.rb
```diff
@@ -7,7 +7,8 @@ class Offer < ActiveRecord::Base
   include OfferSearch
   include OfferFiltering
 
-  NOT_ACTIVE_STATES = ["received", "closed", "cancelled", "inactive"]
+  NOT_ACTIVE_STATES = ["received", "closed", "cancelled", "inactive"].freeze
```
abhinav098: Use `%w` or `%W` for an array of words. eg `%w[a b c d] => ["a", "b", "c","d"]`

app/models/offer.rb
abhinav098: please change this method's name, as it is not just sending `priority_and_non_priority_offers_count_for` as json, but other parameters too.

app/models/offer.rb
```diff
@@ -7,7 +7,8 @@ class Offer < ActiveRecord::Base
...
+  ACTIVE_OFFERS = ["under_review", "reviewed", "scheduled", "receiving"].freeze
```
abhinav098: Use %w or %W for an array of words

app/models/offer.rb
```diff
@@ -207,6 +208,43 @@ def nondraft_states
...
+      offers = filter(state_names: ACTIVE_OFFERS, priority: is_priority, self_reviewer: self_reviewer).group_by(&:state)
+      if is_priority
+        offers_count_per_state(offers).transform_keys { |key| "priority_".concat(key) }
+      else
+        offers_count_per_state(offers)
+      end
+    end
+
+    def offers_count_per_state(offers)
+      offers.each { |key, value|  offers[key] = value.count }
```
patrixr: maybe rename the `offers` variable, it made me think it was an array of offers (actually a hash of states)

app/models/offer.rb
patrixr: Instead of creating the hashes and then renaming keys, you can maybe pass a key prefix to the methods

pseudo code
```ruby
def counter_function(filters, key_prefix: '')
  # ... returns a hash
end

def caller_function
    res = {}
    res.merge! counter_function({ priority: false })
    res.merge! counter_function({ priority: true }, key_prefix: 'priority_')
    res
end

```

patrixr: Also maybe the states could be passed as a be passed as a parameter. Currently the requirements are for `active_states`, but that might change in the future. 

To extend the pseudo code from above :

```ruby


def counter_function(filters, key_prefix: '')
  # ... returns a hash
end

def caller_function
    res = {}
    res.merge! counter_function({ states: ACTIVE_STATES, priority: false })
    res.merge! counter_function({ states: ACTIVE_STATES, priority: true }, key_prefix: 'priority_')
    res
end
```

app/models/offer.rb
patrixr: I'll confess I'm not a big fan of the `%w` ruby pattern. Because :
- I find the difference in readability to be minor
- it obfuscates the data type 
- it isn't explicit in what it does, unlike everything else in ruby with clear expressions such as `unless`, `do`
- it isn't standard across any other language so that creates an extra layer of confusion for ruby learners such as myself.

Maybe I'm just missing the point of it altogether 🤔
Ruby guru @steveyken , would you know actually is the point 😆

I'm fine with either way, just sharing my thoughts and trying to learn : )

app/models/offer.rb
```diff
@@ -207,6 +208,34 @@ def nondraft_states
...
+
+    def prepend_offer_key_with(prefix, offer)
+      offer.transform_keys { |key| "#{prefix}_#{key}" }
+    end
+
+    def offers_count_sum(offer)
+      offer.values.reduce(:+)
+    end
+
+    def offers_count_per_state(offer)
```
patrixr: Quick warning on those 2 functions here

You're receiving an argument and modifying it. 
That's always a dangerous practice, we want to write pure functions as much as possible to avoid potential side effect (big part of state management).

It's fairly contained in this case, which is fine. But generally if you are not 100% sure of the safety of that operation it's better not to. (better safe than sorry)

I try to force myself to never modify any variable (as much as I can).
Our server is stateless, so applying that in our code makes sense.

patrixr: not a change request. just a topic I'm passionate about

app/models/offer.rb
steveyken: not sure on the etymology of %w.... sorry!

PR #797, opened 2019-07-26
GCW-2678 Online order confirmation mail

patrixr:
An email sent when an online order finishes processing

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -278,19 +278,26 @@ class Order < ActiveRecord::Base
   def send_order_submission_email
     return if created_by.nil? || !state.eql?("submitted")
     type = send_submission_pickup_email? ? "submission_pickup" : "submission_delivery"
-    begin
-      SendgridService.new(created_by).send_order_submission_email(self, type)
-    rescue => e
-      Rollbar.error(e, error_class: "Sendgrid Error", error_message: "Sendgrid submission email")
-    end
+    SendgridService.new(created_by).send_email_for_order(self, type)
   end
 
   def send_confirmation_email
-    return if booking_type != BookingType.appointment || created_by.nil?
-    begin
-      SendgridService.new(created_by).send_appointment_confirmation_email(self, "appointment_confirmation")
-    rescue => e
-      Rollbar.error(e, error_class: "Sendgrid Error", error_message: "Sendgrid confirmation email")
+    send_appointment_confirmation_email if booking_type.appointment?
+    send_online_order_confirmation_email if booking_type.online_order?
+  end
+
+  def send_appointment_confirmation_email
+    return unless booking_type.appointment? && created_by.present?
+    SendgridService.new(created_by).send_appointment_confirmation_email(self)
+  end
+
+  def send_online_order_confirmation_email
+    return unless booking_type.online_order? && created_by.present?
+    email_service = SendgridService.new(created_by)
+    if order_transport.pickup?
+      email_service.send_order_confirmation_pickup_email(self)
+    else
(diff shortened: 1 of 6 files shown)


app/services/sendgrid_service.rb
```diff
@@ -1,7 +1,13 @@
 class SendgridService
   attr_accessor :user, :template_name, :substitution_hash
 
-  MAIL_METHODS = %w[send_appointment_confirmation_email send_order_submission_email]
+  MAIL_METHODS = [
+    { name: :send_appointment_confirmation_email, template: 'appointment_confirmation' },
+    { name: :send_order_submission_pickup_email, template: 'submission_pickup' },
```
abhinav098: this method is not getting called from anywhere.

app/services/sendgrid_service.rb
```diff
@@ -1,7 +1,13 @@
...
+    { name: :send_order_submission_delivery_email, template: 'submission_delivery' },
```
abhinav098: Same for this one. From order.rb, `send_email_for_order` is called directly. We either should call these methods from the the order model, or should remove these two items from here.

patrixr: @abhinav098  Well I hesitated, it's either one call with passing the argument, or an `if/else` . Have no strong preference. I added them regardless for consistency

patrixr: Let me know if you'd rather have the explicit call

abhinav098: I think its better to call the methods explicitly. instead of directly accessing the `send_email_order` method. @patrixr .

abhinav098: I would recommend to make this method to `private`.

app/models/order.rb
```diff
@@ -278,20 +278,25 @@ def send_submission_pickup_email?
   def send_order_submission_email
     return if created_by.nil? || !state.eql?("submitted")
     type = send_submission_pickup_email? ? "submission_pickup" : "submission_delivery"
-    begin
-      SendgridService.new(created_by).send_order_submission_email(self, type)
-    rescue => e
-      Rollbar.error(e, error_class: "Sendgrid Error", error_message: "Sendgrid submission email")
-    end
+    SendgridService.new(created_by).send_email_for_order(self, type)
```
abhinav098: Better to do 
```
def send_order_submission_email
    return if created_by.nil? || !state.eql?("submitted")
    sendgrid_instance = SendgridService.new(created_by)
    return sendgrind_instance.send_order_submission_pickup_email if send_submission_pickup_email?    
    sendgrind_instance.send_order_submission_delivery_email
end
```

app/services/sendgrid_service.rb
```diff
@@ -94,6 +103,20 @@ def sendgrid_email_formation(email, name = nil)
...
+  def send_email_for_order(order, template_name)
+    return unless user.email.present?
+    begin
+      @add_bcc = true
+      substitution_hash.merge!(user.email_properties)
+      substitution_hash.merge!(order.email_properties)
+      @mail.from = sendgrid_email_formation(ENV["APPOINTMENT_FROM_EMAIL"], I18n.t("email_from_name"))
+      @mail.template_id = template_id(template_name)
+      send_email
+    rescue => e
```
abhinav098: Avoid rescuing without specifying an error class.

patrixr: This code was written in orders, I just moved it. I don't know what the error class would be

patrixr:
Merging this now, Matt wants to test - Will retroactively fix other comments

PR #800, opened 2019-08-07
GCW-2690 handle heic images and add a rake for fixing old images

abhinav098:
Hello Team, 
This PR handles the heic images coming from new iphones.
Please review
Thanks

### DIFF
--- a/lib/tasks/handle_heic_images.rake
+++ b/lib/tasks/handle_heic_images.rake
@@ -1,0 +1,9 @@
+# rake goodcity:handle_heic_images
+
+namespace :goodcity do
+  task handle_heic_images: :environment do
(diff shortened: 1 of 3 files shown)


lib/tasks/handle_heic_images.rake
```diff
@@ -0,0 +1,9 @@
+# rake goodcity:handle_heic_images
+
+namespace :goodcity do
```
namrataukirde: Could you please add logs to this rake?

abhinav098: sure

abhinav098: added the logs @namrataukirde .

PR #801, opened 2019-08-08
Messages by scope

patrixr:
Addition of a few query params to help with message loading (backwards compatible)

### Messages by scope

The admin app is currently doing `/messages?offer_ids=< all the offer ids >` 

However since we're removing the preload of offers, that won't work anymore.

This PR adds this version of the api call --> `/messages?scope=offer&state=<unread|read>` , and also adds support for pagination

### Marking all messages as read

Currently the admin app iterates over all the messages and marks them individually as read, that won't work any more since we won't load all the messages at once. Also we have seen hundreds of requests firing in the past, which we want to avoid

So this PR adds a new endpoint: `/messages/mark_all_read?scope=offer`

### Offers by notifications

The current paginated and filtered offers endpoint already exists : `/offers/search?<filters>`, this PR adds the following filters when listing offers  --> `/offers/search?with_notifications=<unread|read|all>`


app/controllers/api/v1/messages_controller.rb
```diff
@@ -58,8 +62,38 @@ def mark_read
...
+        return records unless ALLOWED_SCOPES.include? scope
+        records.where("messages.#{scope}_id IS NOT NULL")
+      end
+
+      def paginate_and_render(records)
+        meta = {}
+        if params[:page].present?
+          records = records
+            .page(params["page"])
+            .per(params["per_page"] || DEFAULT_SEARCH_COUNT)
```
steveyken: Good to put an upper limit on `per_page` and `page`... e.g. query param could be `?per_page=100000`

app/controllers/api/v1/messages_controller.rb
```diff
@@ -58,8 +62,38 @@ def mark_read
...
+        end
+        @subscriptions.update_all state: 'read'
+        render json: {}
+      end
+
       private
 
+      def apply_scope(records, scope)
+        return records unless ALLOWED_SCOPES.include? scope
+        records.where("messages.#{scope}_id IS NOT NULL")
```
steveyken: This might get flagged up in auto scanning tools for SQL injection but I can see you've limited scope options.

app/models/ability.rb
```diff
@@ -182,7 +182,7 @@ def message_abilities
         message.offer.created_by_id == @user_id && !message.is_private
       end
     end
-    can [:mark_read], Message, id: @user.subscriptions.pluck(:message_id)
+    can [:mark_read, :mark_all_read], Message, id: @user.subscriptions.pluck(:message_id)
```
steveyken: `mark_all_read` is a route collection action not a member e.g. `/messages/mark_all_read` rather than `/messages/1/mark_all_read` so in this case the ability for a user to mark all their messages as read can be given to every logged in user. You are scoping the messages in the controller query to just `User.current_user` so that's ok.

```suggestion
    can [:mark_read], Message, id: @user.subscriptions.pluck(:message_id)
    can [:mark_all_read]
```

app/models/subscription.rb
```diff
@@ -8,4 +8,5 @@ class Subscription < ActiveRecord::Base
   after_create :send_new_message_notification # PushUpdatesForSubscription
 
   scope :unread, -> { where(state: 'unread') }
+  scope :of_active_user, -> { where(user_id: User.current_user.id) }
```
steveyken: If you pass in the User.current_user.id as a param then it's a useful scope for elsewhere.

```suggestion
  scope :for_user(user_id), -> { where(user_id: user_id) }
```

spec/controllers/api/v1/messages_controller_spec.rb
```diff
@@ -32,6 +32,15 @@
       expect(subject['messages'].length).to eq(2)
     end
 
+    it "supports pagination", :show_in_doc do
+      8.times { create :message, item: item }
+      get :index, page: 1, per_page: 6
+      expect(response.status).to eq(200)
+      expect(subject['meta']['total_pages']).to eq(2)
+      expect(subject['meta']['total_count']).to eq(8)
+      expect(subject['messages'].length).to eq(6)
```
steveyken: nice!

app/controllers/api/v1/messages_controller.rb
patrixr: that's not a bad idea, any idea how that could be implemented globally ? (guessing a concern)

steveyken: We already have, check  out https://github.com/crossroads/api.goodcity/blob/master/app/controllers/api/v1/api_controller.rb#L53

so you can just change the line to `.per(per_page)` and it will grab the param and check it for you

patrixr: @steveyken Cool, although that doesn't look right, `DEFAULT_SEARCH_COUNT` should not be the maximum

app/controllers/api/v1/api_controller.rb
```diff
@@ -52,7 +52,7 @@ def page
     # max limit is 25
     def per_page
       @per_page = params['per_page'].to_i
-      (@per_page.zero? || @per_page > DEFAULT_SEARCH_COUNT) ? DEFAULT_SEARCH_COUNT : @per_page
+      (@per_page.zero? || @per_page > MAX_SEARCH_COUNT) ? MAX_SEARCH_COUNT : @per_page
```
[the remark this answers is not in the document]
abhinav098: Btw if you want 50 records only for `messages`, you can always override this method in the `messages_controller`.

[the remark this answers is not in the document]
patrixr: oh good point, didn't read through the code. No I want 50 as a maximum, the 25 is the "default" so it was wrongly used

app/models/concerns/offer_filtering.rb
```diff
@@ -42,6 +43,13 @@ def self.due_before(time)
       where('schedules.scheduled_at <= (?)', time)
     end
 
+    def self.with_notifications(state)
+      res = joins("LEFT OUTER JOIN subscriptions ON offers.id = subscriptions.offer_id")
+      res = res.where("subscriptions.user_id = (?)", User.current_user.id)
+      res = res.where("subscriptions.state = (?)", state) if ['unread', 'read'].include?(state)
```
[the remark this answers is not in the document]
patrixr: I respectfully disagree that `%w` looks good 🤣

patrixr:
@abhinav098 updated 👍

spec/controllers/api/v1/api_controller_spec.rb
```diff
@@ -66,9 +66,9 @@ def index
       it { expect(subject).to eql(20) }
     end
 
-    context "30 per_page (limit is 25)" do
-      let(:per_page) { '30' }
-      it { expect(subject).to eql(25) }
+    context "30 per_page (limit is 50)" do
+      let(:per_page) { '60' }
+      it { expect(subject).to eql(50) }
```
steveyken: This spec is fine. Did you also know you can use `MAX_SEARCH_COUNT` here... depends on whether you think a spec should be declarative or narrative! I'm fine either way :)

app/controllers/api/v1/api_controller.rb
```diff
@@ -49,10 +49,10 @@ def page
       @page.zero? ? 1 : @page
     end
 
-    # max limit is 25
+    # max limit is 50, default is 25
     def per_page
       @per_page = params['per_page'].to_i
-      return DEFAULT_SEARCH_COUNT if @per_page.zero?
+      return DEFAULT_SEARCH_COUNT if @per_page.zero? || @per_page.negative?
```
steveyken: or `@per_page < 1` ?!

patrixr: in case of `?per_page=-123` , not really needed but I'm filtering out all invalid values

spec/controllers/api/v1/api_controller_spec.rb
patrixr: I thought about it.... I'd rather the spec dictates the implementation rather than the other way around 👍

app/controllers/api/v1/api_controller.rb
patrixr: invalid values -> use default 
large values -> use max

app/models/concerns/user_search.rb
```diff
@@ -18,10 +18,10 @@ module UserSearch
     }
 
     def self.search_user(role_name, search_query, search_text)
-      joins(:roles)
-      .where('roles.name = ?', role_name)
-      .where(search_query, search_text: "%#{search_text}%")
-      .distinct
+      res = joins(:roles)
```
patrixr: @steveyken @abhinav098 I've added this change to this PR.

Since donors don't have a specific roles, we weren't able to find them through the search endpoint. So I made `role_name` optional.  Do you foresee any unexpected impact this could have ?

abhinav098:
👍

PR #808, opened 2019-08-23
Gcw 2763 handle errors graciously

abhinav098:
Send errors in response properly mapped to keys.

### DIFF
--- a/app/controllers/api/v1/api_controller.rb
+++ b/app/controllers/api/v1/api_controller.rb
@@ -29,7 +29,7 @@ module Api
     end
 
     def render_error(error_message, code: 422)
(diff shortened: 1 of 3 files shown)


app/controllers/api/v1/api_controller.rb
```diff
@@ -29,7 +29,8 @@ def save_and_render_object_with_errors(object)
     end
 
     def render_error(error_message, code: 422)
-      render json: { errors: error_message }, status: code
+      errors = { :message => error_message, :status => code }
```
namrataukirde: Use the new Ruby 1.9 hash syntax.

abhinav098: sure.

PR #812, opened 2019-09-02
Resize images on cloudinary

patrixr:
Our Cloudinary account was reaching full capacity (**97%**)

A lot of raw images matched the camera's native resolution (one of them was around 7000x4000)

This rake tasks takes the 500 largest images and downsizes them to FHD.

I ran this a few times:

`bundle exec rake cloudinary:optimize`

And voila ! Magic :

Screenshot 2019-09-02 at 3 23 50 PM

As of now we're at **40%** capacity, with still a thousand images to downsize. Looking good

Next step: Resizing the images before upload

### DIFF
--- a/lib/tasks/cloudinary.rake
+++ b/lib/tasks/cloudinary.rake
@@ -42,6 +42,103 @@ namespace :cloudinary do
       end
   end
 
+  MAX_SIZE_MB = 1
+  MAX_SIZE_BYTES = MAX_SIZE_MB * 1024 * 1024
+  MAX_SIZE = 1920
+  OPTIMIZE_PAGE_SIZE = 500
+
+  desc "Optimize image sizes"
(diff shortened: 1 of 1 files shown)


patrixr:
> Thanks! Image sizes are much more manageable now.
> Small bug (but unlikely to be an issue), if the first 500 images are placeholders then it won't take any action. Is that a likely scenario? Can you exclude placeholders from API search?

@steveyken 

> Thanks! Image sizes are much more manageable now.
> Small bug (but unlikely to be an issue), if the first 500 images are placeholders then it won't take any action. Is that a likely scenario? Can you exclude placeholders from API search?

@steveyken I believe `resource_type:image` excludes placeholder. Haven't seen as single one so far. I kept the manual check "just in case"

abhinav098:
:+1:

PR #815, opened 2019-09-10
Look in packages table for existing inventory numbers.

steveyken:
This is the fix to ensure that inventory numbers reserved in the packages table are taken into account when choosing the next one.

There is one spec failure because the code is returning '000000' as the next number. If you create a package with inventory number '000000' then the spec passes. I think we should adjust the code to start counting from 1 rather than 0.

@patrixr  can I hand this to you?


patrixr:
Fixed the broken tests, can be reviewed now

PR #817, opened 2019-09-11
use save with errors method

abhinav098:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-2691

### What does this PR do?

BUG: show actual error for company controller instead of something went wrong

### Impacted Areas

Company creation page for corporate donor.


abhinav098:
Yeah @patrixr, name sounds nice but it's an old method in api controller that has been used in a lot of places, we might need to change it everywhere. 

We can do it later. I hope that's fine :)

PR #821, opened 2019-09-16
GCW-2801- Error handling for Gogovan Booking Request

abulsayyad123:
### What does this PR do?
Bug: Returns error if GoGo Van service returns some error.
Solution: Return error with error code.

### Linked PR's: https://github.com/crossroads/shared.goodcity/pull/259
Please review .

### DIFF
--- a/app/controllers/api/v1/gogovan_orders_controller.rb
+++ b/app/controllers/api/v1/gogovan_orders_controller.rb
@@ -23,7 +23,11 @@ module Api
       param :offerId, String, desc: "Id of the offer"
       def calculate_price
         order_price = GogovanOrder.place_order(current_user, order_params)
-        render json: order_price.to_json
+        if order_price[:error]
+          render_error(order_price[:error])
+        else
+          render json: order_price.to_json
+        end
       end
 
       def driver_details


app/controllers/api/v1/gogovan_orders_controller.rb
```diff
@@ -23,7 +23,11 @@ class GogovanOrdersController < Api::V1::ApiController
       param :offerId, String, desc: "Id of the offer"
       def calculate_price
         order_price = GogovanOrder.place_order(current_user, order_params)
-        render json: order_price.to_json
+        if order_price[:error]
+          render_error(order_price[:error])
```
namrataukirde: What kind of error does it throw? Is it full_message or with status code? (As per my knowledge it will through error including status code like an example of 503 with service not available,)🤔 

https://github.com/crossroads/go_go_van_api/blob/master/lib/go_go_van_api/order.rb#L38

Looks like it sends with status code so I am not sure we should show that way to a user or not. Let me know your thoughts on this. :)

Could you please add a screenshot of error how it displays on user side?

abulsayyad123:
Closing this PR for now. We cannot show actual server error to the user. Need to consider different scenarios of errors.

PR #827, opened 2019-09-24
No schedule on public holidays

patrixr:
### What does this PR do?

BUG: We suspect some timezone issue allows people to schedule deliveries on public holidays, this adds a backend check to ensure this never happens.

NOTE: I originally went with a model validation, but then realized the controller makes a ggv modification straight away, so I sneaked the verification in the controller to avoid that happening.

### Impacted Areas

Donor app > Transport schedule

### DIFF
--- a/app/controllers/api/v1/deliveries_controller.rb
+++ b/app/controllers/api/v1/deliveries_controller.rb
@@ -98,9 +98,7 @@ module Api
         param :mobile, String
       end
       def confirm_delivery
-        if Holiday.is_holiday(scheduled_date)
-          return render_error(I18n.t('schedule.holiday_conflict', date: scheduled_date));
-        end
+        return unless validate_schedule
 
         @delivery = Delivery.find_by(id: params["delivery"]["id"])
         @delivery.delete_old_associations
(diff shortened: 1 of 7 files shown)


app/controllers/api/v1/deliveries_controller.rb
```diff
@@ -147,6 +149,30 @@ def get_delivery_details
...
+        begin
+          Date.parse(scheduled_at)
+        rescue ArgumentError
+          nil
+        end
+      end
+
+      def validate_schedule
+        if scheduled_date.nil?
+          render_error(I18n.t('schedule.bad_date'));
```
namrataukirde: Semicolon is not required.

app/controllers/api/v1/deliveries_controller.rb
```diff
@@ -147,6 +149,30 @@ def get_delivery_details
...
+          return false
+        end
+
+        if Holiday.is_holiday(scheduled_date)
```
namrataukirde: can we add a question mark at the end of method name `is_holiday?`

namrataukirde:
@patrixr, I feel we should not allow admin's as well to do booking on holiday's but we can ask Matt's thoughts on this.

app/controllers/api/v1/deliveries_controller.rb
```diff
@@ -147,6 +149,30 @@ def get_delivery_details
...
+        return nil unless scheduled_at.present?
+        begin
+          Date.parse(scheduled_at)
+        rescue ArgumentError
+          nil
+        end
+      end
+
+      def validate_schedule
+        if scheduled_date.nil?
```
namrataukirde: Is there any possibility that schedule_date can contain a blank string? 🤔

swatijadhav: `.blank?` will be safer to use.

patrixr: `scheduled_date` is a method that returns `nil?` explicitely

patrixr:
> @patrixr, I feel we should not allow admin's as well to do booking on holiday's but we can ask Matt's thoughts on this.

Cool, I'll move forward with this and discuss later

PR #828, opened 2019-09-24
GCW-2824 fix inventory number remove end-point breaks sometimes

namrataukirde:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2824

### What does this PR do?
It fixes the rollbar issue: https://rollbar.com/crossroads_it/api.goodcity/items/395/

STEPS TO REPRODUCE BUG: 
- Receive item with an inventory number. 
- Click on the receive button. (try creating with inventory number which already exists so it will return 422 from API)
- Now try clicking on cancel button on-screen and nothing happens and it returns above rollbar issue. 

### Impacted Areas
Cancel feature on a page -> Receive with inventory number on admin app.

Please review.

Thanks.

### DIFF
--- a/app/controllers/api/v1/inventory_numbers_controller.rb
+++ b/app/controllers/api/v1/inventory_numbers_controller.rb
@@ -19,7 +19,7 @@ module Api
 
       api :PUT, "/v1/inventory_numbers", "Delete inventory_number"
       def remove_number
-        InventoryNumber.find_by(code: params[:code]).destroy
+        InventoryNumber.find_by(code: params[:code]).try(:destroy)
         render json: {}
       end
     end
--- a/spec/controllers/api/v1/inventory_numbers_controller_spec.rb
+++ b/spec/controllers/api/v1/inventory_numbers_controller_spec.rb
@@ -26,5 +26,13 @@ RSpec.describe Api::V1::InventoryNumbersController, type: :controller do
       expect(response.status).to eq(200)
       expect(parsed_body).to eq( {} )
     end
+
+    it 'returns blank response if inventory_number do not exist in db' do
+      expect {
+        post :remove_number, code: rand(1000..9999)
+      }.to change(InventoryNumber, :count).by(0)
+      expect(response.status).to eq(200)
+      expect(parsed_body).to eq({})
+    end
   end
 end


app/controllers/api/v1/inventory_numbers_controller.rb
```diff
@@ -19,7 +19,9 @@ def create
 
       api :PUT, "/v1/inventory_numbers", "Delete inventory_number"
       def remove_number
-        InventoryNumber.find_by(code: params[:code]).destroy
+        if (inventory_number = InventoryNumber.find_by(code: params[:code]))
+          inventory_number.destroy
+        end
```
patrixr: One liner alternative: `InventoryNumber.find_by(code: params[:code])&.destroy`

Darshan-Chauhan: @patrixr , I was going to comment the same thing. Ruby provides safe navigation operator while rails provides `try` method from ActiveSupport. 

I would like to know which one we should prefer. 
@swatijadhav and @namrataukirde I would love your input here.

I referred: https://stackoverflow.com/questions/45825363/what-is-the-difference-between-try-and-safe-navigation-operator-in-ruby

swatijadhav: As mentioned in the reference, both are valid based on the use case. 
But I would prefer to go with `.try` as it won't raise any unnecessary exception, also throughout the app it would be consistent to use and there won't be confusion.

Darshan-Chauhan: Thanks @swatijadhav  👍

patrixr: I agree with @swatijadhav , but my reasons for preferring `try` is that it's explicit and descriptive as to what it does. Anyone reading it immediately understands that the record might not exist. The `&` kinda goes unnoticed.

patrixr: Also, I have absolutely nothing a good old `if (exists) { destroy }`. Easy to understand, clear, efficient.
--> Less code is not better code

app/controllers/api/v1/inventory_numbers_controller.rb
```diff
@@ -19,7 +19,9 @@ def create
 
       api :PUT, "/v1/inventory_numbers", "Delete inventory_number"
       def remove_number
-        InventoryNumber.find_by(code: params[:code]).destroy
+        if (inventory_number = InventoryNumber.find_by(code: params[:code]))
```
abhinav098: @namrataukirde can we use safe navigation here to find it and destroy it ?
Something like `InventoryNumber.find_by(code: params[:code])&.destroy)`
or `InventoryNumber.find_by(code: params[:code]).try(:destroy)`

app/controllers/api/v1/inventory_numbers_controller.rb
namrataukirde: Thanks, everyone for the input. Nice discussion. 

I missed using try instead of if. In my opinion, I will prefer try over safe navigation for both the same reasons which @swatijadhav mentioned.  

Also, it looks clearer to me. (It's my personal preference over safe navigation 😅)

I will do suggested change and will let you guys know.

PR #830, opened 2019-09-25
#September2019Release2

namrataukirde:
- On Item Search State Filter on applying Image "No Images" gives "Something went wrong" (GCW-2808)
- Add "weight" & "pieces" columns to packages and sync with StockIt (GCW-2780)
- implement new item creation features (GCW-2773) 
- Fix company edit/update with crm_id as nil
- No schedule on public holidays

Please review.


db/schema.rb
```diff
@@ -572,11 +571,14 @@
     t.integer  "favourite_image_id"
     t.date     "stockit_moved_on"
     t.integer  "stockit_moved_by_id"
-    t.boolean  "saleable",                             default: false
+    t.boolean  "saleable",                 default: false
     t.integer  "set_item_id"
     t.string   "case_number"
     t.boolean  "allow_web_publish"
     t.integer  "received_quantity"
+    t.boolean  "last_allow_web_published"
```
patrixr: @namrataukirde do you know what that's about ?

namrataukirde: Correct @patrixr. I don't think it's used anywhere. 

It's not present in current live code: https://github.com/crossroads/api.goodcity/blob/live/db/schema.rb

And no new features or bug fixes uses this field. 

@abhinav098 Could you please help us to understand if it's used anywhere or it's added mistakenly?

abulsayyad123: I am 100% sure it is not used anywhere. I think it's accidentally committed by someone. We can remove this field.

PR #832, opened 2019-10-02
Orders Packages "Actions"

patrixr:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2779

### What does this PR do?

We've long talked about moving a lot of record manipulation logic to the backend, this PR targets the orders packages block on the Stock app. As shown [HERE](https://xd.adobe.com/view/224f7314-f1c0-40b6-7d7a-929e616be6bb-547e/screen/889371e6-fc7b-4556-9a0e-f4649361e6e9/iPhone-6-7-8-122/)

The concept:
1. The app retrieves a list of _"available actions"_
2. The app populates the menu list based on those actions
3. When the button is pressed. The app simply calls `/orders_packages/:id/actions/:action_name` to apply the transformation

### NEW: Feature Testing

I've added [Cucumber](https://github.com/cucumber/cucumber-rails) to the project. It's a "BDD" focused testing library that can work alongside RSPEC.

Feel free to explore and give it a try !

Why it's useful :
- Gherkin files, aka _Feature_ files, can be written in plain English. Very useful to import Matt's requirements (he might actually write some soon which would make testing a breeze for us)
Notice how the mockup's conditions are directly written in the feature file

- It allows testing of large groups of conditions. In this case There's [Order States X OrderPackage States X Quantities Setups] variations, each with different outcomes.

### Impacted Areas

Stock > Items Details Page > Publishing Tab (in development)


app/models/concerns/orders_package_actions.rb
```diff
@@ -0,0 +1,118 @@
...
+    PACKAGE_CANCELLED = -> (model) { model.cancelled? }
+    PACKAGE_DESIGNATED = -> (model) { model.designated? }
+    PACKAGE_DISPATCHED = -> (model) { model.dispatched? }
+
+    def package_has_quantity?
+      @model.package.in_hand_quantity.positive?
+    end
+
+    def can_decrease_qty?
+      @model.quantity > 1
```
swatijadhav: Should it be `@model.quantity > 0` (or simply `@model.quantity`) as even the item has `1` quantity then it can be decreased?

patrixr: @swatijadhav The case where quantity==1 falls under the `cancel` method : )
The menu would have 'Edit Qty` as *disabled* but 'Cancel' would be *enabled*

app/models/concerns/orders_package_actions.rb
```diff
@@ -0,0 +1,118 @@
...
+    end
+
+    def can_increae_qty?
+      package_has_quantity?
```
swatijadhav: we can `alias_method` here

```alias_method : can_increae_qty, : package_has_quantity ```

Comment (patrixr) [reply app/models/concerns/orders_package_actions.rb]: why not : )

File: spec/controllers/api/v1/orders_controller_spec.rb [hunk 3]
```diff
@@ -180,14 +180,14 @@ def create_order_with_transport(state, opts = {})
       end
 
       it "can search orders from a beneficiary's first name" do
-        beneficiary = FactoryBot.create :beneficiary, first_name: 'Frank', last_name: 'Sinatra'
+        beneficiary = FactoryBot.create :beneficiary, first_name: 'Steeeve', last_name: 'Sinatra'
```
steveyken: LOL!

spec/controllers/api/v1/orders_packages_controller_spec.rb
```diff
@@ -64,4 +60,117 @@
       end
     end
   end
+
+  describe 'Executing actions' do
```
steveyken: Great set of specs

app/models/orders_package.rb
```diff
@@ -84,6 +86,11 @@ def update_designation(order_id_to_update)
     update(order_id: order_id_to_update, updated_by: User.current_user)
   end
 
+  def redesignate(new_order_id)
```
steveyken: is it worth wrapping this in a transaction so that the database is rolled back if there is an error partway through? https://api.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html

I'm assuming here that `redesignate` is the main model API entry point for this action.

app/models/concerns/orders_package_actions.rb
```diff
@@ -0,0 +1,116 @@
...
+    def package_has_quantity?
+      @model.package.in_hand_quantity.positive?
+    end
+
+    def can_decrease_qty?
+      @model.quantity > 1
+    end
+
+    def editable_qty?
+      can_decrease_qty? || can_increae_qty?
```
steveyken: typo

```suggestion
      can_decrease_qty? || can_increase_qty?
```

app/models/orders_package.rb
patrixr: oh that's great ! the api certainly lacks atomicity, this is perfect !

app/models/concerns/orders_package_actions.rb
```diff
@@ -0,0 +1,120 @@
...
+  def exec_action(name, opts = {})
+    unless can_exec_action(name)
+      raise ArgumentError.new(I18n.t('orders_package.action_disabled', name: name))
+    end
+
+    action = Actions::ALL_ACTIONS.find { |act| act.name.to_s == name.to_s }
+    action.run(self, opts)
+  end
+
+  private
```
namrataukirde: Is this private keyword is required here?

patrixr: It separates the internal helper methods from the exposed API

PR #837, opened 2019-10-15
GCW-2812 support additional fields

abhinav098:
### What does this PR do?

FEATURE: Adds the ability to add additional fields to the package

DESCRIPTION: StockIt has three extra tables related to items which contain fields that are only relevant to certain types of goods.
We want to duplicate this functionality for GoodCity

### Impacted Areas

Package creation and package details

### Mockup Link

https://xd.adobe.com/view/224f7314-f1c0-40b6-7d7a-929e616be6bb-547e/screen/7d129040-9142-4ed1-ba71-8011197fae58/Create-Item-Computer-form


app/models/electrical.rb
```diff
@@ -0,0 +1,43 @@
...
+  after_update :update_on_stockit
+
+  private
+
+  def set_tested_on
+    self.tested_on = Date.today
+  end
+
+  def downcase_brand
+    self.brand.downcase!
```
swatijadhav: If the `self.brand` value is already downcased, then  `self.brand.downcase!` will return `nil` and `before_save` callback will fail

```suggestion
    self.brand = self.brand.downcase
```

app/models/electrical.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def create_on_stockit
+    response = Stockit::ItemDetailSync.create(self)
+    if response && (errors = response["errors"]).present?
+      errors.each { |key, value| self.errors.add(key, value) }
+    elsif response && (electrical_id = response["electrical_id"]).present?
+      self.update_column(:stockit_id, electrical_id)
```
swatijadhav: **NBD**: Use of extra variables is not required here since we are not calling any method, we are just using hash value.

```
    if response && response["errors"].present?
      response["errors"].each { |key, value| self.errors.add(key, value) }
    elsif response && response["electrical_id"].present?
      self.update_column(:stockit_id, response["electrical_id"])
```

app/models/computer.rb
```diff
@@ -0,0 +1,38 @@
...
+  belongs_to :country, required:false
+  has_one :package, as: :detail, dependent: :destroy
+  after_create :create_on_stockit
+  after_update :update_on_stockit
+  before_save :downcase_brand, if: :brand_changed?
+  before_save :set_updated_by
+
+  private
+
+  def create_on_stockit
```
patrixr: Given that there are a few classes which are very similar, might be worth moving some of this code into a concern. Would make it easier to manage and to add more of those detail models in the future : )

db/package_detail.yml
```diff
@@ -0,0 +1,32 @@
+---
```
patrixr: Think you could add "dummy" in the file name ? to make sure it doesn't get mixed in with seed data :)

or a dummy/ folder works too

lib/classes/package_builder.rb
```diff
@@ -0,0 +1,14 @@
+class PackageBuilder
```
patrixr:
```suggestion
class PackageDetailBuilder
```
Seems more appropriate, since it doesn't return a Package

lib/tasks/sync_subform_to_gc_pkg_types.rake
```diff
@@ -0,0 +1,28 @@
+# rake stockit:sync_subform_to_gc_pkg_types
+
+namespace :stockit do
+  task sync_subform_to_gc_pkg_types: :environment do
+    log = Goodcity::RakeLogger.new("sync_subform_to_gc_pkg_types")
+    codes_json = Stockit::CodeSync.index # fetch stockit codes
+    return if codes_json.blank? || codes_json.nil?
+    stockit_codes = JSON.parse(codes_json["codes"])
+    bar = RakeProgressbar.new(stockit_codes.size)
```
patrixr: `RakeProgressbar`  😱 ❤️

lib/classes/package_builder.rb
abhinav098: yeah, right, but we will be moving the package record building process to this class in future, so name it `PackageBuilder`. Will change that.

db/package_detail.yml
abhinav098: sure @patrixr

app/models/computer.rb
abhinav098: Yes I was also thinking of something like that earlier, creating a concern that would cater all three types. Also in future if we want to add any similar class, we can just include that concern and perform same operations on the record.

app/models/electrical.rb
abhinav098: yeah right @swatijadhav , as @patrixr suggested, i'll create a concern and move the there and will remove the unwanted variables.

app/models/electrical.rb
abhinav098: oh, sure, I'll make that change

app/services/stockit/item_detail_sync.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    def update
+      url = url_for("/api/v1/#{detail_type.pluralize}/update")
+      put(url, detail_params)
+    end
+
+    protected
+
+    def detail_params
+      { "#{detail_type}": detail.attributes.except("id", "stockit_id" "created_at", "updated_at", "updated_by_id") }
```
namrataukirde: a comma is missing between `stcokit_id` and `created_at`.

abhinav098: Good catch 👍 Thanks @namrataukirde . :)

patrixr: Why does it not generate an error ?
I miss statically compiled languages 😅

db/package_detail.yml
patrixr: Thanks :)

lib/classes/package_builder.rb
patrixr: Maybe it can be `PackageDetailBuilder` with just a `build()` method.
And later when you create `PackageBuilder`, it can just call `PackageDetailBuilder` : )

app/controllers/api/v1/computer_accessories_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      api :PUT, "/v1/computers", "Create or Update a computer"
+      param_group :computer_accessory
+      def update
+        @computer_accessory.assign_attributes(computer_accessory_params)
+        if @computer_accessory.valid? and @computer_accessory.save
+          render json: @computer_accessory, serializer: serializer
+        else
+          render_error(@computer_accessory.errors.full_messages.join(', '))
+        end
```
swatijadhav: Can we create something similar for `update` action as we have for `create` action?
https://github.com/crossroads/api.goodcity/blob/master/app/controllers/api/v1/api_controller.rb#L23

```
    def update_and_render_object_with_errors(object)
      if object.valid? and object.save
        render json: object, serializer: serializer_for(object)
      else
        render_error(object.errors.full_messages.join('. '))
      end
    end
```

app/controllers/api/v1/packages_controller.rb
```diff
@@ -270,16 +270,36 @@ def remove_stockit_prefix(stockit_inventory_number)
...
+
+      def electrical_attributes
+        [:brand, :model, :serial_number, :country_id, :standard,
+          :voltage, :frequency, :power, :system_or_region, :test_status,
+          :tested_on, :updated_by_id]
+      end
+
+      def computer_accessory_attributes
+        [:brand, :model, :serial_num, :country_id, :size,
+          :interface, :comp_voltage, :comp_test_status, :updated_by_id]
```
swatijadhav: **NBD**: It would be good if we can arrange these attributes in alphabetical order, so that it would be good to track in future

app/serializers/api/v1/electrical_serializer.rb
```diff
@@ -0,0 +1,15 @@
+module Api::V1
+  class ElectricalSerializer < ApplicationSerializer
+    embed :ids, include: true
+
+    attributes :id, :brand, :model, :serial_number, :country_id, :standard,
+    :voltage, :frequency, :power, :system_or_region, :test_status,
+    :tested_on, :updated_by_id
```
swatijadhav: Fix indentation

spec/controllers/api/v1/computer_accessories_controller_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    serialized_record_with_country = Api::V1::ComputerAccessorySerializer.new(@record, include_country: true).as_json
+    @parsed_body_with_country = JSON.parse( serialized_record_with_country.to_json )
+
+    serialized_record_without_country = Api::V1::ComputerAccessorySerializer.new(@record, include_country: false).as_json
+    @parsed_body_without_country = JSON.parse(serialized_record_without_country.to_json)
+  end
+
+  describe "GET computer_accessories" do
+    it "returns 200", :show_in_doc do
```
swatijadhav: Please add updated `apipie_examples.json` as here `: show_in_doc` is used.
https://github.com/crossroads/api.goodcity/blob/master/doc/apipie_examples.json

app/controllers/api/v1/packages_controller.rb
```diff
@@ -388,6 +409,10 @@ def existing_package
         end
       end
 
+      def assign_detail
```
steveyken: if you can get accepts_attributes_for working correctly, then you can remove this method.

app/models/concerns/subform_callbacks.rb
```diff
@@ -0,0 +1,36 @@
...
+    response = Stockit::ItemDetailSync.create(self)
+    if response && (errors = response["errors"]).present?
+      errors.each { |key, value| self.errors.add(key, value) }
+    elsif response && (stockit_id = response["#{self.class.name.underscore}_id"]).present?
+      update_column(:stockit_id, stockit_id)
+    end
+  end
+
+  def update_on_stockit
+    Stockit::ItemDetailSync.update(self)
```
steveyken: Update should also to assign back the stockit_id because in the case where you call Update and the record doesn't exist in Stockit then stockit will create the Item and return the stockit_id to you.

so I think update and create can be the same method

db/migrate/20190925063509_add_detail_to_packages.rb
```diff
@@ -0,0 +1,5 @@
+class AddDetailToPackages < ActiveRecord::Migration
+  def change
+    add_reference :packages, :detail, polymorphic: true, index: true
```
steveyken: often good to add two indexes for composite keys.

1: index on [detail_type, detail_id]
2: index on [detail_id, detail_type]

lib/classes/package_detail_builder.rb
```diff
@@ -0,0 +1,14 @@
+class PackageDetailBuilder
```
steveyken: Whilst it's great to see a builder class, I'd like to see if possible to remove this by getting accepts_nested_attributes_for working

app/models/ability.rb
```diff
@@ -98,6 +101,18 @@ def beneficiary_abilities
     end
   end
 
+  def computer_abilities
```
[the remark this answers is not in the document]
abhinav098: yes @steveyken , I had that in mind, actually I wanted to do the complete flow once and then add the permissions. I have now added the permissions for the same. viz. `can_manage_package_detail` for Supervisor and Order Fulfillment user. It works the same way as `can_manage_package?` permission works.

app/controllers/api/v1/packages_controller.rb
```diff
@@ -322,6 +342,7 @@ def package_record
         else
           @package.assign_attributes(package_params)
         end
+        @package.detail = assign_detail if params["package"]["detail_type"].present?
```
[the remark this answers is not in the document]
abhinav098: yeah @steveyken I was trying to get it work without this method but it seems that `accept_nested_attributes_for` works for has_one/has_many relationship. `detail` is in `belongs_to` association with package, because detail is polymorphic.

one workaround for this can be -> https://stackoverflow.com/a/3975758
other option is the way I have done it in the controller itself.

cc- @namrataukirde

app/models/electrical.rb
abhinav098: yes @steveyken this is covered in the spec.

app/controllers/api/v1/computer_accessories_controller.rb
abhinav098: yes sure we can @swatijadhav

app/controllers/api/v1/packages_controller.rb
abhinav098: sure @swatijadhav , i'll do it now.

app/serializers/api/v1/electrical_serializer.rb
abhinav098: sure

app/models/concerns/subform_callbacks.rb
abhinav098: sure, it's done

db/migrate/20190925063509_add_detail_to_packages.rb
abhinav098: @steveyken , running the migration has already created index.

`add_index "packages", ["detail_type", "detail_id"], name: "index_packages_on_detail_type_and_detail_id", using: :btree`

do you want me to do it seperately ?

app/controllers/api/v1/api_controller.rb
```diff
@@ -20,72 +20,80 @@ def save_and_render_object(object)
...
-    end
+      def bool_param(key, default)
+        return default if params[key].nil?
+        params[key].to_s == "true"
+      end
 
-    private
+      def time_epoch_param(key)
+        timestamp = params.fetch(key, nil)
+        return timestamp ? Time.at(Integer(timestamp) / 1000).in_time_zone : nil
```
swatijadhav: remove `return`
```suggestion
        timestamp ? Time.at(Integer(timestamp) / 1000).in_time_zone : nil
```

app/controllers/api/v1/computers_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+        render json: @computers, each_serializer: serializer
+      end
+
+      def show
+        render json: @computer, serializer: serializer, include_country: true
+      end
+
+      api :PUT, "/v1/computers", "Create or Update a computer"
+      param_group :computer
+
```
swatijadhav: **nit**: remove this extra linebreak

app/controllers/api/v1/packages_controller.rb
```diff
@@ -35,42 +35,45 @@ class PackagesController < Api::V1::ApiController
...
-          include_orders_packages: true,
-          exclude_stockit_set_item: @package.set_item_id.blank? ? true : false,
-          include_images: @package.set_item_id.blank?,
-          include_allowed_actions: true
-        ).as_json
+                                          serializer: stock_serializer,
+                                          root: "item",
+                                          include_order: true,
+                                          include_orders_packages: true,
+                                          exclude_stockit_set_item: @package.set_item_id.blank? ? true : false,
```
swatijadhav:
```suggestion
                                          exclude_stockit_set_item: @package.set_item_id.blank?,
```

app/models/computer_accessory.rb
```diff
@@ -0,0 +1,19 @@
...
+  before_save :save_correct_country, if: :request_from_stockit?
+  before_save :set_updated_by
+  before_save :map_drop_down_attributes, if: :request_from_stockit?
+  after_commit :create_on_stockit, on: :create
+  after_update :update_on_stockit
+
+  private
+
+  def map_drop_down_attributes
+    self.comp_test_status = Lookup.find_by(name: "comp_test_status", value: comp_test_status).label_en if comp_test_status_changed?
```
swatijadhav: Move this if condition `if comp_test_status_changed?` above where callback is called.

```ruby
before_save :map_drop_down_attributes, if: :request_from_stockit? && :comp_test_status_changed?
```

app/models/concerns/subform_utilities.rb
```diff
@@ -0,0 +1,55 @@
+module SubformUtilities
+  extend ActiveSupport::Concern
+
+  included do
+    extend(ClassMethods)
+  end
+
+  module ClassMethods
+    def distinct_by_column(name)
+      select("distinct on (#{self.name.underscore.pluralize}.#{name}) #{self.name.underscore.pluralize}.*")
```
steveyken: this is an SQL injection issue because `name` is unparsed user input.
best option I think is to whitelist the incoming `name` param and ensure it belongs to a certain list of acceptable fields.
`(x, y z).include?(name)`

db/dummy_package_detail.yml
```diff
@@ -0,0 +1,32 @@
+---
```
steveyken: can you follow the convention in the folder to have 1 file per model?

computers.yml
electricals.yml
computer_accessories.yml etc etc

makes it easier to add more in the future

db/lookup.yml
```diff
@@ -0,0 +1,96 @@
+---
```
steveyken: same name change to `lookups.yml`

app/controllers/api/v1/electricals_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+          param :frequency, String, desc: "ram of the record"
+          param :power, String, desc: "hdd of the record"
+          param :system_or_region, String, desc: "optical of the record"
+          param :test_status, String, desc: "video of the record"
+          param :tested_on, Date, desc: "sound of the record"
+          param :updated_by_id, Integer, desc: "id of the user who updated the record"
+        end
+      end
+
+      def index
```
steveyken: were you going to create an `autocomplete` function? I can't remember what we decided

app/models/concerns/subform_utilities.rb
abhinav098: @steveyken , here the records are getting filtered by column name. Only distinct records will be sent in response for brand. It is not user input. While requesting from client, it is asking for distinct by "brand" records.

``` let details = await this.store.query(selectedSubform, {
              distinct: "brand"
            });```

That's why whitelisting it not required. 

Please let me know your thoughts on this.

File: app/controllers/api/v1/electricals_controller.rb (hunk 25 above)
Comment (abhinav098) [reply app/controllers/api/v1/electricals_controller.rb]: No, I have kept it based on parameters. If we have `params[:distinct]` present, then only we will filter out the data. Otherwise will send all the electricals. 

There was a reason to not implement separate endpoint because, for some of the subforms(eg computer) there are lot of autocomplete fields and sending request everytime user types in the searchText would have reduced the overall performance. So to avoid that, we implemented it this way.

File: db/dummy_package_detail.yml (hunk 23 above)
Comment (abhinav098) [reply db/dummy_package_detail.yml]: sure, I'll do that @steveyken .

File: db/lookup.yml (hunk 24 above)
Comment (abhinav098) [reply db/lookup.yml]: sure @steveyken .

File: app/controllers/api/v1/computer_accessories_controller.rb [hunk 26]
```diff
@@ -0,0 +1,58 @@
...
+          param :size, String, desc: "size of the record"
+          param :interface, String, desc: "interface of the record"
+          param :comp_voltage, String, desc: "comp_voltage of the record"
+          param :comp_test_status, String, desc: "comp_test_status of the record"
+          param :updated_by_id, Integer, desc: "updated_by_id of the record"
+        end
+      end
+
+      def index
+        @computer_accessories = @computer_accessories.distinct_by_column(params["distinct"]) if params["distinct"]
```
namrataukirde: Good to see this method `distinct_by_column` :+1:

app/models/concerns/subform_utilities.rb
steveyken: hi @abhinav098 by 'user input' I mean that the value is coming directly from the params hash which is being used to set name here. It's not safe just because you are setting 'brand' in the end user app. Consider the case where a hacker sends API requests direct to the system bypassing the app. They can put whatever they like in the `params[:distinct]` field. You could even change this in the Chrome debugger if you wanted.

Use `ActiveRecord.santize_sql` and lookup the column name to be safe. Here's an untested example to help get started:

```
def distinct_by_column(column)
  raise StandardError('Column name invalid') unless self.column_names.include?(col)
  tbl_name = self.name.tableize # self.name.underscore.pluralize
  sql_select = self.send(:sanitize_sql, ['distinct on (?.? ?.*)', tbl_name, column, tbl_name])
  select(sql_select)
end
```

Thanks!

steveyken: Let me know if I haven't explained it well enough - it's important we're on the same page on this one!

PR #840, opened 2019-10-21
PackagesInventory Migration

patrixr:
### Ticket Link: 

NOTE: Specify ticket link

### What does this PR do?

FEATURE: Specify feature name

BUG: Specify bug 

NOTE: Give a description of what this PR does. If it fixes any bug, specify cause of an issue and approach/solution added to fix that issue.

### Impacted Areas

### Screenshots

### Mockup Link

Note: Specify mockup link

### Linked PR's:


lib/classes/packages_inventories_importer.rb
```diff
@@ -0,0 +1,203 @@
...
+
+  # --- Sanity checks
+  def verify_package(package)
+    @errors.push(MULTIPLE_LOCATIONS_ERR % [package.id]) if package.locations.length > 1
+    @errors.push(NO_LOCATION_ERR % [package.id]) if package.locations.length.zero?
+    @errors.push(INVALID_QUANTITY % [package.id, package.quantity]) if package.quantity.negative?
+    if is_dispatched?(package)
+      @errors.push(MISSING_ORDERS_PACKAGE % [package.id]) if package.orders_packages.count.zero?
+    end
+  end
```
patrixr: cc @namrataukirde

spec/factories/packages.rb
```diff
@@ -27,11 +27,31 @@
 
     trait :package_with_locations do
       after(:create) do |package|
-        package_location = create :packages_location, package_id: package.id, quantity: package.received_quantity
-        package.location_id = package_location.location_id
-        package.packages_locations ||= []
-        package.packages_locations << package_location
-        package.save
+        build(:packages_location, package_id: package.id, quantity: package.received_quantity).sneaky do |package_location|
```
steveyken: sneaking sneaky into specs now I see :+1:  :100:  :smile:

spec/models/concerns/inventory_legacy_support_spec.rb
```diff
@@ -0,0 +1,399 @@
...
+  let(:other_location) { create(:location) }
+  let(:user) { create(:user) }
+  let(:package) { create(:package) }
+  let(:other_package) { create(:package) }
+  let(:packages_location) { create(:packages_location, quantity: quantity, location: location, package: package) }
+  let(:computer) { PackagesInventory::Computer }
+
+  before { User.current_user = user }
+
+  after do
```
steveyken: Just to make things really clear
```suggestion
  after(:each) do
```

spec/models/packages_inventory_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+  describe 'Associations' do
+    it { is_expected.to belong_to :package }
+    it { is_expected.to belong_to :location }
+    it { is_expected.to belong_to :source }
+  end
+
+  describe 'Validations' do
+    it 'prevents invalid actions' do
+      expect {
+        create(:packages_inventory, action: 'love')
```
steveyken: :heart:

spec/support/touch.rb
```diff
@@ -0,0 +1,3 @@
+module Touch
+  def touch(*toucheables); end
+end
```
steveyken: The antonym for https://en.wikipedia.org/wiki/The_Intouchables :smiley:

db/schema.rb
```diff
@@ -601,6 +604,29 @@
...
+  add_index "packages_inventories", ["location_id"], name: "index_packages_inventories_on_location_id", using: :btree
+  add_index "packages_inventories", ["package_id", "source_type"], name: "index_packages_inventories_on_package_id_and_source_type", using: :btree
+  add_index "packages_inventories", ["package_id"], name: "index_packages_inventories_on_package_id", using: :btree
+  add_index "packages_inventories", ["source_id", "source_type"], name: "index_packages_inventories_on_source_id_and_source_type", using: :btree
+  add_index "packages_inventories", ["source_id"], name: "index_packages_inventories_on_source_id", using: :btree
+  add_index "packages_inventories", ["source_type", "package_id"], name: "index_packages_inventories_on_source_type_and_package_id", using: :btree
+  add_index "packages_inventories", ["source_type", "source_id"], name: "index_packages_inventories_on_source_type_and_source_id", using: :btree
+  add_index "packages_inventories", ["source_type"], name: "index_packages_inventories_on_source_type", using: :btree
+  add_index "packages_inventories", ["user_id"], name: "index_packages_inventories_on_user_id", using: :btree
+
```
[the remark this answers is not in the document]
patrixr: Good point

spec/factories/packages.rb
patrixr: Didn't want a factory to secretly trigger builds of other tables 😄

spec/support/touch.rb
patrixr: Amazing movie btw : )

lib/classes/packages_inventories_importer.rb
```diff
@@ -0,0 +1,203 @@
...
+  def failed?
+    @errors.length.positive?
+  end
+
+  # --- Loops over each package in the database
+  def each_package
+    count = Package.count
+    bar = RakeProgressbar.new(count) unless Rails.env.test?
+    begin
+      Package.find_each do |package|
```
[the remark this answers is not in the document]
namrataukirde: 👍

lib/classes/whereable.rb
```diff
@@ -0,0 +1,16 @@
+module Whereable
+  attr_accessor :relation
+
+  def use_model(mod)
+    @model = mod
+    @relation = mod.all
```
namrataukirde: I think using `.all` is risky if table has a large number of rows. `.all` instructs Active Record to fetch the entire table in a single pass and it stores all records in memory which is not required here as per my understanding.

Method used here : https://github.com/crossroads/api.goodcity/pull/840/files#diff-ed6a25795ba8774bf0e253b8bfab146aR34

Which call https://github.com/crossroads/api.goodcity/pull/840/files#diff-ed6a25795ba8774bf0e253b8bfab146aR79

It looks like we are trying something similar to 
`PackagesInventory.all.where("#{model.class.name.underscore}_id = (?)", Utils.to_id(model))`

which can be done as `PackagesInventory.where("#{model.class.name.underscore}_id = (?)", Utils.to_id(model))`

correct me If I am wrong. :)

patrixr: All is just a base relation which gets extended. If it does indeed get computed instantly then I'll just replace it with `where(nil)`. Will double check

app/models/packages_inventory.rb
```diff
@@ -0,0 +1,80 @@
...
+    errors.add(:errors, I18n.t('package_inventory.bad_action', action: action) )
+  end
+
+  def validate_location
+    errors.add(:errors, I18n.t('package_inventory.invalid_dispatch_location')) if location.dispatch?
+  end
+
+  def validate_quantity
+    return errors.add(:errors, I18n.t('package_inventory.quantities.zero_invalid')) if quantity.zero?
+    if incremental?
```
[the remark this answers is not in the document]
patrixr: sure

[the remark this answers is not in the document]
patrixr: actually that doesn't work. because the case where `incremental=true && quantity=positive` would not land in the second condition when it should not

It can only be
```
if incremental? && quantity.negative?
  ...
elsif  decremental? && quantity.positive?
  ...
end
```

Which kinda adds to the same thing apart from moving a condition one line up

PR #843, opened 2019-10-31
GCW-2812 add lookups for fixed values of some dropdowns

abhinav098:
### What does this PR do?

Adds Lookups table for fixed attributes. eg, electrical frequency, electrical voltage, test_status etc.

LINKED TO PARENT BRANCH, 
i.e. `GCW-2812-support-additional-fields`

### DIFF
--- a/app/models/computer.rb
+++ b/app/models/computer.rb
@@ -4,9 +4,16 @@ class Computer < ActiveRecord::Base
 
   belongs_to :country, required: false
   has_one :package, as: :detail, dependent: :destroy
-  after_commit :create_on_stockit, on: :create
-  after_update :update_on_stockit
   before_save :save_correct_country, if: :request_from_stockit?
   before_save :downcase_brand, if: :brand_changed?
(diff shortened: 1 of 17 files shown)


app/models/computer.rb
```diff
@@ -4,9 +4,16 @@ class Computer < ActiveRecord::Base
...
   before_save :save_correct_country, if: :request_from_stockit?
   before_save :downcase_brand, if: :brand_changed?
   before_save :set_updated_by
+  before_save :map_drop_down_attributes, if: :request_from_stockit?
+  after_commit :create_on_stockit, on: :create
+  after_update :update_on_stockit
+
+  private
+
+  def map_drop_down_attributes
```
steveyken: @abhinav098 there should be no need to map the attributes. use the id of the lookup object. So 

`comp_test_status_id = lookup_id`

I may have confused you with the key/label stuff.
We just need standard rails `reference_id` tooling here.

PR #845, opened 2019-11-06
GCW 2812 lookup id setting

abhinav098:
Related to Parent PR: https://github.com/crossroads/api.goodcity/pull/837
This PR adds lookup-id usage while creation for "test_status, voltage, frequency, comp_test_status."


lib/classes/package_detail_builder.rb
```diff
@@ -1,30 +1,34 @@
...
     klass = detail_type.classify.safe_constantize
-    params = lookup_ids_hash.merge(detail_params_without_id)
+    params = detail_params_without_some_params.merge(lookup_ids_hash)
     klass&.new(params)
   end
 
   private
 
-  def detail_params_without_id
+  def detail_params_without_some_params
```
steveyken: This is a strange name! I think it's fine to just call this method `detail_params` and make a comment inside the method that we are excluding various fields and why we are doing that

lib/classes/package_detail_builder.rb
```diff
@@ -1,30 +1,34 @@
...
+    return {} unless detail_params
     detail_params.except("id", "comp_test_status", "test_status", "frequency", "voltage")
```
steveyken: Suggestion to make this a one line method
```
(detail_params || {} }.except("id", "comp_test_status", "test_status", "frequency", "voltage")
```

lib/classes/package_detail_builder.rb
```diff
@@ -1,30 +1,34 @@
...
   end
 
   # mapping stockit request and setting id for
   def lookup_ids_hash
+    return {} unless @request_from_stockit
```
steveyken: Can you add a code comment for why we need this line? It's not clear to me.

abhinav098: sure, @steveyken , added.

app/services/stockit/item_detail_sync.rb
```diff
@@ -34,7 +34,17 @@ def update
     private
 
     def detail_params
-      params = detail.attributes.except("id", "created_at", "updated_at", "stockit_id", "updated_by_id")
+      params = detail.attributes
+      if (detail_type == "computer") || (detail_type == "computer_accessories")
```
namrataukirde: ['computer',  "computer_accessories"].include?(detail_type)

How about this? 

Just a thought.

lib/classes/package_detail_builder.rb
```diff
@@ -1,14 +1,38 @@
...
+    # same and assign it to params hash for mapping stockit values to lookup ids on GC
+    detail_params["stockit_id"] = detail_params["id"] if detail_params && detail_params["id"]
+    params = detail_params&.except("id", "comp_test_status", "test_status", "frequency", "voltage") || {}
+    # map stockit values to lookup ids if request from stockit
+    lookup_hash = map_lookup_id if @request_from_stockit
+    lookup_hash ? params.merge(lookup_hash) : params
+  end
+
+  def map_lookup_id
+    ["comp_test_status", "test_status", "frequency", "voltage"].each_with_object({}) do |item, hash|
```
namrataukirde: Can we move it to the constant in the same class?

PR #846, opened 2019-11-07
Cache clear: packages_categories

patrixr:
Touch on `package_categories_package_type` to clear the cache when the relationship to package_types changes


abhinav098:
@patrixr could you please explain what will happen with this change ? I mean how this fixes the issue ?

patrixr:
> @patrixr could you please explain what will happen with this change ? I mean how this fixes the issue ?

We manually cleared the cache. This is just for future-proofing. So that the next time we modify the relationship it invalidates the cache

PR #849, opened 2019-11-13
GCW-2890 rake for importing data for package detail from stockit

abhinav098:
### What does this PR do?

This PR adds a rake task to manage previous data for packages. On stockit,
1. Fetch packages(items) from stockit
2. Check for details, if present, create a record for those details on GC
3. If not present, create a blank record on GC which will get synced to Stockit.

### Impacted Areas
Package detail data.

Stockit PR link https://bitbucket.org/crfdevs/stockit/pull-requests/72

### DIFF
--- a/lib/goodcity/detail_factory.rb
+++ b/lib/goodcity/detail_factory.rb
@@ -1,6 +1,6 @@
 module Goodcity
   class DetailFactory
-    PERMITTED_DETAIL_TYPES = %w[Computer Electrical ComputerAccessory].freeze
+    PERMITTED_DETAIL_TYPES = %w[computer electrical computer_accessory].freeze
     FIXED_DETAIL_ATTRIBUTES = %w[comp_test_status test_status frequency voltage].freeze
 
     attr_accessor :item, :package, :stockit_detail_id
@@ -12,7 +12,7 @@ module Goodcity
     end
 
     def run
-      package && import_and_save_detail? && package_updated?
+      import_and_save_detail? && !detail_present? && package_updated?
     end
 
     private
@@ -25,12 +25,16 @@ module Goodcity
       item["id"] && item["detail_type"] && stockit_detail_id
     end
 
+    def detail_present?
+      package.detail.present?
+    end
+
     def detail_type
       item["detail_type"] || package&.package_type&.subform&.titleize
     end
 
     def detail_id
-      return unless PERMITTED_DETAIL_TYPES.include?(item["detail_type"])
+      return unless PERMITTED_DETAIL_TYPES.include?(item["detail_type"].underscore)
       create_detail_record&.id
     end
 
--- a/lib/goodcity/detail_factory.rb
+++ b/lib/goodcity/detail_factory.rb
@@ -2,27 +2,37 @@ module Goodcity
   class DetailFactory
     PERMITTED_DETAIL_TYPES = %w[computer electrical computer_accessory].freeze
     FIXED_DETAIL_ATTRIBUTES = %w[comp_test_status test_status frequency voltage].freeze
+    PACKAGE_DETAIL_ATTRIBUTES = {
+      computer_accessory: %w[brand comp_voltage country_id interface
+        model serial_num size].freeze,
+      computer: %w[brand comp_voltage country_id cpu hdd
+        lan mar_ms_office_serial_num mar_os_serial_num model
+        ms_office_serial_num optical os os_serial_num ram serial_num
+        size sound usb video wireless].freeze,
+      electrical: %w[brand country_id model power serial_number standard
+        system_or_region].freeze
+    }
 
-    attr_accessor :item, :package, :stockit_detail_id
+    attr_accessor :stockit_item_hash, :package, :stockit_detail_id
 
-    def initialize(item, package)
-      @item = item
+    def initialize(stockit_item_hash, package)
+      @stockit_item_hash = stockit_item_hash
       @package = package
-      @stockit_detail_id = item["detail_id"]
+      @stockit_detail_id = stockit_item_hash["detail_id"]
     end
 
     def run
-      import_and_save_detail? && !detail_present? && package_updated?
+      create_detail_and_update_package
     end
 
     private
 
-    def package_updated?
-      package.update_columns(detail_id: detail_id, detail_type: detail_type)
+    def create_detail_and_update_package
+      package && package.update_columns(detail_id: create_detail, detail_type: detail_type)
     end
 
-    def import_and_save_detail?
-      item["id"] && item["detail_type"] && stockit_detail_id
+    def detail_present_on_stockit?
+      stockit_item_hash["id"] && stockit_item_hash["detail_type"] && stockit_detail_id
     end
 
     def detail_present?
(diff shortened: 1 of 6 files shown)


lib/tasks/stockit_item_detail_import.rake
```diff
@@ -0,0 +1,120 @@
...
+      )
+    end
+  end
+
+  def import_and_save_detail?
+    item["id"] && item["detail_type"] && item["detail_id"]
+  end
+
+  def create_new_record?
+    item["detail_id"].nil? || item["detail_id"].blank?
```
steveyken:
```suggestion
    item["detail_id"].blank?
```
`nil` is covered by blank?

lib/tasks/stockit_item_detail_import.rake
```diff
@@ -0,0 +1,120 @@
...
+  end
+
+  def detail_type
+    item["detail_type"] || package.package_type&.subform
+  end
+
+  def detail_id
+    return unless PERMITTED_DETAIL_TYPES.include?(item["detail_type"])
+
+    create_new_record? ? create_empty_record&.id : create_detail_with_attributes&.id
```
steveyken: In a sense, `create_empty_record` is really just `create_detail_with_attributes` where the attributes are empty

lib/tasks/stockit_item_detail_import.rake
```diff
@@ -0,0 +1,120 @@
...
+      Computer.create(computer_attributes)
+    when "electrical"
+      Electrical.create(electrical_attributes)
+    when "computer_accessory"
+      ComputerAccessory.create(computer_accessory_attributes)
+    end
+  end
+
+  def create_empty_record
+    detail_type.classify&.constantize.create unless detail_type.blank?
```
steveyken: Having `unless detail_type.blank?` feels like overly defensive programming here. In the case where `detail_type` is actually blank then we silently continue and don't create a record even though we assume it was done. Better to remove the `unless ...` block and let the method fail so that we expose any potential issues. "Fail early and fast" is a good principle here.

steveyken: That said, I do appreciate this is a 'single use' task and not going to be needed again, but it would be important to confirm that it works correctly as this is data that is being migrated.

abhinav098:
Sure @steveyken , thanks for the early feedback. I'll incorporate your suggestions and work accordingly 👍

spec/lib/goodcity/detail_factory_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+describe Goodcity::DetailFactory do
+  let(:item_params) {  }
+  let(:pacakge) {}
+  let(:detail_factory) { described_class.new(item_params, package) }
+
+  before do
+    stub_request(:get, /goodcitystorage.blob.core.windows.net/).
+      with(headers: { "Accept" => "*/*", "User-Agent" => "Ruby" }).
+      to_return(status: 200, body: file, headers: {}).response.body
+  end
```
steveyken: @abhinav098 could you add a few tests that show the various detail being added. Perhaps one approach would be to stub the method `Stockit::ItemSync.index_with_detail` to return a hash of detail params (rather than stubbing a web request) and confirm that they are applied to the package in GoodCity. The tests can confirm that each parameter is successfully applied.

abhinav098: Sure, @steveyken , I'll work on this, once i'm done with the box and pallet addition/removal of items thingy.

lib/goodcity/detail_factory.rb
```diff
@@ -0,0 +1,94 @@
...
+    def package_updated?
+      package.update_columns(detail_id: detail_id, detail_type: detail_type)
+    end
+
+    def import_and_save_detail?
+      item["id"] && item["detail_type"] && stockit_detail_id
+    end
+
+    def detail_present?
+      package.detail.present?
```
swatijadhav: This is used only at one place [above](https://github.com/crossroads/api.goodcity/pull/849/files#diff-4595f7b4ee82ab803bd631af46579420R15) and also checking the absence of `package.details`. Better to remove it and directly use as `package.detail.blank?`

lib/goodcity/detail_factory.rb
```diff
@@ -0,0 +1,94 @@
...
+
+    def detail_id
+      return unless PERMITTED_DETAIL_TYPES.include?(item["detail_type"].underscore)
+      create_detail_record&.id
+    end
+
+    def create_detail_record
+      GoodcitySync.request_from_stockit = true
+      detail_type.classify.constantize.where(stockit_id: stockit_detail_id)
+                 .first_or_create(send("#{detail_type.underscore}_attributes".to_sym))
```
swatijadhav: Can we simplify this as:

```ruby
detail_type_class = detail_type.classify.constantize
package_attributes = send("#{detail_type.underscore}_attributes".to_sym)
detail_type_class.where(stockit_id: stockit_detail_id).first_or_create(package_attributes)
```

lib/goodcity/detail_factory.rb
```diff
@@ -0,0 +1,94 @@
...
+      attr_hash = {}
+      %w[
+        brand comp_voltage country_id interface
+        model serial_num size
+      ].each do |attr|
+        attr_hash.merge({ "#{attr}": item[attr.to_s] })
+      end
+      attr_hash["stockit_id"] = stockit_detail_id
+      attr_hash.merge(lookup_hash)
+    end
```
swatijadhav: Above can be simplified as 
```
PACKAGE_ATTRIBUTES = {
  computer_accessory: %w[ brand comp_voltage country_id interface model serial_num size],
  electrical: %w[brand country_id model power serial_number standard system_or_region],
  computer: %w[
        brand comp_voltage country_id cpu hdd
        lan mar_ms_office_serial_num mar_os_serial_num model
        ms_office_serial_num optical os os_serial_num ram serial_num
        size sound usb video wireless]
}

def create_detail_record
    detail_type_class = detail_type.classify.constantize
    detail_type_class.where(stockit_id: stockit_detail_id).first_or_create(
        package_type_attributes(PACKAGE_ATTRIBUTES["#{detail_type.underscore}".to_sym])
    )
end

def package_type_attributes(attributes)
      attr_hash = {}
      attributes.each do |attr|
        attr_hash.merge({ "#{attr}": item[attr.to_s] })
      end
      attr_hash["stockit_id"] = stockit_detail_id
      attr_hash.merge(lookup_hash)
    end
end
```

lib/tasks/stockit_item_detail_import.rake
```diff
@@ -0,0 +1,36 @@
...
+          package = Package.find_by(stockit_id: item["package_id"]) if item_id
+          begin
+            is_package_updated = Goodcity::DetailFactory.new(item, package).run
+          rescue StandardError => exception
+            log.error("package: #{package.id}, stockit_item: #{item_id} #{exception.message}")
+          end
+          if is_package_updated
+            print "."
+            count += 1
+          end
```
swatijadhav: This whole block should be inside `if item_id.. end` ? will this need to be executed if `item_id` is blank?

abhinav098:
@swatijadhav , thanks for the feedback and suggestions, I'll pick up this once i'm done with box pallet.

abhinav098:
@steveyken , @swatijadhav this PR is complete from my side. Please re-review this once.
I have added test cases and also made some amendments for the scenarios I encountered while writing test cases.

Thanks :)

PR #854, opened 2019-11-19
Remove old move endpoints

patrixr:
### Ticket Link: 

This relates to a comment from Tim on GCW-2778

### What does this PR do?

There were some move bugs. Given that `LocationOperations` had already been implemented (but were mainly unused), we decided to remove entirely the old endpoints and to write it as a "post-stockit" world would.

The `/move` endpoint now takes `from/to/quantity` parameters.

Another Stock PR is coming soon which uses this endpoint and locks the quantity to the entire amount on the front-end. Making it easier to allow partial quantity in the future.

Note: Moving sets is not supported (Matt confirmed). I removed old code that implied this was possible.


app/models/ability.rb
```diff
@@ -179,7 +179,7 @@ def image_abilities
 
   def item_abilities
     if can_manage_items?
-      can [:index, :show, :create, :update, :messages, :move_stockit_item_set, :move_set_partial_qty,
+      can [:index, :show, :create, :update, :messages, :move_set_partial_qty,
```
steveyken: can we get rid of `move_set_partial_qty` as well?

patrixr: oh, forgot to remove it from the list. Good catch

app/models/concerns/operations/order_fulfilment_operations.rb
```diff
@@ -74,24 +74,8 @@ def order_unprocessed?(order)
...
-      def initialize
-        super(I18n.t('operations.dispatch.unprocessed_order'))
-      end
-    end
-
-    class AlredyDispatchedError < OperationsError
-      def initialize
-        super(I18n.t('orders_package.already_dispatched'))
-      end
+      raise Goodcity::AlredyDispatchedError.new if ord_pkg.dispatched?
```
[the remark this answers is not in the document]
patrixr: Welcome to the wonderful world of Macbook Pro Keyboards !

PR #861, opened 2019-11-27
GCW-2881 inventory number already exists fix

abhinav098:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2881

### What does this PR do?

BUG: If there is any issue in the item form, or subform, it throws an error on API side which prevents it from creating bad data. But, however, item was getting created on stockit side as there was no check for this kind of behavior what so ever. After fixing the data and resubmitting, it used to throw remove node error which has been fixed in this PR https://github.com/crossroads/stock.goodcity/pull/556 .

Along with that, it also fixes some code, and make it better to understand now. 

### Impacted Areas

Item Creation on stock app. Stockit Item sync.

### Linked PR's:

https://github.com/crossroads/stock.goodcity/pull/556
https://bitbucket.org/crfdevs/stockit/pull-requests/69/remove-unwanted-validations/diff

### DIFF
--- a/app/models/package.rb
+++ b/app/models/package.rb
@@ -318,7 +318,8 @@ class Package < ActiveRecord::Base
   end
 
   def add_to_stockit
-    return unless detail && detail.valid?
+    return if detail.present? && !detail.valid?
+
     response = Stockit::ItemSync.create(self)
     if response && (errors = response["errors"]).present?
       errors.each { |key, value| self.errors.add(key, value) }
--- a/app/models/package.rb
+++ b/app/models/package.rb
@@ -318,6 +318,7 @@ class Package < ActiveRecord::Base
   end
 
   def add_to_stockit
+    return unless detail && detail.valid?
     response = Stockit::ItemSync.create(self)
     if response && (errors = response["errors"]).present?
       errors.each { |key, value| self.errors.add(key, value) }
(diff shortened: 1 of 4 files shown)


abhinav098:
@steveyken, do you want me to remove all the validations ? or just the custom validations that I have added in this PR? 

For `mar_os_serial_num` and `mar_ms_office_serial_num` => "14 char" validation is present currently along with the custom one. 

Please let me know. 

Thanks

app/models/package.rb
```diff
@@ -318,6 +318,8 @@ def published?
   end
 
   def add_to_stockit
+    return if detail.present? && !detail.valid?
```
patrixr: If there is such a thing as "detail validity" I'd suggest having a validation on the model to prevent it from being persisted in the first place. Or at least raise an error rather than returning silently.

Since this line doesn't throw, it will surely result in future bugs where we ask "why isn't it syncing with Stockit??"

steveyken: agree, removing this code should pass the validation off to Stockit anyway... see the error catching below

abhinav098: There is a validity on API side, for this and it doesn't allow to create a record on API. But it is going and creating the record on Stockit. 

That's why I have stopped it here itself.

patrixr: It should still throw rather than return. When writing the function you can't **assume** where and when it's being called, or what else is happening.

abhinav098: sure, makes sense.

PR #862, opened 2019-11-29
Release - move refactor + inventory number fix

patrixr:
Release changes :
* Remove old move endpoints (#854) - 2ae7ce6
* GCW-2881 inventory number already exists fix - 65e5c5b


config/locales/zh-tw.yml
```diff
@@ -98,4 +101,5 @@ zh-tw:
     dispatch:
       unprocessed_order: Cannot dispatch packages from an unprocessed order
     move:
-      not_enough_at_source: Not enough packages available to move from that location
\ No newline at end of file
+      not_enough_at_source: Not enough packages available to move from that location
```
namrataukirde: need to add Chinese translation here. Not a blocker.

patrixr: got it fixed in an upcoming pr : )

PR #868, opened 2019-12-02
GCW-2784 multiple printer feature

namrataukirde:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2784

### What does this PR do?

FEATURE: Multiple printer implementation.

### Impacted Areas

Admin app and stock app. Print inventory number features.

Please review.

Thanks.


app/services/label/base_label.rb
```diff
@@ -0,0 +1,35 @@
...
+    def initialize(dots, width, field_orientation, height)
+      self.print_count = nil
+      self.label       = Easyzpl::Label.new(dots: dots,
+                                            width: width,
+                                            field_orientation: field_orientation,
+                                            height: height)
+    end
+
+    def tmp_label_file
+      @file = Tempfile.new("cupsjob")
```
steveyken:
```suggestion
      @file = Tempfile.new("cupsjob") do |file|
        file.write(label.to_s)
      end
```

Using a block automatically closes the file.

app/services/label/inventory_label.rb
```diff
@@ -0,0 +1,47 @@
...
+    INVENTORY_LABEL_HEIGHT   = 1
+    INVENTORY_LABEL_FIELD_ORIENTATION = :landscape
+
+    def initialize(inventory_number, print_count)
+      super(INVENTORY_LABEL_DOTS, INVENTORY_LABEL_WIDTH, INVENTORY_LABEL_FIELD_ORIENTATION, INVENTORY_LABEL_HEIGHT)
+      self.inventory_number = inventory_number
+      self.print_count      = print_count
+    end
+
+    def design_label
```
steveyken: How about calling this method `draw`?

app/services/label/inventory_label.rb
```diff
@@ -0,0 +1,47 @@
...
+    BARCODE_QR_URL = "https://redirect.crossroads.org.hk/inventory?num=".freeze
+
+    INVENTORY_LABEL_DOTS = 300
+
+    # The dimensions of the label
+    INVENTORY_LABEL_WIDTH    = 2
+    INVENTORY_LABEL_HEIGHT   = 1
+    INVENTORY_LABEL_FIELD_ORIENTATION = :landscape
+
+    def initialize(inventory_number, print_count)
```
steveyken: I think it might be better to use Rails `options` hash style arguments here as it allows more flexibility in the parameters that are set... some labels will set different arguments (e.g. inventory_number) and some will have different params. By using `options` you can have a unified argument across all label types whilst being flexible enough for each particular label's needs.

I'll branch off your branch here with some suggestions and you can merge in if you like it.

patrixr: One other benefit we often forget is that options tell us what it means without looking at the method's code

`InventoryLabel.new('123', 3)` -> we don't know what `3` stands for without opening up the file with the method's code
`InventoryLabel.new('123', print_count: 3)` -> we immediately know what it means

❤️ Ruby options ❤️

app/services/label/inventory_label.rb
namrataukirde: Sure.. We can change method name ..  :)

namrataukirde:
As we need it on staging. Suggested changes will be done in subsequent PR as discussed in scrum call.

PR #870, opened 2019-12-04
Fuzzy organisation search

patrixr:
This is a first version of a `FuzzySearch` module, which I've plugged in to the Organisations model

Notes:

It is far from perfect, but Nokia mentioned people were having many problems especially with searches like : `'The ABC Foundation'` which are stored in the database as `'ABC Foundation, The'`

Speed can become a potential issue when searching through many fields, luckily that's not the case for Organisations. If we ever want to use it with Orders, we'll definitely need to assess performance and make improvements.

Other:

- Added a couple of sanity checks to the packages_inventory_importer. The changes were tiny so I squeezed them into this PR

### DIFF
--- a/app/controllers/api/v1/gc_organisations_controller.rb
+++ b/app/controllers/api/v1/gc_organisations_controller.rb
@@ -42,7 +42,7 @@ module Api::V1
       if params['ids'].present?
         records = @organisations.where(id: params['ids']).page(params["page"]).per(params["per_page"] || DEFAULT_SEARCH_COUNT)
       else
-        records = @organisations.search(params["searchText"]).page(params["page"]).per(params["per_page"] || DEFAULT_SEARCH_COUNT)
+        records = Organisation.search(params["searchText"]).page(params["page"]).per(params["per_page"] || DEFAULT_SEARCH_COUNT)
       end
       data = ActiveModel::ArraySerializer.new(records, each_serializer: serializer, root: "gc_organisations").as_json
       render json: { "meta": { total_pages: records.total_pages, "search": params["searchText"] } }.merge(data)
--- a/app/controllers/api/v1/gc_organisations_controller.rb
+++ b/app/controllers/api/v1/gc_organisations_controller.rb
@@ -42,7 +42,7 @@ module Api::V1
       if params['ids'].present?
         records = @organisations.where(id: params['ids']).page(params["page"]).per(params["per_page"] || DEFAULT_SEARCH_COUNT)
       else
-        records = Organisation.search(params["searchText"]).page(params["page"]).per(params["per_page"] || DEFAULT_SEARCH_COUNT)
+        records = @organisations.search(params["searchText"]).page(params["page"]).per(params["per_page"] || DEFAULT_SEARCH_COUNT)
       end
       data = ActiveModel::ArraySerializer.new(records, each_serializer: serializer, root: "gc_organisations").as_json
       render json: { "meta": { total_pages: records.total_pages, "search": params["searchText"] } }.merge(data)
--- a/app/models/concerns/fuzzy_search.rb
+++ b/app/models/concerns/fuzzy_search.rb
@@ -30,19 +30,17 @@ module FuzzySearch
     #     SIMILARITY(name_zh_tw, 'steve') DESC
     #
     def search(search_text)
-      sanitized_text = sanitize(search_text)
-      similarities = self.search_props.map { |f| "SIMILARITY(#{f}, #{sanitized_text})" }
+      similarities = search_props.map { |f| "SIMILARITY(#{f}, #{sanitize(search_text)})" }
 
       fields = similarities + ["#{table_name}.*"]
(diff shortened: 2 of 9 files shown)


app/controllers/api/v1/gc_organisations_controller.rb
```diff
@@ -42,7 +42,7 @@ def find_record_and_render_json(serializer)
       if params['ids'].present?
         records = @organisations.where(id: params['ids']).page(params["page"]).per(params["per_page"] || DEFAULT_SEARCH_COUNT)
       else
-        records = @organisations.search(params["searchText"]).page(params["page"]).per(params["per_page"] || DEFAULT_SEARCH_COUNT)
+        records = Organisation.search(params["searchText"]).page(params["page"]).per(params["per_page"] || DEFAULT_SEARCH_COUNT)
```
steveyken:
```suggestion
        records = @organisations.search(params["searchText"]).page(params["page"]).per(params["per_page"] || DEFAULT_SEARCH_COUNT)
```

patrixr: @steveyken Good point, however the `search` field isn't a scope. Since I had to do a custom `select` with the similarities within it. Any suggestions ?
 can we "update" the select of an existing relation ?

app/models/concerns/fuzzy_search.rb
```diff
@@ -0,0 +1,69 @@
...
+    # @return [<Type>] <description>
+    #
+    def configure_search(props: [], tolerance: 0.1)
+      @search_props ||= []
+      @search_props = (@search_props << props).flatten
+      @similarity_threshold = tolerance
+    end
+
+    def similarity_threshold
+      @similarity_threshold.present? ? @similarity_threshold : 0.1
```
abhinav098: @patrixr , I don't this this method is required, you can set `attr_accessor :similarity_threshold` which will get set in configure search. 

This line is explicitly passing tolerance as `0.1`. Once it is set to an attribute accessor you can use it anywhere in the class, no need of this ternary operator.
https://github.com/crossroads/api.goodcity/pull/870/files#diff-adfd80ef5cd384d3af22bdfc27cb7a76R11

abhinav098: Please let me know if i'm missing anything.

app/models/organisation.rb
```diff
@@ -1,15 +1,14 @@
...
   belongs_to :district
   has_many :organisations_users
   has_many :orders
   has_many :users, through: :organisations_users
 
-  def self.search(search_text)
-    where("name_en ILIKE :search_text OR name_zh_tw ILIKE :search_text",
-      search_text: "%#{search_text}%")
-  end
+  configure_search props: [:name_en, :name_zh_tw], tolerance: 0.1
```
abhinav098: @patrixr , use %i for an array of symbols. (Code Climate)

app/controllers/api/v1/gc_organisations_controller.rb
abhinav098: @patrixr , What's the reason for removing `@organisations` ? 

According to cancan, @organisations will contain the organisations filtered based on permissions and using `Organisation` class will expose all the organisations to this search method.
 
Please let me know if it's intentional or am I missing something ?

abhinav098: @patrixr I think you can scope it in ClassMethods
```
module FuzzySearch
  module ClassMethods
    def search
    end
  end
end
```


I think `@organisations.search` will work. Check `package_filtering.rb` and `packages_controller:151`

app/models/organisation.rb
patrixr: @abhinav098 I'm afraid I disagree with this practice. 
If a good explanation is provided for it then I'll be happy to change it. But so far I don't understand why it is of any value

app/controllers/api/v1/gc_organisations_controller.rb
patrixr: Yeah I need to figure it out, Steve's comment addresses it. But I don't know how to add a custom select whilst keeping a scope search.

@organisations.search would not work unless it's a scope (based on your comment above), unless I'm mistaken.
But scopes expect to be chainable, which contradicts the custom select I created. It's not clear to me either

patrixr: If you take a look at the SELECT statement you'll know what I mean

app/models/concerns/fuzzy_search.rb
patrixr: @abhinav098 
The configure_search call is optional, meaning we have to handle cases where it isn't called. That's what this is about.

@similarity_threshold could be nil, therefore we always account for that scenario by defining a method to wrap it

app/controllers/api/v1/gc_organisations_controller.rb
patrixr: @abhinav098 @steveyken It looks like selects are chainable !! 🎉 🎉 

Which weirds me out, looks and feels wrong. But that means that I can actually use a scope : )

PR #871, opened 2019-12-05
GCW-2774 allow creation of boxes and pallets

abhinav098:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2774

### What does this PR do?

Feature: Support creating boxes or pallets

### Impacted Areas

Package creation

### Mockup Link

https://xd.adobe.com/view/224f7314-f1c0-40b6-7d7a-929e616be6bb-547e/screen/b329497c-a23f-4063-89c8-9ae7b71d253a/Main-Menu?x_product=xd-jiraserver/1.0

### Linked PR's:

Stock PR: https://github.com/crossroads/stock.goodcity/pull/566


app/controllers/api/v1/packages_controller.rb
```diff
@@ -431,6 +432,11 @@ def assign_detail
         ).build_or_update_record
       end
 
+      def assign_storage_type
+        return unless params["package"]["storage_type"]
```
patrixr: should we maybe have a default storage type for the "normal" package (maybe 'unit') ? 🤔

abhinav098: Yep, we can have a default one, but need to decide it's name. Maybe @steveyken can suggest some name for it. 

Also, we can say that if storage_type is nil, it's a package and not anything else. 

I'm open for both. 

Let me know your thoughts on this @patrixr and @steveyken :)

abhinav098: As discussed today in the call, I'll create a storage type for "normal" package. By the name of "Package".

app/models/concerns/settings_validator.rb
```diff
@@ -0,0 +1,24 @@
...
+        error_message_key = "activerecord.errors.models.#{record.class.name.underscore}.#{key.split('.').last}"
+        error_messages << I18n.t(error_message_key)
+      end
+    end
+    record.errors.add(:base, error_messages.join(" ")) if error_messages.length > 0
+  end
+
+  private
+
+  def action_allowed?(key)
```
patrixr: nice, could be nice to have this method on the model itself perhaps ?  e.g `GoodcitySetting.enabled?(key)` 
just a thought

abhinav098: Yeah, I think you are right. I'll make a change. 
Thanks @patrixr .

PR #876, opened 2019-12-14
Gcw 2784 multiple printer feature additions

steveyken:
### Ticket Link: 

GCW-2784

### What does this PR do?

* `PrintJob` is now responsible for instantiating printer 
* Moved label folder to labels
* Used sockets instead of temp files
* Moved constants into `options` hash


steveyken:
I've confirmed the socket printing works on staging if host ip is set correctly instead of 'localhost'. This bypasses the need for using CUPS setup on the local machine.

namrataukirde:
:+1:

steveyken:
This was squashed into dd252e64615cb7fdc41f1862b4724220b2ceaf8a so closing

PR #877, opened 2019-12-17
Printer refactoring

steveyken:
Squashed commits

Move file creating out of label and into PrintLabel control class.

Use sockets to print rather than files.

Removed unnecessary file


steveyken:
Does this need to be merged to master? If not, please close when ready.

PR #885, opened 2019-12-31
GCW-2931-Request cancellation reason for orders

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2931

### What does this PR do?

FEATURE: 
 1) Adding cancellation Reason for Orders on stock 
 2) assigning default cancellation_reason of "Changed mind".  
 3) removing cancellation reason on resubmitting the order.

### Impacted Areas

Go to specific order. Click "Cancel order" popup appears with cancellation message and options.

### Screenshots

image-2019-12-17-14-57-24-480

### DIFF
--- a/app/controllers/api/v1/cancellation_reasons_controller.rb
+++ b/app/controllers/api/v1/cancellation_reasons_controller.rb
@@ -4,7 +4,8 @@ module Api
       load_and_authorize_resource :cancellation_reason, parent: false
 
       def index
-        render json: cancellation_reasons, each_serializer: serializer
+        reasons = @cancellation_reasons.cancellation_reasons_for(params['for'])
+        render json: reasons, each_serializer: serializer
       end
 
       def show
@@ -15,13 +16,6 @@ module Api
 
       def serializer
         Api::V1::CancellationReasonSerializer
-      end
-
-      def cancellation_reasons
-        @cancellation_reasons.cancellation_reasons_by({
-          ids: params["ids"],
-          offer: params["offer"]
-        })
       end
     end
   end
--- a/app/controllers/api/v1/cancellation_reasons_controller.rb
+++ b/app/controllers/api/v1/cancellation_reasons_controller.rb
@@ -4,7 +4,7 @@ module Api
       load_and_authorize_resource :cancellation_reason, parent: false
 
       def index
-        render json: get_cancellation_reasons, each_serializer: serializer
+        render json: cancellation_reasons, each_serializer: serializer
       end
 
       def show
@@ -17,7 +17,7 @@ module Api
         Api::V1::CancellationReasonSerializer
       end
 
-      def get_cancellation_reasons
+      def cancellation_reasons
         @cancellation_reasons.get_cancellation_reasons_by({
           ids: params["ids"],
           offer: params["offer"],
--- a/app/controllers/api/v1/cancellation_reasons_controller.rb
+++ b/app/controllers/api/v1/cancellation_reasons_controller.rb
@@ -18,10 +18,9 @@ module Api
       end
 
       def cancellation_reasons
-        @cancellation_reasons.get_cancellation_reasons_by({
+        @cancellation_reasons.cancellation_reasons_by({
           ids: params["ids"],
-          offer: params["offer"],
-          order: params["stock"]
+          offer: params["offer"]
         })
       end
     end
--- a/app/controllers/api/v1/cancellation_reasons_controller.rb
+++ b/app/controllers/api/v1/cancellation_reasons_controller.rb
@@ -4,8 +4,7 @@ module Api
       load_and_authorize_resource :cancellation_reason, parent: false
 
       def index
-        @cancellation_reasons = @cancellation_reasons.get_cancellation_reasons_by(params)
-        render json: @cancellation_reasons, each_serializer: serializer
+        render json: get_cancellation_reasons, each_serializer: serializer
       end
 
       def show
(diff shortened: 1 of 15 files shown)


app/controllers/api/v1/cancellation_reasons_controller.rb
```diff
@@ -4,12 +4,7 @@ class CancellationReasonsController < Api::V1::ApiController
       load_and_authorize_resource :cancellation_reason, parent: false
 
       def index
-        if params["ids"].present?
-          @cancellation_reasons = @cancellation_reasons.where(id: params["ids"].split(",").flatten)
-        else
-          @cancellation_reasons = @cancellation_reasons.visible
-        end
-        render json: @cancellation_reasons, each_serializer: serializer
+        render json: cancellation_reasons, each_serializer: serializer
```
patrixr: `@cancellation_reasons` ?

app/controllers/api/v1/cancellation_reasons_controller.rb
```diff
@@ -4,12 +4,7 @@ class CancellationReasonsController < Api::V1::ApiController
       load_and_authorize_resource :cancellation_reason, parent: false
 
       def index
-        if params["ids"].present?
-          @cancellation_reasons = @cancellation_reasons.where(id: params["ids"].split(",").flatten)
-        else
-          @cancellation_reasons = @cancellation_reasons.visible
-        end
-        render json: @cancellation_reasons, each_serializer: serializer
```
steveyken: Good refactor. Nice and simple!

app/controllers/api/v1/cancellation_reasons_controller.rb
```diff
@@ -21,6 +16,13 @@ def show
       def serializer
         Api::V1::CancellationReasonSerializer
       end
+
+      def cancellation_reasons
+        @cancellation_reasons.cancellation_reasons_by({
+          ids: params["ids"],
+          offer: params["offer"]
```
steveyken: Rather than `offer` this should be more generic. At the moment we have `offer` and `order` but imagine if we want to add a third or forth option. Could change it to something more like `for` where the allowed options are only `offer` and `order` - sanitize the params

app/models/cancellation_reason.rb
```diff
@@ -15,4 +19,14 @@ def self.unwanted
   def self.donor_cancelled
     find_by(name_en: "Donor cancelled")
   end
+
+  def self.cancellation_reasons_by(options)
+    if options[:ids]
+      where(id: options[:ids].split(",").flatten)
```
steveyken: I know you've left this in as it was there before but I don't think ember actually uses `params[:ids]` so it can be dropped. This would simplify the method to something like

```
def cancellation_reasons_for(for)
  return where() unless CANCELLATION_REASONS_TYPE.include(for)
  send("visible_to_#{for}")
end
```

app/controllers/api/v1/cancellation_reasons_controller.rb
steveyken: Once you've made the above change, I would strongly advocate to remove the `cancellation_reasons` method and put the code into the `index` method to make the code more readable.

app/controllers/api/v1/orders_controller.rb
```diff
@@ -69,12 +69,7 @@ def show
       end
 
       def transition
-        transition_event = params['transition'].to_sym
-        cancellation_reason = params["cancellation_reason"]
-        if @order.state_events.include?(transition_event)
-          @order.fire_state_event(transition_event)
-          @order.update(cancellation_reason: cancellation_reason) if cancellation_reason.presence
-        end
+        update_transition_and_cancel_reason
```
steveyken: I would rather not move this code into a separate `update_transition_and_cancel_reason`. It's now only 2 lines and doesn't warrant it's own method unless it's being reused elsewhere but I don't see that. IT makes it hard to follow the code path if having to jump around too much, esp with many short methods.

app/controllers/api/v1/orders_controller.rb
```diff
@@ -123,6 +118,18 @@ def order_response(records)
           exclude_stockit_set_item: true).as_json
       end
 
+      def update_transition_and_cancel_reason
+        event = params['transition'].to_sym
+        @order.update_transition_and_reason(event, cancel_params) if @order.state_events.include?(event)
```
steveyken: Good idea to move the code into the model - logic like this should be in the model so that it can be used elsewhere (such as rake tasks and other classes.) controllers are about transactional flow and models are about data so you've definitely made the right choice here.

Now that this method is only 2 lines, I would put it back up in the `transition` method so that there are fewer jumps to make when reading the code to see what is going on. (Same comment as above)

app/models/order.rb
```diff
@@ -68,13 +69,13 @@ class Order < ActiveRecord::Base
...
   # Stockit Shipment Status => GoodCity State
-  SHIPMENT_STATUS_MAP = { 
+  SHIPMENT_STATUS_MAP = {
     "Processing" => "processing",
     "Sent" => "closed",
     "Loaded" => "dispatching",
     "Cancelled" => "cancelled",
     "Upcoming" => "awaiting_dispatch"
-  }
+  }.freeze
```
steveyken: :+1:

app/models/order.rb
```diff
@@ -118,6 +119,11 @@ def delete_orders_packages
     end
   end
 
+  def update_transition_and_reason(event, opts)
+    fire_state_event(event)
+    update(cancellation_reason_id: opts[:cancellation_reason_id], cancel_reason: opts[:cancel_reason]) if opts[:cancellation_reason_id]
```
steveyken: Given that the keys in `opts` are the same as the field names, you can just do `update(opts)` as you have sanitized the allowed options in the `params` method of the controller. An even safer way to do this would be...

```
def update_transition_and_reason(event, cancel_opts)
  fire_state_event(event)
  opts = opts.select{|k,v| ['cancellation_reason_id', 'cancel_reason'].include?(k) }
  update(opts)
end
```

abhinav098:
PR looks good now 👍 , please resolve the conflicts.

PR #890, opened 2020-01-08
[Patch] Release: Change action name from 'gain/loss' to 'move'

patrixr:
Post-release fix

I noticed the move operation was registering a 'gain' and 'loss' instead of the 'move' action.
As of now it has no impact because loss/gain aren't yet used
This PR simply replaces the action used to 'move'

This has no impact on any functionality, it just needs to match the correct action for the future


abhinav098:
Looks good. Please make sure to add a release tag. 👍

PR #891, opened 2020-01-08
GCW-2859 box pallet add/remove items

abhinav098:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2859

### What does this PR do?

FEATURE: Allows user to add/remove items to box/pallet

### Mockup Link

https://xd.adobe.com/view/224f7314-f1c0-40b6-7d7a-929e616be6bb-547e/screen/b329497c-a23f-4063-89c8-9ae7b71d253a/Main-Menu?x_product=xd-jiraserver/1.0

### Linked PR's:
https://github.com/crossroads/stock.goodcity/pull/587


lib/classes/box_pallet_manager.rb
```diff
@@ -1,9 +1,11 @@
 class BoxPalletManager
```
patrixr: @abhinav098  All the logical operations are currently in the `concerns/operations` folder. Could you add this feature there ? We're trying to consolidate all the business logic in the same place
There is a `stock_operation.rb` which would work quite well for the box features.

The `BoxPalletManager` can still exist under the hood, but would like to have all the entry points in the same place :)

It populates the `Package::Operations` namespace. Meaning that adding methods there would allow us to do `Package::Operations.pack(package, box: box, quantity: 5)` (or something like that)

abhinav098: sure @patrixr , thanks for the early feedback.

patrixr: Thanks ! Feel free to fiddle with the code 😄 

I'm not dead set on having operations as concerns in that specific way, a namespace `Goodcity::Operations` would also work nicely. I'm open to any change.

The important thing is that we group them and maintain them as a whole 💪

app/models/concerns/operations/stock_operations.rb
```diff
@@ -29,6 +29,56 @@ def register_loss(package, quantity:, from_location:)
...
+          return { packages_inventory: pkg_inventory, success: true }
+        elsif pkg_inventory&.errors
+          return { errors: pkg_inventory.errors.full_messages, success: false }
+        end
+      end
+
+      private
+
+      def register_change
+        return unless action
```
patrixr: I'd personally be in favour of throwing an error in this case. Otherwise a bad action might go unnoticed

app/serializers/api/v1/stockit_item_serializer.rb
```diff
@@ -31,6 +32,32 @@ def include_order?
...
+    def include_in_hand_quantity?
+      @options[:include_in_hand_quantity]
+    end
+
+    def include_added_quantity?
+      @options[:include_added_quantity]
+    end
+
+    def in_hand_quantity
+      sql = "select sum(quantity) from packages_inventories where package_id = #{object.id}"
```
steveyken: do we know what the performance impact of this is?

abhinav098: @steveyken , There won't be much, because it's not called for all items, `in_hand_quantity` it's called only for 25 items at a time, while searching items for adding it to a box. 
So I don't think this will make any performance impact. 
I have used this in the serializer, 
```  def include_in_hand_quantity?
      @options[:include_in_hand_quantity]
    end```

It is false otherwise, but becomes true only when we're searching for items to add on a box or a pallet. 

Using this param I pass this option to the serializer 
`include_in_hand_quantity: !!params["filter_box_pallet"],`
and this param is sent from the client only while searching for items to add on a box or a pallet.

Thanks

Comment (steveyken) [reply app/serializers/api/v1/stockit_item_serializer.rb]: Thanks! A very comprehensive answer 💯

File: app/controllers/api/v1/packages_controller.rb [hunk 4]
```diff
@@ -161,7 +162,7 @@ def search_stockit_items
             with_inventory_no: params["withInventoryNumber"] == "true"
           )
         end
-        params_for_filter = %w[state location].each_with_object({}) { |k, h| h[k] = params[k] if params[k].present? }
+        params_for_filter = %w[state location associated_package_types].each_with_object({}) { |k, h| h[k] = params[k] if params[k].present? }
```
swatijadhav:
```suggestion
        params_for_filter = %w[state location associated_package_types].each_with_object({}) { |k, h| h[k] = params[k].presence }
```

app/controllers/api/v1/packages_controller.rb
```diff
@@ -225,6 +227,31 @@ def send_stock_item_response
...
+        response = Package::Operations.pack_or_unpack(params, User.current_user.id)
+        if response[:success]
+          render json: { packages_inventories: response[:packages_inventory] }, status: 201
+        else
+          render json: { errors: response[:errors] }, status: 422
+        end
+      end
+
+      def fetch_associated_packages
+        entity = Package.find(params[:id]) if params[:id] # fetch or pallet
```
swatijadhav: Should we use `find_by(id: params[:id])` ? as on next line we return if `package` is not present.

app/controllers/api/v1/packages_controller.rb
abhinav098: 👍

app/controllers/api/v1/packages_controller.rb
abhinav098: Thanks @swatijadhav , good catch. Yes it should be `find_by` as find will throw exception incase of not found.

app/models/concerns/operations/stock_operations.rb
```diff
@@ -68,5 +70,98 @@ def register_loss(package, quantity:, from_location:)
         location: from_location
       )
     end
+
+    def pack_or_unpack(params, user_id)
+      raise ActionNotAllowedError.new unless PACK_UNPACK_ALLOWED_ACTIONS.include?(params[:task])
+      PackUnpack.new(params, user_id).public_send(params[:task])
```
swatijadhav: @abhinav098 here not able to locate the definition of `public_send`.. can you please point it?

app/controllers/api/v1/packages_controller.rb
```diff
@@ -238,7 +238,7 @@ def add_remove_item
       end
 
       def fetch_associated_packages
-        entity = Package.find(params[:id]) if params[:id] # fetch or pallet
+        entity = Package.find_by(params[:id]) if params[:id] # fetch box or pallet
```
swatijadhav: @abhinav098 it should be `find_by(id: params[:id])`

refer: https://github.com/crossroads/api.goodcity/pull/891#discussion_r376894703

```suggestion
        entity = Package.find_by(id: params[:id]) if params[:id] # fetch box or pallet
```

app/models/concerns/operations/stock_operations.rb
abhinav098: @swatijadhav , it's a ruby method. https://apidock.com/ruby/Object/public_send
Public send calls the public functions of a class.

app/controllers/api/v1/packages_controller.rb
abhinav098: aahh.. my bad, sorry, fixed the typo @swatijadhav .

app/models/concerns/operations/stock_operations.rb
```diff
@@ -68,5 +70,98 @@ def register_loss(package, quantity:, from_location:)
...
+      end
+
+      def item_designated?
+        @item.order.presence
+      end
+    end
+
+    # --- Exceptions
+
+    class OperationsError < StandardError; end
```
patrixr: All the errors have been moved to `lib/goodcity/errors.rb` , you might want to move those there : )

app/models/concerns/inventory_computer.rb
```diff
@@ -29,6 +29,18 @@ def location_quantity(location)
         quantity_where(location: location)
       end
 
+      def total_quantity_in_box(box)
+        historical_quantity
+          .where(source: box, action: ['pack', 'unpack'])
+          .as_of_now
+      end
+
+      def quantity_of_package_in_box(package:)
```
patrixr: I feel like this doesn't make a lot of sense unless we specify which box to look into.
This will return the quantity in multiple boxes.

app/controllers/api/v1/packages_controller.rb
```diff
@@ -225,6 +227,31 @@ def send_stock_item_response
...
+        render nothing: true, status: 204 and return if params[:quantity].to_i.zero?
+        response = Package::Operations.pack_or_unpack(params, User.current_user.id)
+        if response[:success]
+          render json: { packages_inventories: response[:packages_inventory] }, status: 201
+        else
+          render json: { errors: response[:errors] }, status: 422
+        end
+      end
+
+      def fetch_associated_packages
```
patrixr: the term `associated_packages` feels a bit generic. Maybe something like `contained_packages` would be more descriptive 🤔

app/models/concerns/operations/stock_operations.rb
```diff
@@ -68,5 +70,98 @@ def register_loss(package, quantity:, from_location:)
...
+
+    def pack_or_unpack(params, user_id)
+      raise ActionNotAllowedError.new unless PACK_UNPACK_ALLOWED_ACTIONS.include?(params[:task])
+      PackUnpack.new(params, user_id).public_send(params[:task])
+    end
+
+    class PackUnpack
+      def initialize(params, user_id)
+        @cause = Package.find(params[:id]) # box or pallet
+        @item = Package.find(params[:item_id]) # item to add or remove
```
patrixr: could we use the term `package` instead of `item`. Item has a meaning in the api, it is a separate model

app/models/concerns/operations/stock_operations.rb
```diff
@@ -68,5 +70,98 @@ def register_loss(package, quantity:, from_location:)
         location: from_location
       )
     end
+
+    def pack_or_unpack(params, user_id)
```
patrixr: Recommendation: pass in the params as ruby options arguments, e.g

```
def pack_or_unpack(box_id:, package_id:, :quantity, user_id:)
    ...
end
```

Otherwise we have no idea what we're supposed to pass to this method, or we need to dive deep in the code to figure it out

app/serializers/api/v1/stockit_item_serializer.rb
```diff
@@ -31,6 +32,30 @@ def include_order?
...
 
+    def include_added_quantity?
+      @options[:include_added_quantity]
+    end
+
+    def include_in_hand_quantity?
+      @options[:include_in_hand_quantity]
+    end
+
+    def in_hand_quantity
```
patrixr:
```suggestion
def on_hand_quantity
```

patrixr: Note: Currently working on adding counter cache columns to the package table. Which include `on_hand_quantity`, so I'll remove those methods once that's done : )

app/serializers/api/v1/stockit_item_serializer.rb
```diff
@@ -31,6 +32,30 @@ def include_order?
...
+      object.total_in_hand_quantity
+    end
+
+    def in_hand_quantity__sql
+      "select sum(quantity) from packages_inventories where package_id = #{object.id}"
+    end
+
+    def added_quantity
```
patrixr: The way this `added_quantity` is used doesn't make a lot of sense.

It defines "the quantity which is in all of the boxes"
But according to the Stock PR, it is used to "remove that quantity from a single box"

I think the frontend, for a given package, should fetch the list of boxes and quantities. And use that to trigger pack/unpack commands. Having a single property on the serializer seems faulty to me

app/models/concerns/operations/stock_operations.rb
abhinav098: sure @patrixr , I'll move those errors to that class.

app/models/concerns/inventory_computer.rb
abhinav098: ahh, good catch @patrixr . Will work on fixing this method.

app/controllers/api/v1/packages_controller.rb
abhinav098: haha, sure @patrixr .

app/models/concerns/operations/stock_operations.rb
abhinav098: sure @patrixr .

app/models/concerns/operations/stock_operations.rb
abhinav098: Makes sense. Will do that. 👍

app/serializers/api/v1/stockit_item_serializer.rb
abhinav098: Sure 👍

app/serializers/api/v1/stockit_item_serializer.rb
abhinav098: sure, will check and fix that.

app/models/concerns/operations/stock_operations.rb
patrixr: Thanks : )

PR #892, opened 2020-01-09
OrdersPackage: dispatched quantity

patrixr:
### Ticket Link: 

First part of https://jira.crossroads.org.hk/browse/GCW-2942

### Changes

* Addition of the `dispatched_quantity` column to `OrdersPackaged`. That column is computed when the `packages_inventory` table is updated
* Fix of computation of `PackagesInventory::Computer::dispatched_quantity(...)` which had a bug
* Fix of operation `StockOperations::register_loss` which had a badly written tests which where hiding a bug


lib/tasks/orders_packages.rake
```diff
@@ -0,0 +1,16 @@
...
+      If the orders_package is marked as dispatched, sets the `dispatched_quantity` field to the quantity designated
+    DESC
+    task init_dispatched_quantity: :environment do
+      count = 0
+      OrdersPackage.dispatched.find_each do |orders_package|
+        orders_package.update(dispatched_quantity: orders_package.quantity)
+        count += 1
+      end
+      puts "Task completed. #{count} records modified"
+    end
```
steveyken: This is fine, Can also run a direct SQL query inside the migration to make the upgrade smooth. I'll put it on there as a comment for you to decide..

db/migrate/20200107034249_add_dispatched_quantity_to_orders_packages.rb
```diff
@@ -0,0 +1,5 @@
+class AddDispatchedQuantityToOrdersPackages < ActiveRecord::Migration
+  def change
+    add_column :orders_packages, :dispatched_quantity, :integer, default: 0
```
steveyken: Can also add the following instead of the rake task. See https://guides.rubyonrails.org/active_record_migrations.html#using-reversible

```
    reversible do |dir|
      dir.up do
        # initialise the dispatched_quantity values
        execute <<-SQL
          UPDATE orders_packages SET dispatched_quantity=quantity;
        SQL
      end
    end
```

app/models/orders_package.rb
```diff
@@ -33,6 +33,18 @@ module States
     ])
   }
 
+  PackagesInventory.on [:dispatch, :undispatch] do |pkg_inv|
+    # Compute 'dispatched_quantity' column on change
+    if pkg_inv.source_id.present? || pkg_inv.source_type.eql?('OrdersPackage')
```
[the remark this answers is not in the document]
patrixr: good catch 👍

lib/tasks/orders_packages.rake
patrixr: Oh definitely, I've been aching to do that. I just try to fit the standards

PR #894, opened 2020-01-14
GCW-2932-infinite scroll in active items

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2932

### What does this PR do?

FEATURE:  This PR adds infinite scroll to orders active items page. 
Till now we were making separate request on loading packages associated to orders. This PR removes this extra requests.

### Impacted Areas
Goto any specific Order >  Goto Active Items pages.

### DIFF
--- a/app/controllers/api/v1/orders_controller.rb
+++ b/app/controllers/api/v1/orders_controller.rb
@@ -58,12 +58,12 @@ module Api
         render json: serializer.new(@order,
           root: root,
           exclude_code_details: true,
-          include_packages: bool_param(:include_packages, true),
-          include_order: false,
+          include_packages: false,
+          include_order: true,
           include_territory: true,
           include_images: true,
           include_allowed_actions: true,
-          include_orders_packages: true,
+          include_orders_packages: false,
           exclude_stockit_set_item: true
         )
       end
--- a/app/controllers/api/v1/orders_controller.rb
+++ b/app/controllers/api/v1/orders_controller.rb
@@ -58,12 +58,12 @@ module Api
         render json: serializer.new(@order,
           root: root,
           exclude_code_details: true,
-          include_packages: false,
-          include_order: true,
+          include_packages: bool_param(:include_packages, true),
+          include_order: bool_param(:include_order, false),
           include_territory: true,
           include_images: true,
           include_allowed_actions: true,
-          include_orders_packages: false,
+          include_orders_packages: bool_param(:include_orders_packages, true),
           exclude_stockit_set_item: true
         )
       end
--- a/app/controllers/api/v1/orders_packages_controller.rb
+++ b/app/controllers/api/v1/orders_packages_controller.rb
@@ -87,7 +87,7 @@ module Api
       end
 
       def orders_package_by_order_id
-        records = @orders_packages.get_orders_package_by(params["order_id"])
+        records = @orders_packages.for_order(params["order_id"])
         orders_packages = records.page(params["page"]).per(params["per_page"])
         render json: ActiveModel::ArraySerializer.new(orders_packages,
           each_serializer: serializer,
--- a/app/controllers/api/v1/orders_packages_controller.rb
+++ b/app/controllers/api/v1/orders_packages_controller.rb
@@ -28,6 +28,7 @@ module Api
         return search_by_package_id if params['search_by_package_id'].present?
         # needs to be removed as it makes unwanted orders_packages request and makes the app slow
         return all_orders_packages if params['all_orders_packages'].present?
+        return orders_package_by_order_id if params['order_id'].present?
       end
 
       api :DELETE, '/v1/orders_package/1', "Delete an orders_package"
@@ -36,17 +37,6 @@ module Api
           @orders_package.delete_orders_package
         end
         render json: {}
-      end
-
-      def get_orders_op
-        records = OrdersPackage.joins(:order).where(orders: {id: params["order_id"]})
-        orders_packages = records.page(params["page"]).per(params["per_page"])
-        render json: ActiveModel::ArraySerializer.new(orders_packages,
-          each_serializer: serializer,
-          root: "orders_packages",
-          include_package: true,
-          include_orders_packages: true
-        ).as_json
       end
 
       api :PUT, '/v1/orders_packages/:id/actions/:action_name', 'Executes an action'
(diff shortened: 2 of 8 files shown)


app/serializers/api/v1/orders_package_serializer.rb
```diff
@@ -3,7 +3,7 @@ class OrdersPackageSerializer < ApplicationSerializer
     embed :ids, include: true
     attributes :id, :package_id, :order_id, :state, :quantity, :dispatched_quantity, :sent_on, :designation_id, :item_id, :created_at, :allowed_actions
 
-    has_one :package, serializer: PackageSerializer
+    has_one :package, serializer: PackageSerializer, root: :item
```
abhinav098: This might cause problems elsewhere. In stock we use `item` as root. But in browse it's used as `packages`.

I think it'll break on browse. Have you tested browse after this change ? 

cc- @namrataukirde

app/controllers/api/v1/orders_controller.rb
```diff
@@ -58,12 +58,12 @@ def show
...
-          include_order: false,
+          include_packages: false,
+          include_order: true,
           include_territory: true,
           include_images: true,
           include_allowed_actions: true,
-          include_orders_packages: true,
+          include_orders_packages: false,
           exclude_stockit_set_item: true
         )
```
abhinav098: Instead of putting true false explicitly. Try using app_name booleans. Such as `is_browse_app?` , `is_stock_app?`

Because this might break browse app.

cc - @namrataukirde

spec/models/orders_package_spec.rb
```diff
@@ -114,6 +114,16 @@
     end
   end
 
+  describe "#get_orders_package_by" do
+    it "return orders_package according to order_id" do
+      order = create :order
+      orders_package = create :orders_package, order_id: order.id
```
namrataukirde: how about using rspec create_list here? 🤔

app/controllers/api/v1/orders_controller.rb
patrixr: Agreed that changing those is risky, good catch. However the `is_browse_app?` conditionals have their own problems and we need to phase them out gradually.

I would recommend explicit options to pass in the query. Meaning keeping parameters such as `?include_packages=false`, adding any that you might need, and having the front-end request precisely what *it needs*

#### General server design idea:

The API cannot make _assumptions_ of what the front-end needs, ever.
As a "server", it literally **"serves"** whatever we request of it (within the confines of user permissions), in a generic matter.
It also exposes actions the users are permitted to take on the data

The front-end, on the other hand, is simply a _"window"_ to look into the data, and to connect to those actions.

Designing the API as a "service" is beneficial for future-proofing. Think of the public OAuth APIs such as [Twitter API](https://developer.twitter.com/en/docs/tweets/search/api-reference/get-search-tweets), [Github API](https://developer.github.com/v3/search/#search-repositories), ... Anyone can build an app for those services. We also happen to have a few apps for the Goodcity service, but we may have more/less in the future. Endpoints should therefore be flexible to support different uses.

Sorry for the rant, it felt like an opportune moment to share 😄 

cc - @abulsayyad123 @namrataukirde @abhinav098 @steveyken

app/models/orders_package.rb
```diff
@@ -26,6 +26,7 @@ module States
   scope :get_dispatched_records_with_order_id, ->(order_id) { where(order_id: order_id, state: 'dispatched') }
   scope :designated, ->{ where(state: 'designated') }
   scope :dispatched, ->{ where(state: 'dispatched') }
+  scope :get_orders_package_by, ->(order_id) { joins(:order).where(orders: {id: order_id}) }
```
patrixr: the name `get_orders_package_by` is confusing, since we don't know "by what" unless we look at the implementation.

Suggestion `scope :for_order`

app/serializers/api/v1/orders_package_serializer.rb
patrixr: @abhinav098 @abulsayyad123 @namrataukirde 

I suggest we create a JIRA ticket to refactor Stock's models. Meaning `item` into `package`, and `designation` into `order` to restore consistency. After which all this drama can go away

app/controllers/api/v1/orders_controller.rb
steveyken: Yes I do very much agree with this. We need to move away from things that tie our server state to the front-end. The API should never know which app it is talking too. However, there are various business logic processes that are tied just to the browse app or just to the stock app. Governing these with boolean params is a safer way to go.

app/serializers/api/v1/orders_package_serializer.rb
abulsayyad123: Yes, I am in 😄

app/controllers/api/v1/orders_packages_controller.rb
```diff
@@ -28,6 +28,7 @@ def index
         return search_by_package_id if params['search_by_package_id'].present?
         # needs to be removed as it makes unwanted orders_packages request and makes the app slow
         return all_orders_packages if params['all_orders_packages'].present?
+        return orders_package_by_order_id if params['order_id'].present?
```
patrixr: Note: all of the query parameters of this endpoint are _exclusive_ of each other, which is generally not how we want to build an entry point. Rather we want to be able to use the params in a _complementary_ way to allow a lot of flexibility in its usage.

No need to change anything, as it would mean a lot of extra work.
But to keep in mind for the future, if there's any questions I'm always happy to help.

cc - @abhinav098 @namrataukirde @abhinav098 @akshaygwd @steveyken

spec/controllers/api/v1/orders_packages_controller_spec.rb
```diff
@@ -30,6 +30,15 @@
...
       expect( subject["orders_packages"].size ).to eq(3)
     end
+
+    it "return orders_packages of order_id mentioned in params" do
+      order = create :order
+      order2 = create :order
+      create_list(:orders_package, 2, order_id: order.id)
+      create_list(:orders_package, 2,order_id: order2.id)
+      get :index, order_id: order.id
+      expect( subject["orders_packages"].size ).to eq(2)
```
patrixr: you probably want to test that the `orders_packages` returned are the correct ones.
Having 2 records returned doesn't mean much, they could be the wrong ones

spec/models/orders_package_spec.rb
```diff
@@ -114,6 +114,15 @@
     end
   end
 
+  describe "#for_order" do
+    it "return orders_package according to order_id" do
+      order = create :order
+      create_list(:orders_package, 2, order_id: order.id)
+      orders_package = create :orders_package
+      expect(OrdersPackage.for_order(order.id).size ).to eq(2)
```
patrixr: Same comment as mentioned above

PR #896, opened 2020-01-18
Added orders_packages_count to orders JSON payload

steveyken:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2932

### What does this PR do?

BUG: Provides the data to fix an incorrect count of 'Active' packages on the Order screen.

@abulsayyad123 this is just an API PR, you will need to use this count in a separate Ember PR (I've not written that). Also, note I've just implemented a count of the associated orders_packages, this includes all cancelled, designated and dispatched records. Is that the same as what is shown in the active tab?

### Impacted Areas

Orders > 12345

### Screenshots

Due to the infinite scroll feature, the count on the Active tab here is wrong.

image

The solution is to provide the total count in the JSON data 'meta' dict. For example:

image

### Linked PR's:

PR #894  PR #895 

https://crossroads-hk.slack.com/archives/GBJAFNVFB/p1579250342013900

### DIFF
--- a/app/controllers/api/v1/orders_controller.rb
+++ b/app/controllers/api/v1/orders_controller.rb
@@ -55,7 +55,7 @@ module Api
       api :GET, '/v1/designations/1', "Get a order"
       def show
         root = is_browse_app? ? "order" : "designation"
-        render json: serializer.new(@order,
+        serialized_orders = serializer.new(@order,
           root: root,
           exclude_code_details: true,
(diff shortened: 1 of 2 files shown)


steveyken:
I'm keen for input on this design approach as we will need to use it in other places to ensure the correct counts are shown for data that is not loaded (and therefore not computed on the client side.) For example, counts for orders on the package screen (the corollary to this page)

PR #897, opened 2020-01-18
Remove manage_orders_packages role and replace with finer grained security

steveyken:
### What does this PR do?

BUG: Non-admin users are able to view all orders_packages. This PR resolves that by restricting it to just ones that belong to orders they created or submitted.

NOTE: This shouldn't be merged until Browse and Stock have been well-tested for 401/403 errors

### JIRA ticket

https://jira.crossroads.org.hk/browse/GCW-2963

### Impacted Areas

* Order -> List / Show
* Browse -> Create order, View dashboard, View order


namrataukirde:
Changes look good to me. :+1:

I agree we need proper testing for 403 Error code here.

PR #904, opened 2020-01-29
Partial dispatch

patrixr:
### Allowing partial dispatch

* **before** `Operations.dispatch_full_qty(op)`
* **after**    `Operations.dispatch(op, from_location: opts[:location_id], quantity: opts[:quantity])`

### `OrdersPackage` allowed actions changes

An `orders_package` in a _designated_ state now has the following actions listed:

 * undispatch _(only if partially dispatched)_
 * cancel _(only if dispatch hasn't started)_

Meaning: If you partially dispatch an `orders_package`, you won't be able to cancel it unless you first undispatch whatever has been dispatched.

### Change of the `undispatch` action

The `undispatch` action now undispatches the already dispatched quantity. As of now there is no such thing as a partial undispatch.

## Other changes

* Fixed issue where the "designated" quantity of a package didn't take into account the partially dispatched quantity

* Fixed tests that would randomly fail due to Factories using `rand()` , `initialize_with` and other edge cases

### DIFF
--- a/app/models/orders_package.rb
+++ b/app/models/orders_package.rb
@@ -70,7 +70,7 @@ class OrdersPackage < ActiveRecord::Base
 
     before_transition on: :cancel do |orders_package, _transition|
       if orders_package.designated? && orders_package.dispatched_quantity.positive?
-        raise Goodcity::InvalidStateError.new(I18n.t('orders_package.cancel_requires_undispatch'));
+        raise Goodcity::InvalidStateError.new(I18n.t('orders_package.cancel_requires_undispatch'))
       end
       orders_package.quantity   = 0
       orders_package.updated_by = User.current_user
--- a/app/models/orders_package.rb
+++ b/app/models/orders_package.rb
@@ -69,6 +69,9 @@ class OrdersPackage < ActiveRecord::Base
     end
 
     before_transition on: :cancel do |orders_package, _transition|
+      if orders_package.designated? && orders_package.dispatched_quantity.positive?
+        raise Goodcity::InvalidStateError.new(I18n.t('orders_package.cancel_requires_undispatch'));
+      end
       orders_package.quantity   = 0
       orders_package.updated_by = User.current_user
     end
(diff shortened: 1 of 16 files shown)


app/models/orders_package.rb
```diff
@@ -69,6 +69,9 @@ def set_initial_state
     end
 
     before_transition on: :cancel do |orders_package, _transition|
+      if orders_package.designated? && orders_package.dispatched_quantity.positive?
+        raise Goodcity::InvalidStateError.new(I18n.t('orders_package.cancel_requires_undispatch'));
```
namrataukirde: Remove semicolon here. 😉

spec/controllers/api/v1/orders_packages_controller_spec.rb
```diff
@@ -186,7 +186,7 @@
           it 'fails to dispatch the packages' do
             expect(current_state).to eq('designated')
 
-            put :exec_action, id: orders_package.id, action_name: 'dispatch'
+            put :exec_action, id: orders_package.id, action_name: 'dispatch', quantity: 10, location_id: package.locations.first
```
namrataukirde: Are we passing a location object for location_id? 🤔

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -8,8 +8,9 @@
 
   let(:users) { create_list(:user, 2) }
 
-  let(:charity_users) { (1..26).map { create(:user, :with_multiple_roles_and_permissions,
-    roles_and_permissions: { 'Charity' => ['can_login_to_browse']}, last_name: 'Jones')}}
+  let(:charity_users) { ('a'..'z').map { |i|
+    create(:user, :with_multiple_roles_and_permissions,
+    roles_and_permissions: { 'Charity' => ['can_login_to_browse']}, first_name: "Jane_#{i}", last_name: 'Doe')}}
```
namrataukirde: 👍

spec/controllers/api/v1/orders_packages_controller_spec.rb
patrixr: should be the id, I guess rails is being smart

PR #906, opened 2020-02-06
Inventorize operations

patrixr:
This is the first part of [GCW-3025](https://jira.crossroads.org.hk/browse/GCW-3025)

**Note:** this PR adds internal features and tests but those are not used as of now

#### Inventorize/Uninventorize actions

Those will be used via the admin app to record the initial quantity of a package

file : `models/concerns/operations/stock_operations.rb`

Usage:

```
Package::Operations.inventorize(package, location)
Package::Operations.uninventorize(package)
```

A package can only be uninventorized right after being inventorized. If it is modified in any way, the operation will be rejected

#### PackagesInventory > undo

Added an `undo` method to the PackagesInventory method, allowing to revert an action by adding it's opposite action to the log

e.g

```
last_action = PackagesInventory.find_by(package_id: 1, action: 'inventory')
last_action.undo
puts(PackagesInventory.last) # { action: 'uninventory', package_id: 1, ... }
```

Note: not all types of actions can be undone

### Fixes

https://jira.crossroads.org.hk/browse/GCW-3025

An exception class had be removed, I suspect through a git conflict. (I miss static typing)

### DIFF
--- a/app/models/concerns/operations/stock_operations.rb
+++ b/app/models/concerns/operations/stock_operations.rb
@@ -16,7 +16,7 @@ module StockOperations
     # @param [Location|String] the location to place the package in
     #
     def inventorize(package, location)
-      last = PackagesInventory.order('id DESC').where(package: package).first
+      last = PackagesInventory.order('id DESC').where(package: package).limit(1).first
 
       raise Goodcity::AlreadyInventorizedError if last.present? && !last.uninventory?
 
@@ -37,7 +37,7 @@ module StockOperations
     # @param [Location|String] the location to place the package in
     #
     def uninventorize(package)
-      last_action = PackagesInventory.order('id DESC').where(package: package).first
+      last_action = PackagesInventory.order('id DESC').where(package: package).limit(1).first
       raise Goodcity::UninventoryError if last_action.blank? || !last_action.inventory?
       last_action.undo
     end
--- a/app/models/concerns/operations/stock_operations.rb
+++ b/app/models/concerns/operations/stock_operations.rb
@@ -5,6 +5,43 @@ module StockOperations
 
     module_function
 
+
+    ##
+    # Adds a package to the inventory
+    #
+    #
+    # @raise [Goodcity::AlreadyInventorizedError] thrown if trying to inventorize twice
+    #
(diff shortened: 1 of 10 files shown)


app/models/concerns/operations/stock_operations.rb
```diff
@@ -5,6 +5,43 @@ module StockOperations
...
+    # Adds a package to the inventory
+    #
+    #
+    # @raise [Goodcity::AlreadyInventorizedError] thrown if trying to inventorize twice
+    #
+    # @param [Package|String] the package to inventorize or its id
+    # @param [Location|String] the location to place the package in
+    #
+    def inventorize(package, location)
+      last = PackagesInventory.order('id DESC').where(package: package).first
```
namrataukirde: @patrixr, Just a suggestion: 

`PackagesInventory.order('id DESC').where(package: package)` In this, it will order all table records first then apply where query. Eventually, we will have large number of records inside this table. So I feel first applying where query and then doing order will be better.

`PackagesInventory..where(package: package).order('id DESC').first`

Let me know your thoughts. :)

patrixr: I don't actually think it makes a difference.
ActiveRecord compiles it into SQL right before running it so it wouldn't change anything. You can check that by doing `to_sql` on the query. It will return the same one

Although it's true that it could be slow **IF** `first` is an array method rather than a relation method (actually likely now that I think about it), in which case I think `limit(1)` is probably the optimisation we need

steveyken: When chaining AREL queries, it generally doesn't matter what order you write the clauses in as it will create an abstract syntax tree and combine clauses where necessary. So it's fine for `order` to come before `where` if desired. Just did a test with `first` and confirmed that just does a `limit(1)` and `order('id')` and then returns an instance of the the underlying model class. See the SQL below, which has `ORDER BY` and `LIMIT`

>> Order.first
  Order Load (1.4ms)  SELECT  "orders".* FROM "orders"  ORDER BY "orders"."id" ASC LIMIT 1
 => #<Order id: 1, status: "closed"...>

PR #911, opened 2020-02-13
Gcw 2982 import designations from stockit to api

namrataukirde:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2982

### What does this PR do?

It imports all designations from stockit to API as per mapping mentioned on ticket: https://jira.crossroads.org.hk/browse/GCW-2945

Please review.

Thanks.


lib/tasks/import_designations.rake
```diff
@@ -29,13 +31,31 @@ namespace :stockit do
...
       end
 
-      order.status = value["status"]
+      if ["CarryOut", "Shipment"].include?(value["detail_type"])
+        order.state = Order::SHIPMENT_STATUS_MAP[value["status"]]
+      elsif value["detail_type"] == "LocalOrder"
+        order.state = Order::LOCAL_ORDER_STATUS_MAP[value["status"]]
+      else
+        order.state = value["status"].try(:downcase)
+      end
```
swatijadhav:
```ruby
order.state = case value["detail_type"]
    when "CarryOut", "Shipment"
        Order::SHIPMENT_STATUS_MAP[value["status"]]
    when "LocalOrder"
        Order::LOCAL_ORDER_STATUS_MAP[value["status"]]
    else 
        value["status"].try(:downcase)
    end
```

namrataukirde: Sure, I will do a suggested change. Thanks. :)

lib/tasks/import_designations.rake
```diff
@@ -29,13 +31,32 @@ namespace :stockit do
...
+                      Order::LOCAL_ORDER_STATUS_MAP[value["status"]]
+                    else
+                      value["status"].try(:downcase)
+                    end
+
       order.code = value["code"]
       order.detail_type = value["detail_type"]
       order.created_at = value["created_at"]
-      order.description = value["description"]
+      order.updated_at = value["updated_at"] if value["updated_at"].present?
```
steveyken: Do we need `Time.parse` here? and on other dates?

lib/tasks/import_designations.rake
```diff
@@ -29,13 +31,32 @@ namespace :stockit do
...
+      order.people_helped = value["number_of_people_helped"]
+      order.continuous = value["continuous"]
```
steveyken: Do we need to cooerce to a boolean?

lib/tasks/import_designations.rake
```diff
@@ -29,13 +31,32 @@ namespace :stockit do
...
+      order.shipment_date = value["sent_on"]
+      order.staff_note = value["comments"]
+      order.purpose_description = value["description"]
+
+      %w[process_completed_by_id cancelled_by_id closed_by_id dispatch_started_by_id
```
steveyken: technically not every order imported can be cancelled and closed. I wonder if we should omit these `_by_id` fields instead so that they are null. I think that would be fine as they make no sense in a post-stockit system. So could you remove this block of code?

db/schema.rb
```diff
@@ -11,21 +11,19 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20200207120244) do
+ActiveRecord::Schema.define(version: 20200207141946) do
```
steveyken: There are a lot of changes here. I wonder if you should do clean db migration and see if you get a different result. Looks like stuff from another branch.

lib/tasks/import_designations.rake
namrataukirde: Nope. I verified the dates. It works fine considering timezone as well.

lib/tasks/import_designations.rake
namrataukirde: It converts 'true' to true automatically as its boolean column.

db/schema.rb
namrataukirde: Yes, I will run migration with clean db and update it. Thanks. :)

lib/tasks/import_designations.rake
namrataukirde: Sure, Makes Sense. :+1:

PR #914, opened 2020-02-18
Gcw 2863-Support packages to offer

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2863

### What does this PR do?

FEATURE: Associates Offers to package.

### Mockup Link

https://xd.adobe.com/view/224f7314-f1c0-40b6-7d7a-929e616be6bb-547e/screen/c800238a-5200-4a8a-adea-39f2303e5832/Create-Item-Main-Form?x_product=xd-jiraserver/1.0

### DIFF
--- a/app/controllers/api/v1/offers_controller.rb
+++ b/app/controllers/api/v1/offers_controller.rb
@@ -166,7 +166,8 @@ module Api
         ActiveModel::ArraySerializer.new(records,
           each_serializer: summary_serializer,
           root: "offers",
-          include_messages: params[:include_messages] == "true"
+          include_messages: params[:include_messages] == "true",
+          restrict_payload_for_offers_package: bool_param(:for_offers_package, false),
         ).as_json
       end
 
--- a/app/controllers/api/v1/offers_controller.rb
+++ b/app/controllers/api/v1/offers_controller.rb
@@ -74,7 +74,7 @@ module Api
         records = records.search({ search_text: params['searchText'], states: array_param(:state) }) if params['searchText'].present?
         records = apply_filters(records)
         records = records.page(params["page"]).per(params["per_page"] || params["recent_offer_count"] || DEFAULT_SEARCH_COUNT)
-        offers = offer_summary_response(records.with_summary_eager_load)
+        offers  = offer_summary_response(records)
         render json: {meta: {total_pages: records.total_pages, search: params['searchText']}}.merge(offers)
       end
 
@@ -163,11 +163,12 @@ module Api
       end
 
       def offer_summary_response(records)
+        records = params["shallow"]== 'true' ? records : records.with_summary_eager_load
+        serializer = params[:shallow] == 'true' ? shallow_summary_serializer : summary_serializer
         ActiveModel::ArraySerializer.new(records,
-          each_serializer: summary_serializer,
+          each_serializer: serializer,
           root: "offers",
-          include_messages: params[:include_messages] == "true",
-          restrict_payload_for_offers_package: bool_param(:for_offers_package, false),
+          include_messages: params[:include_messages] == "true"
         ).as_json
       end
 
@@ -213,6 +214,10 @@ module Api
 
       def summary_serializer
         Api::V1::OfferSummarySerializer
+      end
+
+      def shallow_summary_serializer
+        Api::V1::OfferShallowSummarySerializer
       end
 
       def select_serializer
--- a/app/controllers/api/v1/offers_controller.rb
+++ b/app/controllers/api/v1/offers_controller.rb
@@ -163,8 +163,8 @@ module Api
       end
 
       def offer_summary_response(records)
-        records = params["shallow"]== 'true' ? records : records.with_summary_eager_load
-        serializer = params[:shallow] == 'true' ? shallow_summary_serializer : summary_serializer
+        records = params[:companies]== 'true' ? records : records.with_summary_eager_load
+        serializer = params[:companies] == 'true' ? offer_company_serializer : summary_serializer
         ActiveModel::ArraySerializer.new(records,
           each_serializer: serializer,
           root: "offers",
@@ -216,8 +216,8 @@ module Api
         Api::V1::OfferSummarySerializer
       end
 
-      def shallow_summary_serializer
-        Api::V1::OfferShallowSummarySerializer
+      def offer_company_serializer
+        Api::V1::OfferCompanySerializer
       end
 
       def select_serializer
--- a/app/controllers/api/v1/offers_packages_controller.rb
+++ b/app/controllers/api/v1/offers_packages_controller.rb
@@ -23,7 +23,7 @@ module Api
         if @offers_package
           @offers_package.destroy
         end
-        render json: {}
+        render nothing: true, status: 204
       end
 
       private
--- a/app/controllers/api/v1/offers_packages_controller.rb
+++ b/app/controllers/api/v1/offers_packages_controller.rb
@@ -11,8 +11,8 @@ module Api
         error 500, "Internal Server Error"
       end
 
-      def_param_group :orders_packages do
-        param :orders_packages, Hash, required: true do
+      def_param_group :offers_packages do
+        param :offers_packages, Hash, required: true do
           param :package_id, Integer, desc: "Id of package"
           param :offer_id, Integer, desc: "Id of offer"
         end
@@ -28,7 +28,7 @@ module Api
 
       private
 
-      def orders_packages_params
+      def offers_packages_params
         params.require(:offers_packages).permit(:package_id, :offer_id)
       end
     end
--- a/app/controllers/api/v1/offers_packages_controller.rb
+++ b/app/controllers/api/v1/offers_packages_controller.rb
@@ -1,0 +1,36 @@
+module Api
+  module V1
+    class OffersPackagesController < Api::V1::ApiController
+      load_and_authorize_resource :offers_package, parent: false
+
+      resource_description do
+        formats ['json']
+        error 401, "Unauthorized"
+        error 404, "Not Found"
+        error 422, "Validation Error"
+        error 500, "Internal Server Error"
+      end
+
+      def_param_group :orders_packages do
+        param :orders_packages, Hash, required: true do
+          param :package_id, Integer, desc: "Id of package"
+          param :offer_id, Integer, desc: "Id of offer"
+        end
+      end
+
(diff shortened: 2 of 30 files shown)


app/serializers/api/v1/offer_summary_serializer.rb
```diff
@@ -83,5 +99,18 @@ def goodcity_user?
       User.current_user.present?
     end
 
+    def restrict_payload_for_offers_package?
+      !options[:restrict_payload_for_offers_package]
```
patrixr: Doesn't make sense to me that `restrict_payload_for_offers_package?` would equal to the negative `(NOT) !options[:restrict_payload_for_offers_package]`

app/serializers/api/v1/offers_package_serializer.rb
```diff
@@ -0,0 +1,22 @@
+module Api::V1
+  class OffersPackageSerializer < ApplicationSerializer
+    embed :ids, include: true
+    attributes :id, :offer_id, :item_id
```
patrixr: Should just be `package_id` and not `item_id`

Renaming on the backend for the sake of the Stock app cannot happen, ever.
Stock's mistakes should be handled in Stock, we don't bend the rules of the server for one app

:)

app/serializers/api/v1/offer_summary_serializer.rb
```diff
@@ -83,5 +99,18 @@ def goodcity_user?
...
+
+    %w[include_state? include_created_at? include_inactive_at?
+      include_updated_at? include_submitted_at? include_reviewed_at? include_review_completed_at?
+      include_received_at? include_cancelled_at? include_start_receiving_at?
+      include_submitted_items_count? include_accepted_items_count? include_rejected_items_count?
+      include_expecting_packages_count? include_missing_packages_count? include_received_packages_count?
+      include_display_image_cloudinary_id? include_notes? include_closed_by? include_reviewed_by?
+      include_received_by? include_delivery?].each do |method|
+        alias_method method.to_sym, :restrict_payload_for_offers_package?
+      end
```
patrixr: I'm assuming you're trying to exclude those that you do not need, but those seem to be almost ALL of them.
I think at this point we could consider not using a serializer at all

app/serializers/api/v1/offer_summary_serializer.rb
```diff
@@ -7,7 +7,7 @@ class OfferSummarySerializer < ApplicationSerializer
       :received_at, :cancelled_at, :start_receiving_at,
       :submitted_items_count, :accepted_items_count, :rejected_items_count,
       :expecting_packages_count, :missing_packages_count, :received_packages_count,
-      :display_image_cloudinary_id, :notes
+      :display_image_cloudinary_id, :notes, :inventoried_package_count, :unrecorded_package_count
```
patrixr: How are those fields used ? 🤔

patrixr:
@abulsayyad123 

I'm a bit confused, why is there so much serializer re-wiring just to get a list of ids 🤔 

Couldn't we simply have gotten away with `GET /api/v1/offers_packages?for_package=123` ?

app/serializers/api/v1/offer_summary_serializer.rb
abulsayyad123: On Offer search we have to show these count beside offer name.

app/serializers/api/v1/offer_summary_serializer.rb
```diff
@@ -22,6 +22,22 @@ def include_messages?
       @options[:include_messages] == true
     end
 
+    def inventoried_package_count
+      object.inventory_packages.size
+    end
+
+    def inventoried_package_count__sql
+      "SELECT COUNT(*) FROM packages INNER JOIN items ON items.id = packages.item_id INNER JOIN offers ON offers.id = items.offer_id WHERE items.offer_id = 729 and inventory_number is not null"
```
patrixr: There's a hardcoded `items.offer_id = 729`

abulsayyad123: Ahh.. I missed this while doing testing. That's embarrassing 🤦‍♂ .

app/serializers/api/v1/offer_summary_serializer.rb
steveyken: As discussed in slack, abul will remove this as it was an unexpected artifact from the mockup

app/models/concerns/offer_search.rb
```diff
@@ -13,13 +13,13 @@ module OfferSearch
 
     scope :search, -> (options = {}) {
       search_text = options[:search_text] || ''
-      search_query = ['offers.notes', 'users.first_name', 'users.last_name',
+      search_query = ['offers.notes', 'offers.id', 'users.first_name', 'users.last_name',
          'users.email', 'users.mobile', 'items.donor_description',
-         'messages.body', 'companies.name',
+         'messages.body', 'companies.name', 'packages.inventory_number',
```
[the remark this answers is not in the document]
abulsayyad123: Sure, will add one.

app/serializers/api/v1/offer_shallow_summary_serializer.rb
```diff
@@ -0,0 +1,10 @@
+module Api::V1
+  class OfferShallowSummarySerializer < ApplicationSerializer
```
steveyken: Can you call this `OfferCompanySerializer`?

app/controllers/api/v1/offers_controller.rb
```diff
@@ -215,6 +216,10 @@ def summary_serializer
         Api::V1::OfferSummarySerializer
       end
 
+      def shallow_summary_serializer
+        Api::V1::OfferShallowSummarySerializer
```
steveyken: Let's call this the `OfferCompanySerializer` as it's closer to what it means.

app/controllers/api/v1/offers_controller.rb
```diff
@@ -163,11 +163,12 @@ def filter_created_by(offers)
       end
 
       def offer_summary_response(records)
+        records = params["shallow"]== 'true' ? records : records.with_summary_eager_load
+        serializer = params[:shallow] == 'true' ? shallow_summary_serializer : summary_serializer
```
steveyken: Could you switch the logic around and replace `params['shallow']` with params['companies']. The default should be the summary_serializer and the offer_company_serializer can be used if params['companies'] == 'true'
Replace `shallow_summary_serializer` with `offer_company_serializer` to be more explicit about what we are returning. Makes it much clearer.

app/models/ability.rb
```diff
@@ -18,7 +18,7 @@ class Ability
     can_destroy_contacts can_read_or_modify_user can_handle_gogovan_order
     can_read_schedule can_destroy_image can_destroy_package_with_specific_states
     can_manage_locations can_read_versions can_create_goodcity_requests
-    can_manage_settings can_manage_companies can_manage_package_detail can_access_printers can_manage_offers_packages
+    can_manage_settings can_manage_companies can_manage_package_detail can_access_printers can_remove_offers_packages
```
steveyken: 👍 much clearer!

app/serializers/api/v1/offers_package_serializer.rb
```diff
@@ -0,0 +1,23 @@
...
+    embed :ids, include: true
+    attributes :id, :offer_id, :package_id
+    has_one :offer, serializer: OfferShallowSummarySerializer
+
+    def offer_id
+      object.offer_id
+    end
+
+    def offer_id__sql
+      "offer_id"
```
swatijadhav: @abulsayyad123 Can you please explain why we need to explicitly specify `offer_id__sql` and `package_id__sql`? as those are the columns of  table 

https://github.com/crossroads/api.goodcity/blob/62ae2179c4f52957581fcbf94fb8319ba1bf8e25/db/schema.rb#L422-L425

abulsayyad123: Yes.. we can remove this. At first I was renaming `package` to `item` that is why I had added this `__sql` methods.

db/permissions_roles.yml
```diff
@@ -36,6 +36,7 @@ Supervisor:
 - can_manage_deliveries
 - can_manage_orders
 - can_manage_orders_packages
+- can_manage_offers_packages
```
namrataukirde: I think you need to replace this with `can_remove_offers_packages.`

app/controllers/api/v1/offers_packages_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+          param :offer_id, Integer, desc: "Id of offer"
+        end
+      end
+
+      api :DELETE, '/v1/offers_package/1', "Delete an offers_package"
+      def destroy
+        if @offers_package
+          @offers_package.destroy
+        end
+        render json: {}
```
swatijadhav: Instead of passing empty json use `render nothing: true, status: 204 ` https://github.com/crossroads/api.goodcity/blob/e0cf56d732426b13416f791b3b3e71053e1ab47b/app/controllers/api/v1/images_controller.rb#L75

abulsayyad123: `nothing: true` is something new for me :) . Thanks

spec/serializers/api/v1/offer_company_serializer_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  let(:user) { create :user }
+  let(:offer)      { create(:offer, :with_items, company: company, created_by: user) }
+  let(:serializer) { Api::V1::OfferCompanySerializer.new(offer, root: 'offer').as_json }
+
+  subject { JSON.parse( serializer.to_json ) }
+
+  it 'creates json' do
+    expect(subject['offer']['id']).to eq(offer.id)
+    expect(subject['offer']['company_id']).to eq(offer.company.id)
+    expect(subject['offer']['created_by_id']).to eq(offer.created_by.id)
```
swatijadhav: Can we try this `json-matcher` https://relishapp.com/waterlink/rspec-json-expectations/docs/json-expectations/include-json-matcher ?

abulsayyad123: It's not working.. I think it's version issue.

steveyken: This is a nice idea. For time-being, if it's not working, I usually keep the data as a Hash using `as_json` and then you can use standard rspec matchers.

PR #916, opened 2020-02-19
GCW-2985 update package serializer to send on-hand-quantity

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-2985

### What does this PR do?
update package serializer to send on-hand-quantity

### Impacted Areas
Package serializer

### DIFF
--- a/app/serializers/api/v1/package_serializer.rb
+++ b/app/serializers/api/v1/package_serializer.rb
@@ -13,7 +13,7 @@ module Api::V1
                :item_id, :state, :received_at, :rejected_at, :inventory_number,
                :created_at, :updated_at, :package_type_id, :designation_id, :sent_on,
                :offer_id, :designation_name, :grade, :donor_condition_id, :received_quantity,
-               :allow_web_publish, :detail_type, :detail_id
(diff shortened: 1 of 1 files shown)


app/serializers/api/v1/package_serializer.rb
```diff
@@ -39,6 +39,14 @@ def sent_on__sql
       "stockit_sent_on"
     end
 
+    def on_hand_quantity
+      object.total_in_hand_quantity
+    end
+
+    def on_hand_quantity__sql
+      "(SELECT sum(quantity) FROM packages_inventories where packages_inventories.package_id = packages.id)"
```
abhinav098: @swatijadhav I think it should be `packages_inventories.package_id = #{object.id}`

swatijadhav: Verified it on local, and it seems to be working fine. `object` is used mostly used in the related method definition. Ideally `__sql` should not `object` reference.

abhinav098: sure, makes sense 👍

PR #918, opened 2020-02-20
Coronavirus lock appointments

patrixr:
Based on the setting `api.appointments.prevent_booking_until` , all appointment slots are marked as locked before that specific date.

This fix is made to solve the issue of blocking appointments during the Coronavirus phase

### DIFF
--- a/app/models/appointment_slot.rb
+++ b/app/models/appointment_slot.rb
@@ -30,12 +30,30 @@ class AppointmentSlot < ActiveRecord::Base
     val
   end
 
-  def self.calendar(from, to)
+  def self.active_range(booking_type = nil)
+    end_date = Date::Infinity.new
+
+    start_date = begin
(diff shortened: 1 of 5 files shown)


app/models/appointment_slot.rb
```diff
@@ -51,7 +68,7 @@ def self.for_date(date)
 
     # Generate slots based on preset if no special slot have been specified for that date
     AppointmentSlotPreset
-      .where(day: date.wday)
+      .where(day: date.wday + 1)
```
swatijadhav: Will this change affect any existing feature?

patrixr: Ah good catch ! I was halfway on fixing a bug : )

It's a bug that got revealed by the tests (`wdate` is 0-indexed as opposed to our code which is 1-indexed). So Sundays are missed (which has no consequence, but still good to fix). Pushing the fix now

PR #920, opened 2020-02-24
GCW-3044 check on API for dispatched item before pack/unpack operations

abhinav098:
### Ticket Link: 

NOTE: https://jira.crossroads.org.hk/browse/GCW-3044

### What does this PR do?

BUG:  https://jira.crossroads.org.hk/browse/GCW-3044?focusedCommentId=37308&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-37308

### Impacted areas
Box/pallet item addition/removal screen.

### DIFF
--- a/app/models/concerns/operations/stock_operations.rb
+++ b/app/models/concerns/operations/stock_operations.rb
@@ -85,8 +85,8 @@ module StockOperations
 
       def pack
         return error(I18n.t("box_pallet.errors.adding_box_to_box")) if adding_box_to_a_box?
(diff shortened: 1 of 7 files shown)


app/models/concerns/operations/stock_operations.rb
```diff
@@ -123,6 +126,10 @@ def error(error)
         { errors: [error], success: false }
       end
 
+      def box_dispatched?
```
steveyken: As discussed on slack with @abhinav098 , we'll change this to check for `available_quantity > 0` - this is the amount that is left over once a package has been designated or dispatched. The Inventory computer can supply this value direct

PR #924, opened 2020-02-25
Order state transition validations

patrixr:
This covers a few regressions Matt noticed

- Cancelling an order with dispatched orders_packages should not be possible
- Dispatching an orders_package of an `awaiting_dispatch` order should put the order in the `dispatching` state

### DIFF
--- a/app/models/concerns/operations/order_fulfilment_operations.rb
+++ b/app/models/concerns/operations/order_fulfilment_operations.rb
@@ -82,6 +82,9 @@ module OrderFulfilmentOperations
           source: ord_pkg,
           location: location
         )
+
+        ord_pkg.order.start_dispatching if ord_pkg.order.awaiting_dispatch?
+
         unless ord_pkg.dispatched? || dispatched_count(ord_pkg) < ord_pkg.quantity
           ord_pkg.dispatch
           ord_pkg.package.dispatch_stockit_item(ord_pkg) if STOCKIT_ENABLED
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -225,7 +225,7 @@ class Order < ActiveRecord::Base
     end
 
     before_transition on: :cancel do |order|
-      if OrdersPackage.where(order: order).any? { |op| op.dispatched? || op.dispatched_quantity.positive? }
+      if OrdersPackage.for_order(order.id).not_cancellable.exists?
         raise Goodcity::InvalidStateError.new(I18n.t('orders_package.cancel_requires_undispatch'))
       end
 
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -225,6 +225,10 @@ class Order < ActiveRecord::Base
     end
 
     before_transition on: :cancel do |order|
+      if OrdersPackage.where(order: order).any? { |op| op.dispatched? || op.dispatched_quantity.positive? }
+        raise Goodcity::InvalidStateError.new(I18n.t('orders_package.cancel_requires_undispatch'))
+      end
+
       order.cancelled_at = Time.now
       order.cancelled_by = User.current_user
       order.orders_packages.each do |orders_package|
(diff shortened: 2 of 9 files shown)


app/models/order.rb
```diff
@@ -225,6 +225,10 @@ def set_initial_state
     end
 
     before_transition on: :cancel do |order|
+      if OrdersPackage.where(order: order).any? { |op| op.dispatched? || op.dispatched_quantity.positive? }
+        raise Goodcity::InvalidStateError.new(I18n.t('orders_package.cancel_requires_undispatch'))
+      end
```
steveyken: Potential option 

`order.orders_package.where("state = 'dispatched' OR dispatched_quantity > 0").any?`

namrataukirde: Agree with @steveyken and more on same: 

order.orders_package.where("state = 'dispatched' OR dispatched_quantity > 0").exists? will be better in some cases where we have large number of data.

Here we might not need exists but thought to share it.

https://www.ombulabs.com/blog/benchmark/performance/rails/present-vs-any-vs-exists.html

features/support/env.rb
```diff
@@ -40,6 +40,11 @@
 
 World(FactoryBot::Syntax::Methods)
 
+Before do
+  User.current_user ||= FactoryBot.create(:user, :supervisor)
```
namrataukirde: What if we want to test features specific to some other role where we want to write some kind of negative test cases about what specific role cant do.

A negative test case is just an example. I think we need to override it in test cases where we want to use it.

Just adding this comment so I will understand it more.

app/models/concerns/operations/order_fulfilment_operations.rb
```diff
@@ -82,6 +82,9 @@ def dispatch(ord_pkg, quantity:, from_location:)
           source: ord_pkg,
           location: location
         )
+
+        ord_pkg.order.start_dispatching if ord_pkg.order.awaiting_dispatch?
```
namrataukirde: Why we are adding a check `ord_pkg.order.awaiting_dispatch?`.

state_machine handles it on its own. If we try to perform operation which is not allowed from specific states then its throws error.

As per my understanding order can be transitioned to `dispatching` state only from `closed` and `awaiting_dispatched` state.

Do we want to avoid a transition from closed to dispatching state here? Because we cant dispatch items to closed order I think.

Or maybe just added for safety purposes. 

Could you please help me understand? 

Thanks in advance. 😉

patrixr: We only want to start processing if it's in `awaiting_dispatch` state.

Even if the state_machine handles bad cases, I don't rely on it.

Plus the extra `if` adds context to the statement. By reading it we understand that we *can* dispatch an orders_package from a scheduled order.

Even if you _can_ remove a statement, doesn't always mean you _should_

features/support/env.rb
patrixr: Still haven't figured out how to do per-feature Before statements

app/models/order.rb
patrixr: Will update it, although that's just one node of a greater web of potential issues.

Just a few lines below you'll notice `order.orders_packages.each` which has the same potential issues.

If you suspect performance issues I recommend figuring out a general approach to handling them (we can create a ticket for that). Singular changes here and there won't matter enough to make an impact

features/step_definitions/order_state_transitions_steps.rb
```diff
@@ -0,0 +1,46 @@
...
+      location_id:    @orders_package.package.locations.first.id,
+      quantity:       @orders_package.package.received_quantity,
+    )
+  }.to (
+    expect_fail ? raise_error(Goodcity::OperationsError) : change(PackagesInventory, :count).by(1)
+  )
+end
+
+Then(/^The order transitions to the "([^"]*)" state/) do |expected_state|
+  expect(@order.reload.state).to eq(expected_state)
```
steveyken: nice!

PR #925, opened 2020-02-27
GCW-3051 fix duplicate record creation for booking type

namrataukirde:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3051

### What does this PR do?

- Added a unique index to the booking_types name_en column to prevent types with the same. name.
- Added uniq validation name_en and name_zh_tw column of booking type.

Please review.

Thanks.

### DIFF
--- a/spec/controllers/api/v1/orders_controller_spec.rb
+++ b/spec/controllers/api/v1/orders_controller_spec.rb
@@ -1,9 +1,10 @@
 require 'rails_helper'
 
 RSpec.describe Api::V1::OrdersController, type: :controller do
-  let(:booking_type) { create :booking_type }
+  let!(:appointment_type) { create(:booking_type, :appointment) }
+  let!(:online_order_type) { create(:booking_type, :online_order) }
   let(:charity_user) { create :user, :charity, :with_can_manage_orders_permission}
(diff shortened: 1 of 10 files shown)


spec/controllers/api/v1/orders_controller_spec.rb
```diff
@@ -529,9 +524,9 @@ def create_order_with_transport(state)
       end
 
       context 'Processing checklist' do
-        let!(:checklist_it1) { create :process_checklist, booking_type: booking_type }
-        let!(:checklist_it2) { create :process_checklist, booking_type: booking_type }
-        let!(:checklist_it3) { create :process_checklist, booking_type: booking_type }
+        let!(:checklist_it1) { create :process_checklist, booking_type: appointment_type }
+        let!(:checklist_it2) { create :process_checklist, booking_type: appointment_type }
+        let!(:checklist_it3) { create :process_checklist, booking_type: appointment_type }
```
swatijadhav: Can you please check if we can make use of `create_list` here?

namrataukirde: @swatijadhav we cannot use `create_list` here, as the assigned variables are used

PR #926, opened 2020-02-28
Remove quantity field

patrixr:
### Migration

- Removal of the `quantity` field of the `packages` table
- Addition of the following columns on `packages`
   * `available_quantity` (on-site non-designated quantity)
   * `on_hand_quantity` (total on-site quantity)
   * `designated_quantity` (on-site designated quantity)
   * `dispatched_quantity` (total dispatched quantity)

### Code changes

- `Redesignate` operation is now part of `DesignationOperations` concern
- Removal of a lot of unused code
- Removal of all the `quantity` field computations
- Computation of the field mentioned above from the inventory
- Temporary refactor of `packages_controller` to convert StockIt changes into `inventory` rows (Will be removed right after stockit shutdown)
- Fix of a lot of tests & code to adapt to the new quantity field changes (e.g PackageSplitter)
- Addition of `@TODO`s at places with Stockit related code which we'll need to remove after turning stockit off


app/models/concerns/package_filtering.rb
```diff
@@ -109,26 +107,22 @@ def filter_by_location(location_name)
...
-
     private
 
     def received_sql
       "packages.state = 'received'"
     end
 
     def in_stock_sql
-      "packages.state = 'received' AND packages.quantity > 0"
+      "packages.state = 'received' AND packages.available_quantity > 0"
```
steveyken: Do we need to check for `state='received'`?

spec/services/stockit/orders_package_sync_spec.rb
```diff
@@ -54,7 +53,7 @@
     it "sends update request and get success_response" do
       stockit_param_for_inventory = stockit_params.merge({update_gc_orders_package: true,
         orders_package_id: orders_package.id, orders_package_state: orders_package.state})
-      expect( Nestful ).to receive(:put).with( url, stockit_param_for_inventory, options ).and_return( mock_response )
+      expect( Nestful ).to receive(:put).with( url, stockit_param_for_inventory, options ).and_return({ "status" => 201 })
```
steveyken:
```suggestion
      expect( Nestful ).to receive(:put).with( url, stockit_param_for_inventory, options ).and_return(success_response)
```

spec/services/stockit/orders_package_sync_spec.rb
```diff
@@ -27,7 +26,7 @@
 
     it "sends create request and get success_response" do
       stockit_params_for_inventory_number_presence = stockit_params.merge({generate_q_inventory_number: true})
-      expect( Nestful ).to receive(:post).with( url, stockit_params_for_inventory_number_presence, options ).and_return( mock_response )
+      expect( Nestful ).to receive(:post).with( url, stockit_params_for_inventory_number_presence, options ).and_return({ "status" => 201 })
```
steveyken:
```suggestion
      expect( Nestful ).to receive(:post).with( url, stockit_params_for_inventory_number_presence, options ).and_return(success_response)
```

app/models/concerns/package_filtering.rb
patrixr: @steveyken I don't really know, I figured I would just leave it as it was

app/models/concerns/inventory_computer.rb
```diff
@@ -85,11 +93,13 @@ def initialize(model)
       end
 
       def compute
-        as_of(Time.now)
+        as_of(nil)
```
swatijadhav: Are we using this anywhere?

patrixr: Yes this is used to compute quantities _"as of"_ a certain time. I just modified it not to include a timestamp unless needed (default behaviour)

app/controllers/api/v1/packages_controller.rb
```diff
@@ -379,18 +408,68 @@ def package_record
         @package.storage_type = assign_storage_type
         @package.detail = assign_detail if params["package"]["detail_type"].present?
         @package.received_quantity ||= received_quantity
+        @package.offer_id = offer_id
         add_favourite_image if params["package"]["favourite_image_id"]
         @package
       end
 
+      def try_inventorize_package(pkg)
```
abulsayyad123: `@package` is available throughout I think there is no need to pass it.

app/controllers/api/v1/packages_controller.rb
```diff
@@ -129,6 +152,17 @@ def assign_donor_condition?
...
+        raise Goodcity::InventorizedPackageError
+      end
+
+      api :PUT, "/v1/packages/1", "Mark a package as missing"
+      def mark_missing
+        ActiveRecord::Base.transaction do
+          Package::Operations.uninventorize(@package) if PackagesInventory.inventorized?(@package)
+          @package.mark_missing
+        end
+        render json: serializer.new(@package).as_json
```
abulsayyad123: How transaction failing is handled ?

app/controllers/api/v1/packages_controller.rb
patrixr: I prefer to write methods this way. Makes them less dependant on state, which in turn makes them moveable, re-usable. As well as more readable since you see the parameter, therefore don't need to guess _what_ will be inventoried.

I'm not a fan of methods called with implicit context generally

app/controllers/api/v1/packages_controller.rb
patrixr: Operations raise Goodcity exceptions, which are caught by the controller

PR #927, opened 2020-02-28
#Release1 March 2020

abhinav098:
# #Release1  March2020

**This PR contains the following changes:**
- GCW-2986 add missing seed data
- Inventorize operations
- GCW-2905-stockit-enabledDon't add inventory to stockit if it's disabled.
- GCW-3035-fix-uninventorise-items-visible-in-stock-app
- GCW-2931-request-cancellation-for-order
- GCW-2859-add-items-to-box-pallet
- GCW-2985 update package serializer to send on-hand-quantity
- Coronavirus lock appointments
- GCW-3044-disable-pack-operations-for-dispatched-box
- GCW-2982-import-designations-from-stockit-to-api
- Order state transition validations
- GCW-2863-Support linking of packages to offers

Rake task list:
 **rake goodcity:add_new_cancellation_reasons**


app/serializers/api/v1/package_serializer.rb
```diff
@@ -39,6 +39,14 @@ def sent_on__sql
       "stockit_sent_on"
     end
 
+    def available_quantity
+      object.total_available_quantity
+    end
+
+    def available_quantity__sql
+      "(packages.received_quantity - ((SELECT COALESCE(sum(quantity), 0) FROM orders_packages WHERE orders_packages.state = 'designated' AND orders_packages.package_id = packages.id) - (SELECT COALESCE(sum(dispatched_quantity), 0) FROM orders_packages WHERE orders_packages.state = 'designated' AND orders_packages.package_id = packages.id)))"
```
steveyken: Can I confirm that the equation should be `available_quantity = received_quantity - designated_quantity - dispatched_quantity`? If so, the SQL isn't quite right here... we need:

- quantity from designated orders_packages
- dispatched_quantity from dispatched orders_packages

```suggestion
      "(packages.received_quantity - ((SELECT COALESCE(sum(quantity), 0) FROM orders_packages WHERE orders_packages.state = 'designated' AND orders_packages.package_id = packages.id) - (SELECT COALESCE(sum(dispatched_quantity), 0) FROM orders_packages WHERE orders_packages.state = 'dispatched' AND orders_packages.package_id = packages.id)))"
```

app/models/cancellation_reason.rb
```diff
@@ -15,4 +19,9 @@ def self.unwanted
   def self.donor_cancelled
     find_by(name_en: "Donor cancelled")
   end
+
+  def self.cancellation_reasons_for(type)
+    return where unless CANCELLATION_REASONS_TYPE.include?(type)
```
steveyken:
```suggestion
    return where(nil) unless CANCELLATION_REASONS_TYPE.include?(type)
```

app/controllers/api/v1/packages_controller.rb
```diff
@@ -225,6 +228,39 @@ def send_stock_item_response
...
+                  )
+        if response[:success]
+          render json: { packages_inventories: response[:packages_inventory] }, status: 201
+        else
+          render json: { errors: response[:errors] }, status: 422
+        end
+      end
+
+      def contained_packages
+        entity = Package.find(params[:id])
```
swatijadhav: Should it be `find_by(id: params[:id])`, as on next we are verifying if `entity` is present or not.

app/controllers/api/v1/packages_controller.rb
```diff
@@ -225,6 +228,39 @@ def send_stock_item_response
...
+
+        contained_pkgs = entity.associated_packages&.page(page)&.per(per_page)
+        render json: contained_pkgs, each_serializer: stock_serializer, include_items: true,
+          include_orders_packages: false, include_storage_type: false,
+          include_donor_conditions: false, exclude_stockit_set_item: true, root: "items"
+      end
+
+      def fetch_added_quantity
+        entity_id = params[:entity_id]
+        package = Package.find(params[:id])
```
swatijadhav: Same here, Should it be `find_by(id: params[:id])` ?

app/serializers/api/v1/package_serializer.rb
swatijadhav: This SQL is formed by referring following code 

```ruby
        query = OrdersPackage.designated.where(package: package)
        query = query.where(order: to_order) if to_order.present?
        query.reduce(0) { |sum, op| sum + op.quantity - op.dispatched_quantity }

```
https://github.com/crossroads/api.goodcity/blob/master/app/models/concerns/inventory_computer.rb#L44-L48

The dispatched quantity considered here is from the partial dispatch, and hence we are looking into `designated` state only. 

cc: @patrixr

PR #939, opened 2020-03-06
GCW-3036 support package actions

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3036

### What does this PR do?

Stock - Supporting follwoing actions:

- **loss**
- **trash**
- **recycle**
- **process** 

### Impacted Areas

### Screenshots

### Mockup Link

Note: Specify mockup link

### Linked PR's:


app/models/concerns/operations/stock_operations.rb
```diff
@@ -33,11 +33,10 @@ def inventorize(package, location)
     #
     # @raise [Goodcity::UninventoryError] thrown if actions were taken since the initial inventory action
     #
-    # @param [Package|String] the package to inventorize or its id
     # @param [Location|String] the location to place the package in
     #
-    def uninventorize(package)
-      last_action = PackagesInventory.order('id DESC').where(package: package).limit(1).first
+    def uninventorize(package_inventory_id)
+      last_action = PackagesInventory.find(package_inventory_id)
```
patrixr: @swatijadhav Could you explain this change ?

This looks wrong to me. A package can only be un-inventorized right after the inventory action. But that logic should remain encapsulated in the operations concern. This extracts it out and opens the doors for misuse

app/controllers/api/v1/packages_controller.rb
```diff
@@ -207,6 +208,19 @@ def designate
         send_stock_item_response
       end
 
+      api :PUT, "/v1/packages/1/actions/:action_name", "Executes an action on a package"
+      param_group :operations
+
+      def exec_action
+        Package::Operations.perform_action(@package,
```
patrixr: @swatijadhav This `perform_action` method would accept basically any action here, which means it's basically opening up the inventory to any writes, which is way too permissive.

For example, `dispatch`, `inventory`, `move` actions are reserved to some unique operations, they all come with their own validation/conditions. So those should not be allowed in the inventory unless we go through the proper entry points.

I would suggested locking down the entrypoint to something which represents `"pure"` quantity changes such as a `loss/recycled/...`. There's already `register_loss/register_gain` entry points which include the proper validation.

I've actually added some code in my quantity-related PR which could be useful

```
    ##
    # Registers either a gain or a loss action depending on the change value
    #
    # @param [Package] package the package that went missing
    # @param [Integer] quantity the quantity that was lost
    # @param [Location|Id] location the location to add the quantity to (or its id)
    #
    def register_quantity_change(package, delta:, location:)
      return if delta.zero?
      delta.positive? ?
        register_gain(package, quantity: delta, to_location: location) :
        register_loss(package, quantity: delta, from_location: location)
    end
```

swatijadhav: @patrixr 

In `stock_operations` concern, for every action check will be performed for valid incremental or decremental actions. refer https://github.com/crossroads/api.goodcity/pull/939/files#diff-1ec071393023822c7c83f0bae29e140aR72

Currently only added check for decremental actions. (later will add a check for other incremental actions) and for quantity, we are accepting only positive values and those will be updated further based on action type (either positive or negative type, I think you have already added changes for it, which am reusing.)

Should we add changes in that implementation and pass the `-/+` quantity value from the params?

app/models/concerns/operations/stock_operations.rb
swatijadhav: Added this change initially, while working on `undo` actions. But as discussed we are not covering undo actions in this PR, I will revert this change.

patrixr: Thanks :)

Note: even in the case of an undo feature, this change would be wrong. We can discuss a proper undo/redo design when the time comes

app/controllers/api/v1/packages_controller.rb
patrixr: @swatijadhav Yes, that is a good check for `loss/gain`, `trash/untras`, ...

But we still shouldnt allow `move`, `dispatch`, `inventory`.

The fact that some actions are only decremental or only incremental has nothing to do with the meaning of the action. e.g, `uninventory` is a decremental action, which would pass through those validation, except it should **never** be allowed outside of its own entry point because of all of the business logic associated with that action :)

swatijadhav: Agree, will update the check at https://github.com/crossroads/api.goodcity/pull/939/files#diff-1ec071393023822c7c83f0bae29e140aR72 and will only include the actions which will be allowed for quantity change and not other actions.

patrixr: Note, just in case it isn't clear, the `quantity` param is always a delta, meaning the amount that changed. And not the resulting quantity

patrixr: Thanks :)
Maybe we can define those "pure" actions somewhere. Meaning the simple ones that don't have any business logic attached to them. Can't think of a good name though 🤔 

Note, all `gain actions` *must* check the following : 

`raise Goodcity::NotInventorizedError unless PackagesInventory.inventorized?(package)`

app/models/concerns/operations/stock_operations.rb
```diff
@@ -62,11 +62,25 @@ def register_loss(package, quantity:, from_location:)
...
+    def perform_action(package, quantity:, location_id:, action: 'loss', description: nil)
+      if PackagesInventory::QUANTITY_LOSS_ACTIONS.include?(action)
+        register_loss(
+          package,
+          quantity: quantity,
+          location_id: location_id,
+          action: action,
+          description: description)
+      else
+        raise Goodcity::ActionNotAllowedError.new
```
patrixr: Are we not allowing gains yet ? Don't think there's anything preventing us from doing so.

Also, what do you think of changing the name of the method to `register_quantity_change` , to be more specific on what it does

swatijadhav: For this ticket, we are only considering actions related to `loss`. The `gain` related actions and undo actions will be covered in another ticket.

so once we have gain related actions, will add `PackagesInventory::QUANTITY_GAIN_ACTIONS` and if falls under that then will call `register_gain` for it.

> Also, what do you think of changing the name of the method to register_quantity_change , to be more specific on what it does

yes, will rename it from `perform_action` to `register_quantity_change`.

patrixr: @swatijadhav Sounds good, was just wondering :)

app/serializers/api/v1/stockit_item_serializer.rb
```diff
@@ -42,7 +43,7 @@ def on_hand_quantity
     end
 
     def on_hand_quantity__sql
-      "select sum(quantity) from packages_inventories where package_id = #{object.id}"
+      "select sum(quantity) from packages_inventories where packages_inventories.package_id = packages.id"
```
steveyken: 👍

PR #940, opened 2020-03-06
Gcw 2945 shipment specific changes

namrataukirde:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2945

### What does this PR do?

It adds the required changes to deal with shipments :
- Send emails only for GoodCity Orders.
- Handle order processing workflow for shipments.
- Remove status column dependency.
 
Please review.

Thanks.


app/models/concerns/order_filtering.rb
```diff
@@ -56,11 +56,19 @@ def priority
     end
 
     def due_after(time)
-      where('order_transports.scheduled_at >= ?', time).join_order_transports
+      where("CASE WHEN orders.detail_type='GoodCity' THEN
+                order_transports.scheduled_at >= (?)
+            ELSE
+                shipment_date >= (?)
+            END", time, time.to_date).join_order_transports
```
steveyken: what happens if more than one detail_type is selected?

db/schema.rb
```diff
@@ -448,6 +448,7 @@
   add_index "order_transports", ["scheduled_at"], name: "index_order_transports_on_scheduled_at", using: :btree
 
   create_table "orders", force: :cascade do |t|
+    t.string   "status"
```
steveyken: aren't we removing this field? or later?

app/models/concerns/order_filtering.rb
patrixr: unrelated to comment above, but using heredoc `<<-SQL` would fit nicely here : )

db/schema.rb
namrataukirde: It's in another branch. I will raise PR for it later separately. :)

app/models/concerns/order_filtering.rb
namrataukirde: @steveyken, It works fine with more than one detail_type selected.

PR #948, opened 2020-03-19
GCW-3075 fixed issue with twilio inbound calls

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3075

### What does this PR do?
fixed issue with twilio inbound calls

### DIFF
--- a/app/controllers/concerns/twilio_config.rb
+++ b/app/controllers/concerns/twilio_config.rb
@@ -20,11 +20,19 @@ module TwilioConfig
   private
 
   def activity_sid(friendly_name)
-    task_router.activities.list(friendly_name: friendly_name).first.sid
+    twilio_client.taskrouter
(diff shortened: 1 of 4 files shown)


app/controllers/concerns/twilio_config.rb
```diff
@@ -80,7 +87,7 @@ def twilio_creds
   end
 
   def user(mobile = nil)
-    @user ||= User.find_by_mobile(mobile || params["From"])
+    @user = User.find_by(mobile: mobile.presence || params["From"])
```
steveyken: what does `find_by(nil)` and `find_by('')` do? Just want to make sure it returns no users!

swatijadhav: Ideally if there are user records with nil or empty values then it will return those records else it will return no record.

PR #950, opened 2020-03-20
GCW-2658 fixed issue at setting favourite image for package from admin app

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-2658

### What does this PR do?

Package images are set while it is added to an item. New copy of the original image is created and it is assigned to a package.

When favourite package image is updated, it was checking only for the images present for the packages and not looking into its associated item's images and hence was failing to set favourite image.

Also, favourite_image_id is not updated on the package and also not sent via a serializer.

Fixed all above issue in this PR.


app/models/package.rb
```diff
@@ -48,6 +48,7 @@ class Package < ActiveRecord::Base
   after_commit :update_set_item_id, on: :destroy
   before_save :assign_stockit_designated_by, if: :unless_dispatch_and_order_id_changed_with_request_from_stockit?
   before_save :assign_stockit_sent_by_and_designated_by, if: :dispatch_from_stockit?
+  after_save :set_favourite_image_id
```
patrixr: Wouldn't it be `before_save` ?

app/models/package.rb
```diff
@@ -141,6 +142,10 @@ def assign_stockit_designated_by
     end
   end
 
+  def set_favourite_image_id
+    self.update_column(:favourite_image_id, self.images.find_by(favourite: true)&.id)
+  end
```
patrixr: Is it needed to do this with every save, isn't the `update_favourite_image` entrypoint enough ?

PR #956, opened 2020-03-26
#Release March 26 2020 - (quantities)

patrixr:
Note: Will not merge until we're ready to do so

* GCW-2944 Quantity changes
* GCW-3057 Quantity Split fix
* GCW-3036 Supporting gain, loss, trash, recycle & process actions
* GCW-3051 Show uniq types on logistics order type
* GCW-3075 Incoming call from donor is not showing up on admin app
* GCW-2722 fixed issue with resubmission of offer added by admin
* GCW-3038 Offers created on Admin have problems when arranging transport

(Might be missing some ticket numbers)


app/models/concerns/subform_utilities.rb
```diff
@@ -18,6 +18,7 @@ def distinct_by_column(column)
 
   def create_on_stockit
     return if request_from_stockit?
+
```
steveyken: And the "Best Whitespace Commit of the Year Award" goes to...? :)

cc @patrixr @abulsayyad123 @swatijadhav @bharat619

PR #961, opened 2020-03-27
GCW-3093 Avoid creation of stock-items with invalid inventory numbers

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3093

### What does this PR do?
Makes the code column of inventory_numbers table as integer

BUG: https://rollbar.com/crossroads_it/api.goodcity/items/3165/ 

### Impacted Areas
Stock -> Item creation

https://crossroads-hk.slack.com/archives/GBJAFNVFB/p1584588449123700


bharat619:
@steveyken @patrixr 
 I put some more thought on it. Putting 999999 is not a good idea.
 It was my assumption that inventory number will be of length 6. I think stockit will get 
 alphanumeric codes and GC will get numeric codes. If stockit is removed in future then all the 
 inventory codes will be numeric. It will cross the 999999 limit in future and again we might need 
 to increase the limit.
 So lets not put any limit to it. Let the max value be the range of INT, so it has max value of 
 2,147,483,647
 On crossing this range, lets show error message for invalid inventory number. In my latest 
 commit, I have made that change. 
 Let me know your thoughts on this.

cc: @swatijadhav @abulsayyad123 @shreyas098

steveyken:
@bharat619 I've reviewed the code as I was also working on inventory number fix. Going to take some changes from here and being them over to another branch to merge the changes. Will close this PR and reference from there.

steveyken:
New PR is https://github.com/crossroads/api.goodcity/pull/1013

PR #963, opened 2020-03-30
GCW-2658 avoid creation of duplicate package images

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-2658

### What does this PR do?
GCW-2658 avoid the creation of duplicate package images

### DIFF
--- a/app/models/package.rb
+++ b/app/models/package.rb
@@ -392,7 +392,8 @@ class Package < ActiveRecord::Base
       item_image = item.images.find_by(id: favourite_image_id)
 
       if item_image
-        current_image = Image.create(item_image.attributes.slice("cloudinary_id", "angle"))
+        current_image = images.find_by(cloudinary_id: item_image.cloudinary_id) ||
+          Image.create(item_image.attributes.slice("cloudinary_id", "angle"))
         self.images << current_image
       end
     end


app/models/package.rb
```diff
@@ -392,7 +392,8 @@ def set_favourite_image
       item_image = item.images.find_by(id: favourite_image_id)
 
       if item_image
-        current_image = Image.create(item_image.attributes.slice("cloudinary_id", "angle"))
+        current_image = images.find_by(cloudinary_id: item_image.cloudinary_id) ||
+          Image.create(item_image.attributes.slice("cloudinary_id", "angle"))
```
patrixr: Dumb question, but does that fill the polymorphic fields of the Image model (`item_id` / `item_type`). Since we don't pass them to `Image.create` ? 🤔 

And won't `self.images << current_image` cause the Package to have the same record twice in the `images` array if it's already there ?

swatijadhav: Here 
- first, it will verify if the `package` has similar image existing using Cloudinary URL so that image only be duplicated. `images.find_by(cloudinary_id: item_image.cloudinary_id)`
- If not then it will create a new image record, and later it will add it to packages using `self.images << current_image`

`self.images << current_image` will just form an association between package and the image record. it will not duplicate anything. if already exist it will not be affected.

steveyken: https://guides.rubyonrails.org/association_basics.html#scopes-for-has-many-distinct would suggest that `<<` does duplicate so worth checking this

swatijadhav: Thanks for the reference @steveyken 
Verified it on local rails console, by adding an existing associated image again to package using `package.images << image_record` and it does not duplicate record.

From the above reference, my understanding is, records will be duplicated when the records are added using `through` relationship, where it has an extra table which stores `through` relationship (primary keys of both tables) and there are the chances of creation of duplicate records.

And in our case we have direct association as 
`has_many   :images, as: :imageable, dependent: :destroy`

Please correct me in case if am missing anything. Thanks.

PR #965, opened 2020-04-01
Fix rollbar error for booking type

bharat619:
### Ticket Link:
https://jira.crossroads.org.hk/browse/GCW-3108
### Description
This fixes the occurrence of https://rollbar.com/crossroads_it/api.goodcity/items/663/

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -13,6 +13,8 @@ class Order < ActiveRecord::Base
     ]
   end
 
+  validates :booking_type_id, presence: true
+
   belongs_to :cancellation_reason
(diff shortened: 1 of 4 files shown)


app/models/order.rb
```diff
@@ -13,6 +13,8 @@ class Order < ActiveRecord::Base
     ]
   end
 
+  validates :booking_type_id, presence: true
```
steveyken, line 16: @bharat619 some orders (such as shipments) don't have a specific booking type so I'm not sure if we can require this field

cc @patrixr

bharat619: sure @steveyken ,
could you take a look now. Changes made as per our discussion

PR #976, opened 2020-04-16
GCW-3121 Add create order role for order-fulfilment users

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3121

### What does this PR do?
Gives permission for order fulfilment users to create new orders


bharat619:
@steveyken , the `can_read_or_modify_user` is required as when you click on `Create New Order`, the api `api/v1/recent_users` is fired, which requires this permission

steveyken:
ok that's fine thanks

PR #977, opened 2020-04-16
GCW-3120 Send order-related messages

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3120

### What does this PR do?
Changes to send messages related to orders.

### Linked PR's:
https://github.com/crossroads/stock.goodcity/pull/708

### DIFF
--- a/app/controllers/api/v1/orders_controller.rb
+++ b/app/controllers/api/v1/orders_controller.rb
@@ -118,7 +118,7 @@ module Api
           include_packages: true,
           include_order: false,
           include_images: true,
-          include_messages: params["include_messages"],
+          include_messages: bool_param(:include_messages, false),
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/orders_controller.rb
```diff
@@ -118,6 +118,7 @@ def order_response(records)
           include_packages: true,
           include_order: false,
           include_images: true,
+          include_messages: params["include_messages"],
```
patrixr, line 121:
```suggestion
          include_messages: bool_param(:include_messages, false),
```

The query params are all strings, so if `include_message` is a string which says "false" it will still evaluate to something truthy. (same thing below)

PR #979, opened 2020-04-17
GCW-3104 - Fetch package containers

patrixr:
## Changes

* Adds an endpoint to fetch the parent containers of a package
* Fix specs that are failing on Master
* Small refactor of existing box code
* Fix bugs that were hidden in specs

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -7,7 +7,7 @@ default: &default
 
 development:
   <<: *default
-  database: goodcity_server_development
+  database: goodcity_server_development_2
 
 test:
   <<: *default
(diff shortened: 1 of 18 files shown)


config/database.yml
```diff
@@ -7,7 +7,7 @@ default: &default
 
 development:
   <<: *default
-  database: goodcity_server_development
+  database: goodcity_server_development_2
```
swatijadhav, line 10: @patrixr Can you please rollback this change?

patrixr: already did

PR #992, opened 2020-04-24
GCW-3126 Fix Unable to delete beneficiary

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3126

### What does this PR do?
- Fixes the ability for charity user unable to delete the beneficiary
- Added rake task to delete the permissions

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -106,7 +106,7 @@ class Ability
 
   def beneficiary_abilities
     can :create, Beneficiary
-    can [:index, :show, :update, :delete], Beneficiary, created_by_id: @user_id
+    can %i[index show update delete], Beneficiary, created_by_id: @user_id
     can [:show, :update], Beneficiary, order: { submitted_by_id: @user_id }
     can [:show, :update], Beneficiary, order: { created_by_id: @user_id }
     if can_manage_orders? || @api_user
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -106,7 +106,7 @@ class Ability
 
   def beneficiary_abilities
     can :create, Beneficiary
-    can [:index, :show, :update], Beneficiary, created_by_id: @user_id
+    can [:index, :show, :update, :delete], Beneficiary, created_by_id: @user_id
     can [:show, :update], Beneficiary, order: { submitted_by_id: @user_id }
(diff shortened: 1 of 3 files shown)


app/models/ability.rb
```diff
@@ -106,7 +106,7 @@ def appointment_slot_abilities
 
   def beneficiary_abilities
     can :create, Beneficiary
-    can [:index, :show, :update], Beneficiary, created_by_id: @user_id
+    can %i[index show update delete], Beneficiary, created_by_id: @user_id
```
steveyken, line 109: can you add delete to the lines below so the order submitter and the order creator can also delete the beneficiary

lib/tasks/remove_extra_role_permissions.rake
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+# rake goodcity:remove_extra_role_permissions
+
+namespace :goodcity do
+  task remove_extra_role_permissions: :environment do
+    permissions_roles = YAML.load_file("#{Rails.root}/db/permissions_roles.yml")
```
steveyken, line 7: Given this is quite a core task, I think it would be good to be able to test this with some specs. 

To do this, extract the work out of the rake task into a class called something like `lib/roles_permissions_mappings.rb` and then write some specs to prove that it adds missing permissions and remove extraneous permissions. Then we can be confident it will work as expected and can also start using in capistrano tasks.

e.g. The rake task becomes just a wrapper for calling the main class.

```
require 'lib/roles_permissions_mapping'
namespace :goodcity do
  task apply_roles_permissions_mapping: :environment do
    RollsPermissionsMapping.apply!
  end
end
```

lib/roles_permissions_mapping.rb

```
class RolesPermissionsMapping

  def self.apply!
    self.new.apply!
  end

  def apply!
    add_missing_roles
    remove_extraneous_roles
  end

  def add_missing_roles
    ..
  end

  def remove_extraneous_roles
    ...
  end
end
```

Now you can write spec/lib/roles_permissions_mapping_spec.rb

Makes sense? It's a general pattern I'm keen to use in the project. Keeping the main code outside of the rake task itself so that it is easy to test and can also be run in a different scenario... e.g. a console or a sidekiq job or even an admin api endpoint... etc etc

PR #994, opened 2020-04-28
Adding packages valuations (v2)

steveyken:
### Ticket Link: 

### What does this PR do?

FEATURE: Adds to ability to calculate default valuations of packages based on their type, condition and grade.

### Impacted Areas

API / DB schema only. No frontend changes at present.

### Any Open question(s) or challenge(s):

* Need to add default valuations to package types


db/seeds.rb
```diff
@@ -160,6 +160,14 @@
   end
 end
 
+# Valuation matrix
+valuation_matrix = YAML.load_file("#{Rails.root}/db/valuation_matrix.yml")
```
bharat619, line 164: @steveyken , wondering if it would be better in a rake file? Thoughts

steveyken: You're thinking in the right direction! Rails actually does this for us in for `db/seeds.rb`. There is a rake task called `rake db:seed` which is very useful for setting up a database from scratch, esp in development mode. This ties in to that process so no need to change anything here.

PR #995, opened 2020-04-30
GCW-2979 Package valuations - Controller and Model actions

bharat619:
### Ticket Link: 
This also supports for https://jira.crossroads.org.hk/browse/GCW-3127

### What does this PR do?
FEATURE: Adds to ability to calculate default valuations of packages based on their type, condition and grade.

### Linked PR's:
This supports https://github.com/crossroads/stock.goodcity/pull/723
https://github.com/crossroads/api.goodcity/pull/994

### DIFF
--- a/app/controllers/api/v1/packages_controller.rb
+++ b/app/controllers/api/v1/packages_controller.rb
@@ -3,8 +3,10 @@ module Api
     class PackagesController < Api::V1::ApiController
       include GoodcitySync
 
-      load_and_authorize_resource :package, parent: false
+      load_and_authorize_resource :package, parent: false, except: :package_valuation
       skip_before_action :validate_token, only: [:index, :show]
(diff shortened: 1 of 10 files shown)


app/controllers/api/v1/packages_controller.rb
```diff
@@ -181,6 +183,15 @@ def print_barcode
         print_inventory_label
       end
 
+      def package_valuation
```
patrixr, line 186: Could you add the apipie controller descriptors above the method ?

e.g
``` ruby
      api :DELETE, "/v1/packages/1", "Delete an package"
      description "Deletion of the Package item in review mode"
      def destroy
          ...
      end
```

We kinda stopped doing that, and now we sometimes can't tell endpoint methods from helper methods 😆 
I'm trying to bring back that habit :)

PR #997, opened 2020-05-05
GCW-2909 Fix editing the order after cancelling it

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-2909

### What does this PR do?
This Fixes the scenario where the user is able to edit the order even after its cancelled

### Impacted Areas
Browse -> Request Purpose and Appointment screens

### Linked PR's:
https://github.com/crossroads/browse.goodcity/pull/556

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -243,7 +243,7 @@ class Ability
     can %i[index show transition], Order, created_by_id: @user_id
 
     if can_manage_orders? || @api_user
-      can [:create, :index, :show, :transition, :summary], Order
+      can %i[create index show transition summary], Order
       can :index, ProcessChecklist
     end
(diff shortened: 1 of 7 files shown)


app/models/ability.rb
```diff
@@ -240,11 +240,13 @@ def offers_package_abilities
 
   def order_abilities
     can :create, Order
-    can %i[index show update destroy transition], Order, created_by_id: @user_id
+    can %i[index show transition], Order, created_by_id: @user_id
+
     if can_manage_orders? || @api_user
-      can [:create, :index, :show, :update, :transition, :destroy, :summary], Order
```
steveyken, line 245: keep update and destroy here for those who `can_manage_orders`

bharat619: This change has been made as we discussed that no matter who the user is, if the state is cancelled they cannot update or destroy the order

PR #998, opened 2020-05-05
GCW-3105-Enchanced item filters

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3105

### What does this PR do?

Added query for Item filter for inventory actions along with specs.

### DIFF
--- a/app/models/concerns/package_filtering.rb
+++ b/app/models/concerns/package_filtering.rb
@@ -42,7 +42,7 @@ module PackageFiltering
       storage_type_name = options['storage_type_name']
       query = where.not(inventory_number: nil)
       state_filters = states & %w[in_stock received designated ]
-      loss_state_filters = states & %w[dispatched processed lost packed thrashed recyled]
+      loss_state_filters = states & %w[dispatched processed lost packed thrashed preserve]
 
       query = query.where_states(state_filters) if state_filters.any?
       query = query.filter_by_loss_states(loss_state_filters) if loss_state_filters.any?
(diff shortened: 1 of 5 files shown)


app/models/concerns/package_filtering.rb
```diff
@@ -61,6 +63,11 @@ def where_states(states)
       where(sql_query)
     end
 
+    def filter_by_loss_states(states)
+      states = "'#{states.join("','")}'"
+      joins(:package_actions).where("packages_inventories.action in (#{states})")
```
steveyken, line 68: This leads to SQL injection issues. Always write queries with parameters so that ActiveRelation will automatically sanitize them for you: 

```
where("packages_inventories.action IN (:terms)", terms: states)
```

app/models/package.rb
```diff
@@ -38,8 +38,7 @@ class Package < ActiveRecord::Base
   has_many   :requested_packages, dependent: :destroy
   has_many   :offers_packages
   has_many   :offers, through: :offers_packages
-  has_many   :package_actions, -> { where action: %w[trash process recycle loss gain] },
-    class_name: "PackagesInventory"
+  has_many   :package_actions, -> { where action: PackagesInventory::DECREMENTAL_ACTIONS }, class_name: "PackagesInventory"
```
steveyken, line 41: Good change 👍

spec/models/concerns/package_filtering_spec.rb
```diff
@@ -87,6 +87,88 @@
     end
   end
 
+  describe "loss actions on packages" do
```
steveyken, line 90: good work on the specs

app/models/concerns/package_filtering.rb
```diff
@@ -40,10 +40,12 @@ def filter(options={})
       location = options['location']
```
patrixr, line 40: I think we're missing the `dispatched` filter 🤔

abulsayyad123: Dispatched is already implemented in `state_filters` array. On this line
https://github.com/crossroads/api.goodcity/pull/998/files#diff-5cb04308b5cbd66a34bdcc27a04e0759R44

PR #999, opened 2020-05-13
GCW-3132 Support mentioning of a specific user in a message

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3132

NOTE: Specify ticket link

### What does this PR do?
This PR will add the polymorphic relationship for subscriptions and the messages table with offers and orders table.

FEATURE: Specify feature name
Support mentioning of a specific user in a message


app/controllers/api/v1/messages_controller.rb
```diff
@@ -35,11 +35,7 @@ class MessagesController < Api::V1::ApiController
       param :scope, String, desc: "The type of record associated to the messages (order/offer/item)"
       def index
         @messages = apply_scope(@messages, params[:scope]) if params[:scope].present?
-        @messages = @messages.where(id: params[:ids].split(",")) if params[:ids].present?
-        @messages = @messages.where(offer_id: params[:offer_id].split(",")) if params[:offer_id].present?
-        @messages = @messages.where(order_id: params[:order_id].split(",")) if params[:order_id].present?
-        @messages = @messages.where(item_id: params[:item_id].split(",")) if params[:item_id].present?
-        @messages = @messages.with_state_for_user(User.current_user, params[:state]) if params[:state].present?
+        apply_filters
```
patrixr, line 38: I would suggest writing the method as a pure function (generally that's always better than writing a method that modifies the state)

e.g

```
messages = apply_scope(@messages, params[:scope]) if params[:scope].present?
messages = apply_filters(messages, params)
paginate_and_render(messages)
```

It's a general practice we want to push forward with :)

app/models/concerns/message_subscription.rb
```diff
@@ -1,87 +1,11 @@
...
-  private
-
-  # A public subscriber is defined as :
-  #   > Anyone who has a subscription to that record
-  def public_subscribers_to(klass, id)
-    Subscription
-      .joins(:message)
-      .where("#{klass}_id": id, messages: { is_private: false })
-      .pluck(:user_id)
+    Messages::Operations.new(message: self).subscribe_users_to_message
```
patrixr, line 9: I like scoping the operations into its own namespace.

But there's no need to instantiate an object if it gets thrown out immediately after. You can use class methods here

`Messages::Operations.subscribe_users_to_message(message: self)`

app/models/message.rb
```diff
@@ -49,6 +59,6 @@ def mark_read!(user_id, app_name)
 
   # To make up for the lack of polymorphism between offer/item/order. Cached
   def related_object
-    @_obj ||= (offer || order || item)
+    @_obj ||= messageable.instance_of?(Item) ? messageable.offer : messageable
   end
```
patrixr, line 63: I'm not sure this method is really needed, since we can use `messageable`. Or maybe make an alias out of it ?

Also I'm not sure I understand `messageable.instance_of?(Item) ? messageable.offer : messageable`

What if the record is an Order ? then `messageable.offer` wouldn't make any sense.

Looks like an over-complication of things

app/serializers/api/v1/message_serializer.rb
```diff
@@ -3,12 +3,20 @@ class MessageSerializer < ApplicationSerializer
...
-      object.order_id
+      object.messageable_type == 'Order' ? object.messageable_id : nil
+    end
+
+    def offer_id
+      object.messageable_type == 'Offer' ? object.messageable_id : nil
+    end
+
+    def item_id
+      object.messageable_type == 'Item' ? object.messageable_id : nil
```
patrixr, line 19: Good to keep those for backwards compatibility. Although you probably want to mark them as deprecated in a comment. We probably shouldn't rely on them anywhere in the apps. Better to rely on `messageable`

app/services/messages/operations.rb
```diff
@@ -0,0 +1,140 @@
+# frozen_string_literal: true
+
+module Messages
```
patrixr, line 3: Option, you could also make a rails concern out of this. And nest it under the model

We've done this for a few things: e.g https://github.com/crossroads/api.goodcity/blob/master/app/models/concerns/operations/designation_operations.rb

db/migrate/20200512044127_change_messages_to_polymorphic.rb
```diff
@@ -0,0 +1,46 @@
...
+        offer = message.offer
+        message.messageable_type = offer.class.name
+        message.messageable_id = offer.id
+      elsif message.order_id
+        order = message.order
+        message.messageable_type = order.class.name
+        message.messageable_id = order.id
+      end
+      message.save!(validate: false)
+    end
```
patrixr, line 20: Wondering if we want this as a rake task ? I personally like the idea of modifying the data as part of a migration but I know @steveyken has some reservations regarding that 🤔

db/migrate/20200512044127_change_messages_to_polymorphic.rb
```diff
@@ -0,0 +1,46 @@
...
+      case message.messageable_id
+      when 'Order'
+        message.order_id = message.messageable_id
+      when 'Offer'
+        message.offer_id = message.messageable_id
+      when 'Item'
+        message.item_id = message.messageable_id
+        message.offer_id = message.messageable.offer.id
+      end
+    end
```
patrixr, line 42: Am as above - cc @steveyken

db/migrate/20200512044127_change_messages_to_polymorphic.rb
steveyken: Yep, you're right! :) The main reason is because I've been bitten in the past by cases where the model has changed enough over the years as to render the migration unable to run which immediately halts development and CI specs (if they use `rake db:migrate` instead of `rake db:schema:load`).

Here's a scenario:

One day:
- Create a `Message` model with `super_funky_method`
- Write a migration that utilises `super_funky_method`

Some time later:
- Delete `super_funky_method` from Message model
- Try to run `rake db:migrate`
- Previous migration fails because `Message#super_funky_method` no longer exists

Useful lessons I've learnt:
- Rake migrations that manipulate data should ideally focus on db columns (easiest way to ensure this is by sticking to SQL)
- Don't use any of your own model methods (standard Rails AR methods probably ok)

Should it be a rake task or in a migration? I'd ask myself the following questions:
- Would it be useful to run the code again another time?
- Does this code need to be tested? (much easier to test if abstracted to a standard ruby class). Whilst a TDD purest would always answer yes to this question, I'm open to standard non-complex SQL one-liners inside the migration)
- Does the migration need to trigger other actions that happen in ActiveRecord (lifecycle hooks like `after_save`). If not, SQL code in a migration might be fine.

Bit of a long answer, but thought it worth explaining my thinking behind my reservations.

steveyken: Given the above, here's what I'd do...

```
def up
  add_column :messages, :messageable_type, :string
  add_column :messages, :messageable_id, :int
  execute("UPDATE messages SET messageable_type=''Offer", messageable_id=offer_id WHERE offer_id IS NOT NULL")
  execute("UPDATE messages SET messageable_type=''Item", messageable_id=item_id WHERE item_id IS NOT NULL")
  execute("UPDATE messages SET messageable_type=''Order", messageable_id=order_id WHERE order_id IS NOT NULL")
end
```

db/migrate/20200512044127_change_messages_to_polymorphic.rb
```diff
@@ -0,0 +1,46 @@
...
+    remove_column :messages, :item_id
+  end
+
+  def down
+    add_column :messages, :offer_id, :int
+    add_column :messages, :order_id, :int
+    add_column :messages, :item_id, :int
+
+    Message.all.map do |message|
+      case message.messageable_id
```
steveyken, line 33: this should be `message.messageable_type` but would suggest rewriting the migration in an SQL way using a similar method I posted in comments for `up`

app/models/concerns/message_subscription.rb
bharat619: @patrixr , cannot use class method here, as `subscribe_users_to_message` is utilising other methods that are instance methods.
and also, the initilization is important for `subscribe_users_to_message`

app/models/message.rb
bharat619: This method is needed
Here's why
A message can be for Offer, Order or Item
Before the polymorphic changes, if the message has an item_id then it will have an order associated with it as well
So `@_obj ||= (offer || order || item)` would always give either the associated offer or the order. Interpretation, if there is an item then the related object is the offer not the item.

So using the instance_of thing over here

app/serializers/api/v1/message_serializer.rb
bharat619: sure. But I think i'll remove them. We can add it if required

app/models/concerns/offer_search.rb
```diff
@@ -3,36 +3,34 @@ module OfferSearch
...
-        .joins("LEFT OUTER JOIN items ON offers.id = items.offer_id")
-        .joins("LEFT OUTER JOIN messages ON offers.id = messages.offer_id OR items.id = messages.item_id")
-        .joins("LEFT OUTER JOIN packages ON packages.item_id = items.id")
-        .joins("LEFT OUTER JOIN package_types ON package_types.id = packages.package_type_id")
-        .joins("LEFT OUTER JOIN deliveries ON offers.id = deliveries.offer_id")
-        .joins("LEFT OUTER JOIN gogovan_orders ON gogovan_orders.id = deliveries.gogovan_order_id")
+        .joins('LEFT OUTER JOIN companies ON offers.company_id = companies.id')
+        .joins('LEFT OUTER JOIN users ON offers.created_by_id = users.id')
+        .joins('LEFT OUTER JOIN items ON offers.id = items.offer_id')
+        .joins("LEFT OUTER JOIN messages ON offers.id = messages.messageable_id and messages.messageable_type = 'Offer'")
```
steveyken, line 28: Don't forget we need messages on items too. You can reference `items.id` as this is already joined above. 

```suggestion
        .joins("LEFT OUTER JOIN messages ON offers.id = messages.messageable_id and messages.messageable_type = 'Offer'")
        .joins("LEFT OUTER JOIN messages ON items.id = messages.messageable_id and messages.messageable_type = 'Item'")
```

Make sure there are some specs to confirm this works and doesn't load items from other unrelated offers.

app/models/message.rb
steveyken: Now that we have `messageable` I would like for us to remove the need for `related_object` entirely and update the front end apps to understand that when `messageable_type=Item` it will know what to do properly. This change would not affect browse app, just donor and admin apps. We should introduce this change carefully and keep `related_object` intact for the timebeing as older apps will use it, but update the data payload receivers all the front-ends as follows:

Remove:

* related_object
* item_id
* identity (this should be called `messageable_type`

Add

* messageable_type
* messageable_id

app/models/concerns/push_updates_for_subscription.rb
```diff
@@ -39,4 +29,26 @@ def app_name
...
   end
 
+  def data_to_send
+    data = { category: 'message',
+             message: message_body,
+             is_private: message.is_private,
+             item_id: item_id,
+             author_id: message.sender_id,
+             message_id: message.id }
+    data[identity] = message.related_object.id
```
steveyken, line 39: Do we really need to open up the related_object just to get the id? Can't we use `messageable_id` ?

```suggestion
    data[messageable_id] = message.messageable_id
```

I don't feel it's a good idea to introduce another column name `identity`. Instead, send `messageable_id` and `messageable_type`

app/models/concerns/push_updates_for_subscription.rb
```diff
@@ -39,4 +29,26 @@ def app_name
...
+    data
+  end
+
+  def message_body
+    message.body.truncate(150, separator: ' ')
+  end
+
+  def item_id
```
steveyken, line 47: I think we can start to remove this from the front-ends now that we will start sending `messageable_id` and `messageable_type`. Perhaps comment the method to say it will be deprecated soon as we can't remove it until the apps are fixed

app/models/concerns/push_updates_for_subscription.rb
```diff
@@ -39,4 +29,26 @@ def app_name
...
+    message.messageable.instance_of?(Item) && message.messageable_id
+  end
+
+  def identity
```
steveyken, line 51: as commented above, the apps need to start to understand `messageable_type` and `messageable_id` . Don't include a new method called `identity`

app/models/offer.rb
```diff
@@ -20,6 +20,8 @@ class Offer < ActiveRecord::Base
   belongs_to :crossroads_transport
   belongs_to :cancellation_reason
   belongs_to :company
+  has_many :messages, as: :messageable, primary_key: :id
+  has_many :subscriptions, as: :subscribable
```
steveyken, line 24: why has the `dependent: :destroy` param been removed for both of these?

app/models/subscription.rb
```diff
@@ -3,7 +3,8 @@ class Subscription < ActiveRecord::Base
   belongs_to :user
   belongs_to :message
   belongs_to :offer
-  belongs_to :order, inverse_of: :subscriptions
+  belongs_to :order
```
steveyken, line 6: Can we get rid of the following associations now?

```
belongs_to :offer
belongs_to :order
```

app/serializers/api/v1/item_serializer.rb
```diff
@@ -3,26 +3,16 @@ class ItemSerializer < ApplicationSerializer
     embed :ids, include: true
 
     attributes :id, :donor_description, :state, :offer_id, :reject_reason,
-      :created_at, :updated_at, :package_type_id, :message_ids, :saleable,
+      :created_at, :updated_at, :package_type_id, :saleable,
       :rejection_comments, :donor_condition_id, :rejection_reason_id
 
     has_many :packages, serializer: PackageSerializer
+    has_many :messages, serializer: MessageSerializer, polymorphic: true
```
steveyken, line 10: I'm not sure about this change... this removes the ability to filter out the non_private and private messages from Items. Can you ping me about his @bharat619 

Also, we don't want to include messages in the data payload for items, but moving them to a `has_many` means they will be included and this will dramatically slow down offer search (which loads items and therefore now, messages)

We should be able to add back in the `message_ids` reference with the following SQL

```
    def message_ids	
      User.current_user.try(:donor?) ? object.messages.non_private.pluck(:id) : object.messages.pluck(:id)	
    end	

    def message_ids__sql	
      need_private = User.current_user.try(:donor?) ?	
        "AND messages.is_private = 'f'" : ""	
      "coalesce((select array_agg(messages.id) from messages where	
        messageable_type = 'Item' AND messageable_id = items.id #{need_private}), '{}'::int[])"	
    end	
```

app/serializers/api/v1/message_serializer.rb
```diff
@@ -3,14 +3,10 @@ class MessageSerializer < ApplicationSerializer
     embed :ids, include: true
 
     attributes :id, :body, :state, :is_private, :created_at,
-      :updated_at, :offer_id, :item_id, :designation_id, :order_id
+      :updated_at, :messageable_type, :messageable_id
```
steveyken, line 6: This change really pleases me! It's the whole reason for switching to polymorphism

app/services/messages/base.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    def parse_id_from_mention_text
+      message.body.scan(/\[:\d+\]/)
+    end
+
+    def parse_id_from_decorated_ids(ids)
+      ids.map { |id| id.match(/\d+/).to_s }
+    end
+  end
+end
```
steveyken, line 33: :100:  @bharat619 for making a base class for this.

I do agree with @patrixr this file would be better located in `app/models/concerns/mentionable.rb` It's adding the 'Mentionable' feature to messages model so goes well there.

app/models/concerns/message_subscription.rb
patrixr: Those other instance methods would also become class methods

bharat619: i have refactored this even a bit more. i'll be removing this completely

app/controllers/api/v1/messages_controller.rb
```diff
@@ -74,9 +69,18 @@ def mark_all_read
...
+        end
+        messages = messages.with_state_for_user(current_user, options[:state].split(',')) if options[:state].present?
+        messages
+      end
+
       def apply_scope(records, scope)
         return records unless ALLOWED_SCOPES.include? scope
-        records.where("messages.#{scope}_id IS NOT NULL")
+
+        records.where("messages.messageable_type = '#{scope.camelize}'")
```
patrixr, line 83: Potential SQL injection, can be sanitized with

`records.where("messages.messageable_type = (?)", scope.camelize)`

app/controllers/api/v1/users_controller.rb
```diff
@@ -60,8 +60,23 @@ def orders_count
...
+                      current_user: current_user)
+                 .related_users
+        render json: { users: @users }
+      end
+
       private
 
+      def messageable_obj
+        return Order.find(params[:order_id]) if params[:order_id].present?
+        return Offer.find(params[:offer_id]) if params[:offer_id].present?
```
patrixr, line 77: If no params are given, this would return nil. Could you make sure there's a test for this scenario :)

app/models/concerns/message_subscription.rb
patrixr: Cool, let me know when it's updated.

As a general rule, we don't want to instanciate anything that will only be used to call a method once.

```ruby
# Instanciated version

class MyService
  def do_something; end
end

MyService.new.do_something # Not a good way to expose the entry point

# Static version

class MyService
  def self.do_something; end
end

MyService.do_something # Better way to expose the entry point

# Hybrid version

class MyService
  def self.do_something
    MyService.new.do_something
  end

  def do_something; end
end

MyService.do_something # Same as above, except it's just a syntaxic sugar and you can still instanciate it if more operations are required

```

app/models/message.rb
patrixr: Agreed, this `related_object` shouldn't exist anymore. 

Note: saying `A message can be for Offer, Order or Item` is actually wrong. It can be anything, it just happens to be the case that we've only used it for those models so far :)

spec/controllers/api/v1/offers_controller_spec.rb
```diff
@@ -161,6 +161,20 @@
       get :show, id: offer.id
       expect(response.status).to eq(200)
     end
+
+    context 'polymorphic associations' do
+      let(:item) { create(:item, offer: offer) }
+      let!(:item_message) { create(:message, messageable: item) }
+      let!(:offer_message) { create(:message, messageable: offer) }
+      let!(:order_message) { create(:message, :with_order, sender: user) }
```
patrixr, line 169: Could you modify the `item`, `offer` to share the same ID ? That would allow this to test for the serializer bug I mentioned before :)

app/controllers/api/v1/users_controller.rb
```diff
@@ -60,6 +60,16 @@ def orders_count
         render json: Order.counts_for(params[:id])
       end
 
+      def mentionable_users
+        @users = Messages::Channels
+                 .new(app_name: app_name,
+                      messageable: Offer.find(params[:offer],
```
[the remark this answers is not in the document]
bharat619: yeah, i had kept it as I wasn't completely aware how the interaction would happen with the ember apps. this has been updated now

patrixr:
@steveyken @bharat619 

Just had a thought. Since messages are used by donors, we _NEED_ this change to be backwards compatible. How are we looking on that end ?

app/serializers/api/v1/message_serializer.rb
```diff
@@ -3,18 +3,18 @@ class MessageSerializer < ApplicationSerializer
     embed :ids, include: true
 
     attributes :id, :body, :state, :is_private, :created_at,
-      :updated_at, :offer_id, :item_id, :designation_id, :order_id
+      :updated_at, :messageable_type, :messageable_id, :lookup
```
steveyken, line 6: As per @patrixr note re backwards compatiblity, can you add item_id, order_id and message_id methods here.

E.g.

```
def item_id
  object.messageable_type == 'Item' ? object.messageable_id if  : nil
end

def item_id__sql
  "CASE WHEN messageable_type='Item' THEN messageable_id ELSE NULL END"
end

```

bharat619: @patrixr @steveyken 
cool. I had added this in one of my previous commit, and again removed it thinking it wont be require. I'll add it back. Thanks

app/controllers/api/v1/messages_controller.rb
```diff
@@ -51,7 +47,6 @@ def show
       api :POST, "/v1/messages", "Create an message"
       param_group :message
       def create
-        @message.order_id = order_id
         @message.sender_id = current_user.id
```
steveyken, line 50: To maintain backwards compatibility with the `item_id, order_id, offer_id` on old apps, create the following function and then call it in the `create` and `update` methods (if it exists)

```
def create
...
  if !@message.messageable_id
    @message.messageable_type, @message.messagable_id = old_messagesable_params
  end
...
end
```
and then in the controller, a method to manage the params

```
# for backwards compatibility
def old_messagesable_params
  case
  when params["item_id"].present?
    ['Item', params["item_id"]]
  when params["offer_id"].present?
    ['Offer', params["offer_id"]]
  when params["order_id"].present?
    ['Order', params["order_id"]]
  else
    [nil,nil]
  end
end
```

Don't forget to add some specs for this method too

bharat619: I have updated. Had to make manipulation to the params, as the abilities will not allow to create message. [This commit](https://github.com/crossroads/api.goodcity/pull/999/commits/0fa396579e35c6ea634fbbe50faeef1a94480f6a#diff-e34206da28447c019a2c53781f9bbdbaR110) has the changes

app/models/concerns/message_subscription.rb
bharat619: this has been done : )

app/models/message.rb
bharat619: 100% agreed with @patrixr , I just gave an example of the existing scenario/behaviour : ).

@steveyken I have made the changes, and added a deprecated comment and kept it since it will be required for the older version of the apps.

app/services/messages/operations.rb
```diff
@@ -0,0 +1,152 @@
...
+      new(message).handle_subscriptions
+    end
+
+    def handle_subscriptions
+      format_message_body
+      notify_users unless mentioned_ids.empty?
+      subscribe_users_to_message
+    end
+
+    def subscribe_users_to_message
```
steveyken, line 24: I felt that the code in `message_subscription` was easier to read. It's a bit hard to follow lots of mini-functions (my brain stack doesn't go very deep!)

It seems like `notify_users` just adds more users to the list of subscribed users so that could be absorbed into `subscribe_users_to_message` perhaps renamed as `add_direct_mentioned_users`

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -199,6 +199,96 @@
...
+      context 'if no messageable id is passed in params' do
+        it 'return empty array' do
+          get :mentionable_users, app_name: ADMIN_APP, offer_id: nil, is_private: false
+          expect(parsed_body['users']).to be_empty
+        end
+      end
+
+      context 'if public' do
+        context 'admin app' do
+          it 'returns reviewers and donors' do
```
steveyken, line 228:
```suggestion
          it 'returns reviewers and supervisors' do
```

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -199,6 +199,96 @@
...
+            generate_and_set_token(supervisor)
+            get :mentionable_users, app_name: ADMIN_APP, offer_id: offer.id, is_private: false
+            users = [[User.supervisors.map(&:id), User.reviewers.map(&:id)].flatten - [supervisor.id]].flatten.map { |id| {'id' => id, 'name' => User.find(id).full_name } }
+            expect(parsed_body['users']).to match_array(users)
+          end
+
+          it 'returns donors and supervisors' do
```
steveyken, line 235: This test is the same as the previous one except the user is different. Not sure we need it. However, we do need a spec for `is_private: true`

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -199,6 +199,96 @@
...
+            generate_and_set_token(reviewer)
+            get :mentionable_users, app_name: ADMIN_APP, offer_id: offer.id, is_private: false
+            users = [[User.supervisors.map(&:id), User.reviewers.map(&:id)].flatten - [reviewer.id]].flatten.map { |id| {'id' => id, 'name' => User.find(id).full_name} }
+            expect(parsed_body['users']).to match_array(users)
+          end
+        end
+
+        context 'donor app' do
+          it 'returns reviewers and supervisors' do
```
steveyken, line 244: how about a spec for `is_private: true` to confirm donor's can't access anything extra just by changing a param

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -199,6 +199,96 @@
...
+
+          it 'returns order_fulfulment and charity users' do
+            generate_and_set_token(order_administrator)
+            get :mentionable_users, app_name: STOCK_APP, order_id: order.id, is_private: false
+            users = [[User.order_administrator.map(&:id), User.order_fulfilment.map(&:id), charity.id].flatten - [order_administrator.id]].flatten.map { |id| {"id" => id, "name" => User.find(id).full_name} }
+            expect(parsed_body['users']).to match_array(users)
+          end
+
+          it 'returns order_administrator and charity users' do
+            generate_and_set_token(order_fulfilment)
```
steveyken, line 263: This 2nd spec really just tests the current_user is not included in the list.
Can we have a spec for `is_private: true` and not revealing a subscriber on the private messaging channel?

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -199,6 +199,96 @@
...
+            expect(parsed_body["users"]).to match_array(users)
+          end
+        end
+
+        context 'browse app' do
+          before { request.headers["X-GOODCITY-APP-NAME"] = BROWSE_APP }
+
+          it 'returns order_administrator and order_fulfulment users' do
+            generate_and_set_token(charity)
+            get :mentionable_users, app_name: STOCK_APP, order_id: order.id, is_private: false
```
steveyken, line 275:
```suggestion
            get :mentionable_users, app_name: BROWSE_APP, order_id: order.id, is_private: false
```

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -199,6 +199,96 @@
...
+            expect(parsed_body["users"]).to match_array(users)
+          end
+        end
+      end
+
+      context 'if private' do
+        context 'admin app' do
+          it 'returns supervisors and rviewers' do
+            generate_and_set_token(supervisor)
+            get :mentionable_users, app_name: ADMIN_APP, offer_id: offer.id, is_private: false
```
steveyken, line 286: Why is `is_private` false?

spec/models/abilities/message_abilities_spec.rb
```diff
@@ -1,109 +1,141 @@
...
-      let(:can)    { [:index, :show, :create, :update, :destroy] }
-      let(:cannot) { [:manage] }
-      it{ can.each do |do_action|
-        is_expected.to be_able_to(do_action, message)
-      end}
-      it{ cannot.each do |do_action|
-        is_expected.to_not be_able_to(do_action, message)
-      end}
+
+    context 'and message is not is_private' do
```
steveyken, line 29: why this change? feels less concise perhaps?

spec/models/concerns/order_filtering_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    context 'with notifications as all' do
+      it 'returns all notifications for the user' do
+        res = Order.filter(with_notifications: 'all')
+        expect(res.count).to eq(2)
+      end
+
+      it 'does not return notification of other users' do
+        res = Order.filter(with_notifications: 'all')
+        expect(res.map(&:subscriptions).flatten.map(&:user_id).uniq).to include(user.id)
```
steveyken, line 46: this doesn't actually test that the other users are not included... Something like:

```suggestion
        expect(res.map(&:subscriptions).flatten.map(&:user_id).uniq).to == user.id
```

spec/models/concerns/order_filtering_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  describe '.filter' do
+    context 'with notifications as unread' do
+      it 'returns orders with unread notifications' do
+        res = Order.filter(with_notifications: 'unread')
+        expect(res.count).to eq(2)
+      end
+
+      it 'does not return notification of other users' do
+        res = Order.filter(with_notifications: 'unread')
+        expect(res.map(&:subscriptions).flatten.map(&:user_id).uniq).to include(user.id)
```
steveyken, line 29: see comment below

spec/services/messages/channels_spec.rb
```diff
@@ -0,0 +1,177 @@
...
+          it 'does not fetch the current user' do
+            result = Messages::Channels.new(current_user: donor, is_private: false, app_name: DONOR_APP, messageable: offer).related_users
+            expected = { :id => donor.id, :name => donor.full_name }
+            expect(result).not_to include(expected)
+          end
+
+          it 'does not fetch any stock or browse related users' do
+            result = Messages::Channels.new(current_user: donor, is_private: false, app_name: DONOR_APP, messageable: offer).related_users
+
+            expected = [[User.order_administrator.map(&:id), User.order_fulfilment.map(&:id)].flatten].flatten.map { |id| {:id => id, :name => User.find(id).full_name} }
```
steveyken, line 51:
```suggestion
            not_expected = [[User.order_administrator.map(&:id), User.order_fulfilment.map(&:id)].flatten].flatten.map { |id| {:id => id, :name => User.find(id).full_name} }
```

also need to include charity user in the exclusion list here

spec/services/messages/channels_spec.rb
```diff
@@ -0,0 +1,177 @@
...
+          result = Messages::Channels.new(current_user: charity, is_private: false, app_name: BROWSE_APP, messageable: order).related_users
+
+          expected = [User.order_fulfilment.map(&:id), User.order_administrator.map(&:id)].flatten.map { |id| { :id => id, :name => User.find(id).full_name } }
+
+          expect(result).to match_array(expected)
+        end
+
+        it 'does not fetch any admin or donor related users' do
+          result = Messages::Channels.new(current_user: charity, is_private: true, app_name: BROWSE_APP, messageable: order).related_users
+          expected = [User.reviewers.map(&:id), User.supervisors.map(&:id)].flatten.map { |id| {:id => id, :name => User.find(id).full_name} }
```
steveyken, line 161: donor?

spec/services/messages/channels_spec.rb
```diff
@@ -0,0 +1,177 @@
...
+        it 'fetches users with "order fulfilment" or "order administrator" role and charity user' do
+          result = Messages::Channels.new(current_user: order_fulfilment, is_private: false, app_name: STOCK_APP, messageable: order).related_users
+
+          expected = [[User.order_fulfilment.map(&:id), User.order_administrator.map(&:id), charity.id].flatten - [order_fulfilment.id]].flatten.map { |id| {:id => id, :name => User.find(id).full_name } }
+
+          expect(result).to match_array(expected)
+
+          result = Messages::Channels.new(current_user: charity, is_private: false, app_name: BROWSE_APP, messageable: order).related_users
+
+          expected = [User.order_fulfilment.map(&:id), User.order_administrator.map(&:id)].flatten.map { |id| { :id => id, :name => User.find(id).full_name } }
```
steveyken, line 154: duplicated test here?

app/serializers/api/v1/item_serializer.rb
```diff
@@ -3,26 +3,16 @@ class ItemSerializer < ApplicationSerializer
     embed :ids, include: true
 
     attributes :id, :donor_description, :state, :offer_id, :reject_reason,
-      :created_at, :updated_at, :package_type_id, :message_ids, :saleable,
+      :created_at, :updated_at, :package_type_id, :saleable,
       :rejection_comments, :donor_condition_id, :rejection_reason_id
 
     has_many :packages, serializer: PackageSerializer
     has_many :images,   serializer: ImageSerializer, polymorphic: true
+    has_many :messages,   serializer: MessageSerializer, polymorphic: true
```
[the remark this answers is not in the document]
bharat619: @steveyken , yes we need to add it here
`api/v1/messages.json` this call has a couple of options passed, which includes the filters like unread messages

Before the polymorphic changes were introduced, a message record would have `offer_id` and `item_id` value filled. So the OrderSerializer would pick the messages related to both order and item (query => `select * from messages inner join orders on orders.id = messages.order_id`)

But after the polymorphic changes, Item is a separate entity and Order is a separate entity for Message, the OrderSerializer has no way to pick up the items related to offer (order.messages query => `select * from messages inner join orders on orders.id = messages.messageable_id and messages.messageable_type ='Order' `)

So I think itemserializer needs this

app/services/messages/operations.rb
bharat619: I removed the `notify_users` method. since it just adds subscription, it can be handled in `subscribe_users_to_message`

db/migrate/20200512044127_change_messages_to_polymorphic.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    Message.all.map do |message|
+      case message.messageable_type
+      when 'Order'
+        message.order_id = message.messageable_id
+      when 'Offer'
+        message.offer_id = message.messageable_id
+      when 'Item'
+        message.item_id = message.messageable_id
+        message.offer_id = message.messageable.offer.id
```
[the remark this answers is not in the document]
bharat619: This will be needed, as before polymorphic changes, if a message had item_id, then it would have its related offer_id as well

app/services/messages/operations.rb
```diff
@@ -0,0 +1,152 @@
...
+      ref_count = 0
+      sanitize_user_ids(ids)
+      message.body = message.body.gsub(/\[:\d+\]/) do
+        ref_count += 1
+        "[:#{ref_count}]"
+      end
+    end
+
+    def sanitize_user_ids(ids)
+      @mentioned_ids = parse_id_from_decorated_ids(ids)
```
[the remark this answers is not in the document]
bharat619: this is defined in the `app/services/messages/base.rb`

app/serializers/api/v1/item_serializer.rb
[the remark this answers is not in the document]
steveyken: Good explanation. Ok this is fine. I've checked for cases of `ItemSerializer` usage to ensure messages aren't embedded elsewhere unexpectedly and I can't see any so we are good to go. `PackageSerializer` uses `has_one :item, serializer: BrowseItemSerializer` which won't call `ItemSerializer`. Noting here that as confirmed in slack, default scope will send non-private messages only.

db/migrate/20200512044127_change_messages_to_polymorphic.rb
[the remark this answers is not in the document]
steveyken: understood

app/services/messages/channels.rb
```diff
@@ -0,0 +1,54 @@
...
+          .joins(:user_roles)
+          .joins(:roles)
+    end
+
+    def get_users_by_roles(roles)
+      user_roles.where(roles: { name: roles }).uniq
+                .map { |user| { id: user.id, name: user.full_name } }
+    end
+  end
+end
```
steveyken, line 54: As we discussed offline, we'll only have the @ mention functionality for Admin and Stock apps. therefore this class can be massively simplified. In fact, I wonder if it's best to handle the id, full_name in a new serializer and the decisions about which user roles to include can be made using params and scopes in the controller... here's what I mean:

```
def UsersController
    def mentionable_users
        ALLOWED_ROLES = ['Reviewer', 'Supervisor', 'Order Administrator', 'Order Fulfilment']
        roles = (ALLOWED_ROLES & params['roles'].split(',').map(&:strip).uniq) # allow only certain roles
        @users = @users.not_disabled
          .not_deleted
          .not_current_user(current_user.id)
          .with_roles(roles)
          render json: @users, each_serializer: UserMentionableSerializer
      end
end

class User
  scope  :not_disabled, ->{ where(is_disabled: false) }
  scope :with_roles, ->(role_names) { roles.where(name: role_names) }
end

class UserMentionableSerializer < ApplicationSerializer
  embed :ids
  attributes :id, :full_name
end
```

Doing this would mean you can delete the `messages/channels.rb` class

bharat619:
@steveyken , all the comments have been addressed. The things related to donor and charity user mentioning has also been done. And also the message_subscripiton concern refactoring and removing channels.rb file

app/controllers/api/v1/messages_controller.rb
```diff
@@ -74,9 +69,18 @@ def mark_all_read
 
       private
 
+      def apply_filters(messages, options)
+        %i[ids offer_id order_id item_id].map do |f|
```
abulsayyad123, line 73: 👌

app/models/concerns/mentionable.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    format_message_body
+    add_mentions_lookup(user_ids)
+  end
+
+  def add_mentions_lookup(user_ids)
+    lookup = {}
+    lookup_ids = parse_id_from_decorated_ids(parse_id_from_mention_text)
+    lookup_ids.map.with_index do |lookup_id, idx|
+      lookup[lookup_id] = { type: 'User', id: user_ids[idx], display_name: User.find(user_ids[idx]).full_name }
```
abulsayyad123, line 16: Can we cache `display_name`? 🤔   
```
def display_name(user_ids, idx):
    user ||= User.find(user_ids[idx]).full_name
end
```

app/models/concerns/message_subscriptions.rb
```diff
@@ -1,46 +1,44 @@
 #
 # Logic for which users to subscribe to each new message
-module MessageSubscription
+module MessageSubscriptions
```
abulsayyad123, line 3: 👍

app/models/concerns/mentionable.rb
bharat619: that would be problematic, if in case there are multiple users mentioned, it would always display the cached user name

abulsayyad123: Yeah.. make sense. I thought different id will fetch new result.

app/controllers/api/v1/users_controller.rb
```diff
@@ -60,12 +60,23 @@ def orders_count
...
+      end
+
       private
 
       def serializer
         Api::V1::UserSerializer
       end
 
+      def mentionable_role
+        (ALLOWED_MENTIONAblE_ROLES & params['roles'].split(',').map(&:strip).uniq)
```
steveyken, line 77:
```suggestion
        (ALLOWED_MENTIONABLE_ROLES & params['roles'].split(',').map(&:strip).uniq)
```

app/models/ability.rb
```diff
@@ -401,9 +401,9 @@ def stockit_abilities
 
   def user_abilities
     can [:current_user_profile], User
-    can [:show, :update, :orders_count], User, id: @user_id
-    can [:index, :show, :update, :recent_users, :create], User if can_read_or_modify_user?
-    can [:create, :show], User if can_create_donor?
+    can %i[show update orders_count mentionable_users], User, id: @user_id
```
steveyken, line 404: We should restrict the `mentionable_users` action to only users that are allowed to use it.

```
can %i[mentionable_users], User if can_mention_users?
```

Give the `can_mention_users` permission to the roles for Sys Admin, Reviewer, Supervisor, Order Admin..., Order Fulfillment

app/models/concerns/mentionable.rb
steveyken, line 16: We should handle users that have been subsequently deleted.
e.g. if we delete a user then perhaps replace full_name with 'Unknown User' rather than having an error raised for `nil.full_name`. Good to have a spec for this

app/models/concerns/mentionable.rb
```diff
@@ -0,0 +1,42 @@
+module Mentionable
```
steveyken, line 1: This is a well written concern. Simple to read, comprehend and does it's job cleanly.

app/models/concerns/message_subscriptions.rb
```diff
@@ -1,46 +1,44 @@
...
-        obj&.created_by_id.present? && (user_ids == [self.sender_id])
-      end
+    subscribe_all_staff = if is_private
+                            first_message?
+                          else
+                            obj&.created_by_id.present? && (user_ids == [sender_id])
+                          end
 
     user_ids += User.staff.pluck(:id) if subscribe_all_staff
+    user_ids += mentioned_ids if mentioned_ids.present?
```
steveyken, line 38: :+1:

app/models/concerns/message_subscriptions.rb
```diff
@@ -49,39 +47,37 @@ def subscribe_users_to_message
...
 
   # A private subscriber is defined as :
   #   > A supervisor or reviewer who has answered the private thread
-  def private_subscribers_to(klass, id)
+  def private_subscribers_to(obj)
     User.staff
         .joins(messages: [:subscriptions])
         .where(messages: { is_private: true })
-        .where(subscriptions: { "#{klass}_id": id })
+        .where(subscriptions: { subscribable: obj })
```
steveyken, line 63: Great!. Nice simplification. +1 for polymorphism

app/models/concerns/message_subscriptions.rb
```diff
@@ -49,39 +47,37 @@ def subscribe_users_to_message
...
         .pluck(:id)
   end
 
-  def is_first_message_for(klass, id)
-    Message.where(is_private: is_private, "#{klass}_id": id).count.eql? 1
+  def first_message?
```
steveyken, line 67: yes much clearer name

app/models/concerns/push_updates_for_subscription.rb
steveyken: understood this is for backwards compatibility

app/models/message.rb
```diff
@@ -23,32 +23,54 @@ class Message < ActiveRecord::Base
 
   scope :with_eager_load, -> { includes([:sender]) }
   scope :non_private, -> { where(is_private: false) }
-  scope :donor_messages, ->(donor_id) { joins(:offer).where(offers: { created_by_id: donor_id }, is_private: false) }
+  scope :offer, -> { joins("INNER JOIN offers ON messages.messageable_id = offers.id and messages.messageable_type = 'Offer'") }
```
steveyken, line 26: we have a `belongs_to :offer` above, that can be adjusted to understand polymorphism rather than needing a scope.

app/models/message.rb
```diff
@@ -23,32 +23,54 @@ class Message < ActiveRecord::Base
...
+  scope :filter_by_ids, ->(ids) { where(id: ids.split(",")) }
+  scope :filter_by_offer_id, ->(offer_id) { where(messageable_id: offer_id.split(","), messageable_type: "Offer") }
+  scope :filter_by_order_id, ->(order_id) { where(messageable_id: order_id.split(","), messageable_type: "Order") }
+  scope :filter_by_item_id, ->(item_id) { where(messageable_id: item_id.split(","), messageable_type: "Item") }
 
   # used to override the state value during serialization
   attr_accessor :state_value, :is_call_log
 
   after_create do
+    set_mentioned_users
```
steveyken, line 38: :+1:

app/models/message.rb
```diff
@@ -23,32 +23,54 @@ class Message < ActiveRecord::Base
...
+    lookup.each_key { |k| parsed = parsed.gsub("[:#{k}]", lookup[k]["display_name"]) }
+    parsed
+  end
+
+  def mentioned_ids
+    return [] if lookup.empty?
+
+    lookup.keys.map do |k|
+      lookup[k]['id'].to_i
+    end
```
steveyken, line 58:
```suggestion
    (lookup.keys || []).map do |k|
      lookup[k]['id'].to_i
    end
```

app/models/user.rb
```diff
@@ -47,9 +47,13 @@ class User < ActiveRecord::Base
   scope :reviewers, -> { where(roles: {name: "Reviewer"}).joins(:roles) }
   scope :supervisors, -> { where(roles: {name: "Supervisor"}).joins(:roles) }
   scope :order_fulfilment, -> { where(roles: {name: "Order fulfilment"}).joins(:roles) }
+  scope :order_administrator, -> { where(roles: { name: 'Order administrator' }).joins(:roles) }
   scope :system, -> { where(roles: {name: "System"}).joins(:roles) }
   scope :staff, -> { where(roles: {name: ["Supervisor", "Reviewer"]}).joins(:roles) }
   scope :except_stockit_user, -> { where.not(first_name: "Stockit", last_name: "User") }
+  scope :active, -> { where(disabled: false) }
+  scope :exclude_current_user, ->(id) { where.not(id: id) }
```
steveyken, line 55: By changing the name, the method sounds more generically useful.

```suggestion
  scope :exclude_users, ->(id) { where.not(id: id) }
```

app/serializers/api/v1/user_mentions_serializer.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
```
steveyken, line 1: great!

config/initializers/constants.rb
```diff
@@ -33,3 +33,4 @@
   SETTINGS_EDITOR_APP => 'can_manage_settings'
 }
 PACK_UNPACK_ALLOWED_ACTIONS = %w[pack unpack].freeze
+ALLOWED_MENTIONAblE_ROLES = ['Reviewer', 'Supervisor', 'Order administrator', 'Order fulfilment'].freeze
```
steveyken, line 36:
```suggestion
MENTIONABLE_ROLES = ['Reviewer', 'Supervisor', 'Order administrator', 'Order fulfilment'].freeze
```
Need to update the name slightly (capitalization)

Also, I wonder if this might be better as a class constant as it puts it in context for the specific role names. I can go either way on this. Your call

```
class Roles
  MENTIONABLE_ROLES = 
```
and accessible as Roles::MENTIONABLE_ROLES

db/migrate/20200513033844_change_subscriptions_to_polymorphic.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    remove_column :subscriptions, :offer_id
+    remove_column :subscriptions, :order_id
+  end
+
+  def down
+    add_column :subscriptions, :offer_id, :int
+    add_column :subscriptions, :order_id, :int
+
+    Subsctiption.all.map do |subscription|
```
steveyken, line 17:
```suggestion
    Subscription.all.map do |subscription|
```

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -199,6 +199,54 @@
...
+
+      context 'stock app' do
+        it 'returns order_administrator and order_fulfulment users' do
+          generate_and_set_token(order_administrator)
+          get :mentionable_users, order_id: order.id, roles: 'Order administrator, Order fulfilment'
+          users = [[User.order_administrator.map(&:id), User.order_fulfilment.map(&:id)].flatten - [order_administrator.id]].flatten.map { |id| {'id' => id, 'first_name' => User.find(id).first_name, 'last_name' => User.find(id).last_name } }
+          expect(parsed_body['users']).to match_array(users)
+        end
+      end
+    end
```
steveyken, line 250: Can you add a spec for when donors and charity users try to call this method. Should get 403 unauthorized

spec/models/concerns/mentionable_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      end
+
+      context 'if there are no mentions' do
+        let(:message) { create(:message, body: 'Hello. How are you?') }
+        it 'adds no lookup' do
+          message.set_mentioned_users
+          expect(message.reload.lookup).to be_empty
+        end
+      end
+    end
```
steveyken, line 27: Can you add a spec for a user_id reference that no longer exists in our database?

spec/models/concerns/message_subscriptions_spec.rb
```diff
@@ -0,0 +1,214 @@
...
+
+          # The supervisor receives subsequent message of the thread
+          expect(message2).to receive(:add_subscription).with('read', reviewer.id)
+          expect(message2).to receive(:add_subscription).with('unread', supervisor.id)
+          message2.subscribe_users_to_message
+        end
+      end
+    end
+
+    describe '#handle_subscriptions' do
```
steveyken, line 188: this method doesn't exist any more. I think you can move the spec below for `should subscribe mentioned_users` into the `private messages` block above.

spec/models/concerns/mentionable_spec.rb
steveyken: Also might be worth adding specs for some of the other methods in this concern e.g. `extract_user_ids_from_message_body`

spec/serializers/api/v1/user_mentions_serializer_spec.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+context Api::V1::UserMentionsSerializer do
+  let(:user) { create :user }
+  let(:serializer) { Api::V1::UserMentionsSerializer.new(user).as_json }
+  let(:json) { JSON.parse(serializer.to_json) }
+
+  it 'creates JSON' do
```
steveyken, line 10: :+1:

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -199,6 +199,64 @@
...
+      end
+
+      context 'stock app' do
+        it 'returns order_administrator and order_fulfulment users' do
+          generate_and_set_token(order_administrator)
+          get :mentionable_users, order_id: order.id, roles: 'Order administrator, Order fulfilment'
+          users = [[User.order_administrator.map(&:id), User.order_fulfilment.map(&:id)].flatten - [order_administrator.id]].flatten.map { |id| {'id' => id, 'first_name' => User.find(id).first_name, 'last_name' => User.find(id).last_name } }
+          expect(parsed_body['users']).to match_array(users)
+        end
+      end
```
steveyken, line 259: 👍

spec/models/concerns/mentionable_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+        end
+
+        it 'parses only valid user_id' do
+          expect(message2.lookup).to match('2' => { 'id' => user2.id, 'type' => 'User', 'display_name' => user2.full_name })
+          expect(message2.body).to eq('Hello [:0]. I need help from you and [:2]')
+        end
+      end
+    end
+
+    describe '#extract_user_ids_from_message_body' do
```
steveyken, line 39: 👍

spec/models/concerns/message_subscriptions_spec.rb
```diff
@@ -0,0 +1,214 @@
...
+            # The supervisor answers on the the private thread
+            create :message, sender: other_reviewer, messageable: offer, is_private: true
+
+            # The supervisor receives subsequent message of the thread
+            expect(message2).to receive(:add_subscription).with('read', reviewer.id) # sender
+            expect(message2).to receive(:add_subscription).with('unread', other_reviewer.id)
+            message2.subscribe_users_to_message
+          end
+
+          context 'message mentions' do
```
steveyken, line 166: 👍

PR #1005, opened 2020-05-18
GCW-3131-Add mixed condition

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3131

### What does this PR do?

FEATURE: Add a new category called "Mixed" in Package creation drop down in stock and admin. But it should not be visible in other donor app Item `donor_condition`

### DIFF
--- a/app/models/donor_condition.rb
+++ b/app/models/donor_condition.rb
@@ -8,4 +8,9 @@ class DonorCondition < ActiveRecord::Base
   translates :name
 
   validates :name_en, presence: true
+
+  def self.donor_condition_for(type=nil)
+    return where(nil) if type
+    where(visible_to_package: false)
+  end
 end
--- a/app/models/donor_condition.rb
+++ b/app/models/donor_condition.rb
@@ -8,9 +8,4 @@ class DonorCondition < ActiveRecord::Base
   translates :name
 
   validates :name_en, presence: true
-
-  def self.donor_condition_for(type=nil)
-    return where(nil) if type
-    where(visible_to_package: false)
-  end
 end
--- a/db/donor_conditions.yml
+++ b/db/donor_conditions.yml
@@ -1,16 +1,16 @@
 --- !omap
 New:
   :name_zh_tw: 全新
-  :visible_to_package: false
+  :visible_to_donor: true
 Lightly Used:
   :name_zh_tw: 接近全新
-  :visible_to_package: false
+  :visible_to_donor: true
 Heavily Used:
   :name_zh_tw: 較為殘舊
-  :visible_to_package: false
+  :visible_to_donor: true
 Broken:
   :name_zh_tw: 需要維修
-  :visible_to_package: false
+  :visible_to_donor: true
 Mixed:
   :name_zh_tw: 混合的
-  :visible_to_package: true
+  :visible_to_donor: false
--- a/db/donor_conditions.yml
+++ b/db/donor_conditions.yml
@@ -1,9 +1,16 @@
 --- !omap
 New:
   :name_zh_tw: 全新
+  :visible_to_package: false
 Lightly Used:
   :name_zh_tw: 接近全新
+  :visible_to_package: false
 Heavily Used:
   :name_zh_tw: 較為殘舊
+  :visible_to_package: false
 Broken:
   :name_zh_tw: 需要維修
+  :visible_to_package: false
+Mixed:
+  :name_zh_tw: 混合的
+  :visible_to_package: true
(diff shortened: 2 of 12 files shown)


app/models/donor_condition.rb
```diff
@@ -8,4 +8,9 @@ class DonorCondition < ActiveRecord::Base
   translates :name
 
   validates :name_en, presence: true
+
+  def self.donor_condition_for(type=nil)
+    return where(nil) if type
+    where(visible_to_package: false)
+  end
```
patrixr, line 15: something looks off here. type isn't really used

db/donor_conditions.yml
```diff
@@ -1,9 +1,16 @@
...
+  :visible_to_package: false
 Heavily Used:
   :name_zh_tw: 較為殘舊
+  :visible_to_package: false
 Broken:
   :name_zh_tw: 需要維修
+  :visible_to_package: false
+Mixed:
+  :name_zh_tw: 混合的
+  :visible_to_package: true
```
patrixr, line 16: `visible_to_package` doesn't seem to describe what it does properly.

If you want it to only be available on stock and admin, maybe something like `internal` or `staff_only` makes more sense ?

cc @steveyken

lib/tasks/add_mix_donor_condition.rake
```diff
@@ -0,0 +1,15 @@
...
+
+    donor_conditions.each do |name, value|
+      DonorCondition.where(name_en: name).first_or_create do |donor_condition|
+        donor_condition.name_en = name
+        donor_condition.name_zh_tw = value[:name_zh_tw]
+        donor_condition.visible_to_package = value[:visible_to_package]
+      end
+    end
+  end
+end
```
patrixr, line 15:
```suggestion
# rake goodcity:initialize_donor_conditions
namespace :goodcity do
  desc 'initialize_donor_conditions'
  task initialize_donor_conditions: :environment do
    donor_conditions = YAML.load_file("#{Rails.root}/db/donor_conditions.yml")
    donor_conditions.each do |name, fields|
      DonorCondition.where(name_en: name).first_or_create(fields)
    end
  end
end 
```

patrixr: Name of the file would be modified as well. This task does not add the 'Mix' condition. It adds ALL conditions

db/donor_conditions.yml
steveyken: Yeah, would be better to call it `visible_to_donor` perhaps? (and flip the logic around)

steveyken: Make it a required non null field

db/migrate/20200513200816_add_visible_to_package_to_donor_condition.rb
```diff
@@ -0,0 +1,5 @@
+class AddVisibleToPackageToDonorCondition < ActiveRecord::Migration
+  def change
+    add_column :donor_conditions, :visible_to_package, :boolean, default: false
```
steveyken, line 3:
```suggestion
    add_column :donor_conditions, :visible_to_donor, :boolean, default: true
```

lib/tasks/add_mix_donor_condition.rake
```diff
@@ -0,0 +1,15 @@
...
+namespace :goodcity do
+  desc 'Add "MIX" donor_condition'
+  task add_mix_donor_condition: :environment do
+    donor_conditions = YAML.load_file("#{Rails.root}/db/donor_conditions.yml")
+
+    donor_conditions.each do |name, value|
+      DonorCondition.where(name_en: name).first_or_create do |donor_condition|
+        donor_condition.name_en = name
+        donor_condition.name_zh_tw = value[:name_zh_tw]
+        donor_condition.visible_to_package = value[:visible_to_package]
```
steveyken, line 11:
```suggestion
        donor_condition.visible_to_package = value[:visible_to_donor]
```

spec/controllers/api/v1/donor_conditions_controller_spec.rb
```diff
@@ -29,9 +29,16 @@
         create(:donor_condition, name_en: name_en)
       end
     end
-    it "return serialized donor_conditions", :show_in_doc do
+    it "returns donor_conditions not visible for package create/update if `for:package` is not provided in params", :show_in_doc do
       get :index
       expect(response.status).to eq(200)
+      donor_condition = DonorCondition.where('visible_to_package': false)
```
steveyken, line 35:
```suggestion
      donor_condition = DonorCondition.where('visible_to_donor': false)
```

PR #1006, opened 2020-05-19
GCW-3101 Package sets

patrixr:
Changes:

* New `PackageSet` model/controller
* Removal of `set_item_id` on packages
* Addition of `package_set_id` on packages

Lifecycle:

* Packages created with the same parent item (from donor) are automatically added to a common set
* When a package is removed from a set (or destroyed), if a set is found with 1 item or less, it is destroyed


app/models/package_set.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  private
+
+  def unlink_packages
+    Package.where(package_set_id: id).update_all(package_set_id: nil)
+  end
+
+  def ensure_type_integrity
+    errors.add(:errors, I18n.t('package_sets.cannot_change_type')) if package_type_id_changed? && packages.length.positive?
```
steveyken, line 29:
```suggestion
    errors.add(:errors, I18n.t('package_sets.cannot_change_type')) if package_type_id_changed? && packages.size.positive?
```

config/routes.rb
```diff
@@ -1,4 +1,5 @@
 Rails.application.routes.draw do
+  resources :package_sets, except: [:new, :edit]
```
steveyken, line 2: this is outside the `/api/v1/` scope. is this intended?

db/migrate/20200423034154_create_package_sets.rb
```diff
@@ -0,0 +1,16 @@
...
+  def change
+    create_table :package_sets do |t|
+      t.integer :package_type_id
+      t.text :description
+
+      t.timestamps null: false
+    end
+
+    add_column :packages, :package_set_id, :integer
+    remove_column :packages, :set_item_id, :integer
```
steveyken, line 11: Is 'set_item_id' column required for migration or just using item relationship?

config/routes.rb
patrixr: oh, I used `rails generate scaffold` that's why. Will fix it

db/migrate/20200423034154_create_package_sets.rb
patrixr: Yes, using the normal item relationship. Not needed

PR #1009, opened 2020-05-20
Fix case where Stockit will not send a value for number_of_people_helped

steveyken:
### What does this PR do?

This is a hot-fix for an issue where Stockit Shipments were not being updated because they were invalid. This happened because we added a constraint for `number_of_people_helped` to not be null, however this is not compatible with Stockit shipments and local orders.

Reverting change.


steveyken:
I've hot-patched live site so this is already running but submitting for approval to master!

PR #1012, opened 2020-05-21
Inventory Number fixes

steveyken:
### Ticket Link: 

### What does this PR do?

BUG: Fixed a bug whereby the wrong next inventory number was being issued

There were two defects causing the code to issue the wrong next available inventory number.

1. Using regex in Postgresql via ruby strings requires extra care due to string interpolation of ' and \
2. When writing multi-line SQL statements, it's best to use %{...} ruby syntax rather than opening " " so that everything is interpolated in a way that is compatible with PostgreSQL.

### Impacted Areas

Stock: Add new item screen (getting repeated errors saying "Inventory Number is already taken")

### DIFF
--- a/app/models/inventory_number.rb
+++ b/app/models/inventory_number.rb
@@ -7,30 +7,52 @@ class InventoryNumber < ActiveRecord::Base
     self.create!(code: next_code)
   end
 
+  # Generates the next available inventory_number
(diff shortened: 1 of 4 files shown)


steveyken:
I'm going to close this and re-open with some other merged changes

steveyken:
Ref https://github.com/crossroads/api.goodcity/pull/1013

PR #1017, opened 2020-05-26
GCW-3133 Allow creation of Users with user-roles

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3133

### What does this PR do?
- Allow creation of users with extra parameters like disabled, image_id, user_roles
- Only `supervisor`, `order fulfillment` and `order administrator` has ability to set roles for a user


app/controllers/api/v1/users_controller.rb
```diff
@@ -45,7 +55,7 @@ def show
 
       def update
         @user.update_attributes(user_params)
-        if params["user"]["user_role_ids"]
+        if params["user"]["user_role_ids"] && User.current_user.can_manage_users?
```
patrixr, line 58: Is there any way of managing this through the abilities file ? @steveyken

app/controllers/api/v1/users_controller.rb
```diff
@@ -20,6 +20,7 @@ class UsersController < Api::V1::ApiController
         For a donor, this will be just themselves. For administrators, this will be all users.
       EOS
       def index
+        @users = @users.stock_users if bool_param(:stockUsers, false)
```
patrixr, line 23: We generally want to move away from `per app` server design. Everything should be app-agnostic.
What you can do is request per role, or per permission. Which wouldn't involve any app name

swatijadhav: here `stock_users` scope will fetch user having roles `["Supervisor", "Reviewer", "Order fulfilment", "Order administrator"]`
and if `stockUsers` param value is true, it will give users with above roles only..

Should I just rename the param attribute `stockUsers` to something else? or some other way you are suggesting @patrixr

patrixr: I would suggest an API filter

`GET /users?roles=supervisor,reviewer,order_fulfilment,...`

That would allow any app to filter the user for any purpose.

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -119,51 +120,44 @@
   end
 
   describe "PUT user/1" do
```
steveyken, line 122: I think the structure of the `describe "PUT user/`" specs needs to be reworked. I realise this expands the scope of the ticket a bit but I think it's very worthwhile as it's a key area to be confident in.

Here are some thoughts:

```
context "PUT user/1" do
  context "as a user with can_read_and_manage_users permission" do
    it "I can edit first_name of another user"...
    it "I can set the roles of another user"...
  end

  context "as a user when I edit my own details" do
    it "I can change my first name"
    it "I cannot change my roles"
    it "I cannot change the first name of another user"
  end
end
```

I would then move the specs for `create_or_remove_user_roles` from the User controller to the User model spec file

```
context "create_or_remove_user_roles" do
  context "when new user role is added"
  context "when a role is removed (not present in params)"
  context "when params are nil"
end
```

```

**Comment by swatijadhav** [comment]
Thanks for the suggestions @steveyken 👍  I have updated specs for users controller.
Can you please have a look and let me know if any further changes/suggestions.

File: app/controllers/api/v1/users_controller.rb [hunk 4]
```diff
@@ -28,7 +29,20 @@ def index
 
       api :POST, '/v1/users', "Create user"
       def create
-        save_and_render_object_with_errors(@user)
+        @user.assign_attributes(user_params)
+
+        if @user.save
+          if params["user"]["organisations_users_ids"].present?
+            @user.organisations << Organisation.find_by(id: params["user"]["organisations_users_ids"])
```
patrixr, line 36: shouldn't that be run before `@user.save` ? 🤔

abulsayyad123: nope..  `organisations_user` will be created before creating user and `<<` operator directly inserts in db

app/controllers/api/v1/users_controller.rb
```diff
@@ -28,7 +29,20 @@ def index
...
+          end
+
+          if params["user"]["user_role_ids"] && User.current_user.can_manage_users?
+            @user.create_or_remove_user_roles(params["user"]["user_role_ids"])
```
steveyken, line 40: I think it's time to wrap some further constraints around the `create_or_remove_user_roles` method. E.g. we need to prevent a user from being able to give roles that are more elevated than they themselves have. Let's create a separate user concern to manage this... 

```
app/models/concerns/manage_user_roles.rb

class ManageUserRoles

  def update_roles(user, role_ids)
    max_level = User.current_user.roles.max(:level)
    allowed_roles = Role.where(id: role_ids).where("level =< ?", max_level)
    ... rest of code
  end

end

```

So the idea here is that we add a new column "level" to roles.

|Role name | Level |
| Donor | 1 |
| Charity | 1 |
| Reviewer | 5 |
| Supervisor | 10 |
| Order Fulfilment | 5 |
| Order Administrator | 10 |
| Stock Fulfillment | 5 |
| Stock Administrator | 10 |
| System | 15 |
| api-write | 15 |

If you are a supervisor (level 10) then you can add the following roles to users: Donor, Charity, Reviewer, Order Fulfilment, Order Administrator, Stock Fulfilment, Stock Administrator. But NOT System or api-write

app/controllers/api/v1/users_controller.rb
```diff
@@ -70,14 +84,23 @@ def search_user_and_render_json
...
+        User.current_user.id != params["id"].to_i
+      end
+
       def user_params
-        attributes = %i[last_connected last_disconnected
-        first_name last_name email receive_email other_phone title mobile printer_id]
-        attributes.concat([:user_role_ids]) if User.current_user.supervisor?
+        attributes = %i[image_id first_name last_name email receive_email
+          other_phone title mobile printer_id]
+        attributes.concat([:disabled]) if can_update_other_user?
```
steveyken, line 102: We need to guard the setting of the disabled attribute. Can you create a 'can_disable_users' permission and assign to System administrators.

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -119,51 +167,92 @@
...
+          existing_user_roles = user.roles
+          put :update, id: user.id, user: { user_role_ids: [charity_role.id] }
+          expect(user.reload.roles.pluck(:id)).not_to include(charity_role.id)
+          expect(user.reload.roles).to match_array(existing_user_roles)
+        end
 
-      it "removes user roles if user_role_ids parameter is nil" do
-        put :update, id: reviewer.id, user: { user_role_ids: [] }
-        expect(reviewer.reload.roles.count).to eq(0)
+        it "can not update user's disabled value", :show_in_doc do
```
steveyken, line 217:
```suggestion
        it "I cannot disable myself", :show_in_doc do
```

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -119,51 +167,92 @@
...
 
-      it "adds user role", :show_in_doc do
-        expect{
-          put :update, id: reviewer.id, user: { user_role_ids: [role.id] }
-        }.to change(UserRole, :count).by(1)
-        expect(response.status).to eq(200)
-        expect(reviewer.reload.roles).to include(role)
-      end
+      context "as a user when I edit my own details" do
+        it "update last_connected and last_disconnected time", :show_in_doc do
```
steveyken, line 178:
```suggestion
        it "I can update last_connected and last_disconnected time", :show_in_doc do
```

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -119,51 +167,92 @@
...
+
+          expect(response.status).to eq(200)
+          expect(reviewer_user.reload.last_connected.to_date).to eq(5.days.ago.to_date)
+          expect(reviewer_user.reload.last_disconnected.to_date).to eq(3.days.ago.to_date)
+        end
 
-      it "removes user role if existing role_id is not present in params", :show_in_doc do
-        put :update, id: reviewer.id, user: { user_role_ids: [] }
-        expect(reviewer.reload.roles.count).to eq(0)
+        it "can not update user's disabled value", :show_in_doc do
```
steveyken, line 188:
```suggestion
        it "I cannot disable myself", :show_in_doc do
```

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -119,51 +167,92 @@
...
+          put :update, id: reviewer_user.id, user: {disabled: true}
+          expect(response.status).to eq(200)
+          expect(reviewer_user.reload.disabled).to eq(false)
+        end
+
+        it "not have permission to add new roles and remove old roles" do
```
steveyken, line 194:
```suggestion
        it "I cannot edit my own roles" do
```

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -119,51 +167,92 @@
...
+          put :update, id: user.id, user: {disabled: true}
+          expect(response.status).to eq(200)
+          expect(user.reload.disabled).to eq(false)
+        end
       end
 
-      it "do not removes user roles if user_role_ids parameter is not present in request" do
-        put :update, id: reviewer.id, user: { first_name: "abc" }
-        expect(reviewer.reload.roles.count).to eq(1)
+      context "as a user when I edit other user details" do
```
steveyken, line 224:
```suggestion
      context "when I edit another user's details" do
```

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -119,51 +167,92 @@
...
+        it "not have permission to add new roles and remove old roles" do
+          existing_user_roles = reviewer_user.roles
+          put :update, id: reviewer_user.id, user: { user_role_ids: [role.id] }
+          expect(reviewer_user.reload.roles.pluck(:id)).not_to include(role.id)
+          expect(reviewer_user.reload.roles).to match_array(existing_user_roles)
+        end
       end
+    end
+
+    context "Supervisor" do
```
steveyken, line 203:
```suggestion
    context "as a Supervisor" do
```

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -119,51 +167,92 @@
...
+      let(:supervisor) { create(:user_with_token, :supervisor) }
+      let(:charity_role) { create(:role, name: "Charity") }
+
+      before { generate_and_set_token(user) }
+
+      context "as a user when I edit my own details" do
```
steveyken, line 209:
```suggestion
      context "when I edit my own details" do
```

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -119,51 +167,92 @@
...
+        end
       end
 
-      it "do not removes user roles if user_role_ids parameter is not present in request" do
-        put :update, id: reviewer.id, user: { first_name: "abc" }
-        expect(reviewer.reload.roles.count).to eq(1)
+      context "as a user when I edit other user details" do
+        before { generate_and_set_token(user) }
+
+        it "Update only allowed roles for Reviewer user [low level role] " do
```
steveyken, line 227:
```suggestion
        it "Update only allowed roles for Reviewer user [low level role] " do
```
```suggestion
        it "I can change the roles of a Reviewer [lower level role] " do
```

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -119,51 +167,92 @@
...
+          put :update, id: reviewer_user.id,
+            user: {user_role_ids: [charity_role.id, system_admin_role.id]}
+
+          expect(response.status).to eq(200)
+          expect(reviewer_user.roles.pluck(:id)).to include(charity_role.id)
+          expect(reviewer_user.roles.pluck(:id)).to_not include(system_admin_role.id)
```
steveyken, line 233: good spec 👍

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -119,51 +167,92 @@
...
+        end
+
+        it "Does not update roles of user [high level role]" do
```
steveyken, line 236:
```suggestion
        it "I cannot change the roles of a System Administrator [higher level role]" do
```

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -119,51 +167,92 @@
...
+          put :update, id: system_admin_user.id,
+             user: {user_role_ids: [charity_role.id]}
+
+          expect(response.status).to eq(200)
+          expect(system_admin_user.roles.pluck(:id)).to_not include(charity_role.id)
+        end
       end
+    end
+
+    context "System administrator User" do
```
steveyken, line 246:
```suggestion
    context "as a System Administrator user" do
```

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -119,51 +167,92 @@
...
+      let(:supervisor) { create(:user_with_token, :supervisor) }
+      before { generate_and_set_token(system_admin_user) }
 
-      it "adds new roles and removes old roles as per params", :show_in_doc do
-        existing_user_roles = reviewer.roles
-        put :update, id: reviewer.id, user: { user_role_ids: [role.id] }
-        expect(reviewer.reload.roles).to include(role)
-        expect(reviewer.reload.roles).not_to include(existing_user_roles)
+      context "as a user when I edit other user details" do
```
steveyken, line 250:
```suggestion
      context "when I edit another user's details" do
```

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -119,51 +167,92 @@
...
+        it "can update user's disabled value", :show_in_doc do
```
steveyken, line 251:
```suggestion
        it "I can disable the user", :show_in_doc do
```

app/models/concerns/manage_user_roles.rb
```diff
@@ -0,0 +1,44 @@
+module ManageUserRoles
```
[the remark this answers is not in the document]
steveyken: @swatijadhav it would be good to have some unit level specs for this module. I know a lot of integration specs are covered in the users_controller_spec but here is a good place to have `specs/models/concerns/manage_user_roles_spec.rb` to cover a more unit style approach.

app/models/concerns/manage_user_roles.rb
```diff
@@ -0,0 +1,44 @@
...
+    def can_update_roles_for_user?(other_user)
+      self != other_user &&
+      self.max_role_level >= other_user.max_role_level
+    end
+
+    def max_role_level
+      self.roles.maximum("level") || 0
+    end
+
+    def update_allowed_on_roles
```
steveyken, line 22:
```suggestion
    # allow the user to only update roles up to and including their level
    def allowed_roles_to_update
```

steveyken: Even better, this could be a scope on Role

`Role.max_role_level(10)` provides all roles up to and including level 10 roles

`scope :max_role_level, ->(level) { where('level <?', level) }`

app/models/concerns/manage_user_roles.rb
```diff
@@ -0,0 +1,44 @@
...
+      Role.where("level <= ?", max_role_level)
+    end
+
+    def create_or_remove_user_roles(role_ids)
```
steveyken, line 26: As above, don't switch context but continue to pass the `user` in as the object to be changed.

```suggestion
    def create_or_remove_user_roles(user, role_ids)
      role_ids = user.role_ids || [] ...
```

swatijadhav: since using it as instance method, not passed use reference, and other imp thing is `role_ids` value can be empty, as all the roles can be removed (rare case).

app/models/concerns/manage_user_roles.rb
```diff
@@ -1,12 +1,29 @@
...
+  # User has ability to update roles of other users who have max role level same as current-user or
+  # less that it.
+  # Also user has access to roles which has level less than user's max role level.
+  #
+  # Example:
+  # Available Roles: Reviewer(level-5), Supervisor(level-10), Charity(level-1), Donor(level-1)
+  # Consider User has roles Charity and Reviewer, hence User's max role level 5
+  # User has access to only to Charity, Donor and Reviewer roles, can not access Supervisor role.
+  # User can update roles of other users who have max role level <= 5.
+  #
```
steveyken, line 17: 👍

spec/models/role_spec.rb
```diff
@@ -11,4 +11,18 @@
   describe 'Database columns' do
     it { is_expected.to have_db_column(:name).of_type(:string)}
   end
+
+  describe "scope: allowed_roles" do
+    let!(:charity_role) { create :charity_role }
+    let!(:reviewer_role) { create :reviewer_role }
+    let!(:supervisor_role) { create :supervisor_role }
+
+    it "return records having visible_to_admin as true" do
```
steveyken, line 20: visible_to_admin?

app/models/concerns/manage_user_roles.rb
```diff
@@ -1,12 +1,29 @@
...
   included do
 
     def update_roles_for_user(user, role_ids)
       return unless self.can_update_roles_for_user?(user)
 
-      allowed_role_ids = self.update_allowed_on_roles.where(id: role_ids).pluck(:id)
+      allowed_role_ids = Role.allowed_roles(max_role_level)
+                             .where(id: role_ids)
+                             .pluck(:id)
       user.create_or_remove_user_roles(allowed_role_ids)
```
steveyken, line 27:
```suggestion
      create_or_remove_user_roles(user, allowed_role_ids)
```

app/models/concerns/manage_user_roles.rb
steveyken: How about something like this:

```
# Sets the roles for a given user
def update_user_roles(user, role_ids)
  role_ids_to_remove = user.roles.pluck(:id) - role_ids
  user.user_roles.where(role_id: role_ids_to_remove).destroy_all
  role_ids.each do |role_id|
    user.user_roles.where(role_id: role_id).first_or_create
  end
end
```

Or how about this crazy idea... does ActiveRecord handle the adding and removal for us? 

```
def update_user_roles(user, role_ids)
  user.roles = Role.find(id: role_ids)
end
```

swatijadhav: 💯 
```rb
user.roles = Role.find(id: role_ids)
```

app/models/concerns/manage_user_roles.rb
```diff
@@ -36,20 +36,8 @@ def max_role_level
...
-        user_roles.where(role_id: role_id).first_or_create
-      end
-    end
-
-    private
-
-    def remove_old_user_roles(role_ids)
-      role_ids_to_remove = roles.pluck(:id) - role_ids
-      user_roles.where("role_id IN(?)", role_ids_to_remove).destroy_all
+    def create_or_remove_user_roles(user, role_ids)
```
steveyken, line 39: great! given that the simplification works and it's just one line, why not merge it into `update_roles_for_user` and delete `create_or_remove_user_roles`

spec/models/role_spec.rb
```diff
@@ -17,7 +17,7 @@
     let!(:reviewer_role) { create :reviewer_role }
     let!(:supervisor_role) { create :supervisor_role }
 
-    it "return records having visible_to_admin as true" do
+    it "return records having level less than or equals given level value" do
```
steveyken, line 20:
```suggestion
    it "return roles having level less than or equal to a given value" do
```

PR #1023, opened 2020-06-08
GCW-3170 Include messages for order response

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3170

### What does this PR do?
Include messages for an order response


swatijadhav:
@steveyken @patrixr @abulsayyad123 @bharat619 Please review this PR.

app/serializers/api/v1/order_shallow_serializer.rb
```diff
@@ -9,6 +10,12 @@ class OrderShallowSerializer < ApplicationSerializer
       :dispatch_started_by_id, :submitted_at, :submitted_by_id, :people_helped, :beneficiary_id,
       :address_id, :district_id, :booking_type_id, :staff_note, :cancellation_reason_id, :shipment_date
 
+    has_many :messages, serializer: MessageSerializer
```
steveyken, line 13: this will include all messages, including ones with `is_private` set to true. Need to send messages separately to order payload... in admin and donor app we use the `api/v1/messages` endpoint. Same here perhaps?

app/controllers/api/v1/orders_controller.rb
```diff
@@ -56,6 +56,7 @@ def show
         render json: serializer.new(@order,
           root: root,
           exclude_code_details: true,
+          include_messages: bool_param(:include_messages, false),
```
steveyken, line 59: As mentioned further below, I think we shouldn't send the messages with `orders` payload but instead use the `api/v1/messages` endpoint. If we include messages in orders, then we won't be able to filter out is_private messages and the order creator will receive them.

app/models/concerns/message_subscription.rb
```diff
@@ -24,6 +24,7 @@ def subscribe_users_to_message
     #   - SystemUser and StockitUser
     #   - donor/charity user if the message is private (supervisor channel) or offer/order is cancelled
     user_ids = user_ids.flatten.uniq
+    debugger
```
steveyken, line 27: whoops!

app/models/concerns/message_subscriptions.rb
```diff
@@ -45,6 +45,18 @@ def subscribe_users_to_message
...
+  def first_message_subscribers(klass)
+    roles = if klass == 'order'
+      ['Order fulfilment', 'Order administrator']
+    elsif  ['offer', 'item'].include?(klass)
+      ['Reviewer', 'Supervisor']
+    else
+      []
+    end
+
+    User.by_roles(roles).pluck(:id)
```
steveyken, line 57: @bharat619 just flagging potential merge issue with #1030

app/serializers/api/v1/order_shallow_serializer.rb
steveyken: Confirmed with swati that we have a default scope to set `is_private: false`

app/models/concerns/message_subscriptions.rb
bharat619, line 57: @swatijadhav u can use the scope `with_roles`

PR #1030, opened 2020-06-22
GCW-3176 Message Permissions

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3176

### What does this PR do?
Fix message_abilities method

### Impacted Areas
Message controller actions


app/models/ability.rb
```diff
@@ -204,18 +206,22 @@ def item_abilities
...
-    elsif can_create_and_read_messages?
-      can [:index, :show, :create], Message
+      # Supervisors and Reviewers can see all messages on all offers (and items belonging to an offer)
+      # Order fulfillment and Order Administrators are allowed to see all messages on all orders.
+      can %i[index show create], Message
     else
-      can [:index, :show, :create], Message, Message.donor_messages(@user_id) do |message|
+      # A donor / charity user is allowed to read any non-private message
+      # that is on an offer / order created by them
+      can [:index, :show], Message, @user.messages do |message|
```
steveyken, line 216: Not quite... we need to cover the following cases:

1. Donor can `index, show` messages they created: `message.sender_id == donor.id`
2. Donor can `index, show` messages sent to them: `message.subscriptions.map(&:user_id).include(donor.id)`

The way we did this before was with the `donor_messages` method which scoped to non-private messages on offers created by the donor. We need to expand this to include orders.

The problem with the code above is that `@user.messages` is only messages created by the user and doesn't include messages sent to them.

Taking a step back, I think we can simplify this. We need a list of all messages a user is allowed to see. We already control this with the `subscriptions` table. So can't we just do this?

```
class Ability

  def message_abilities
    can [:index, :show, :create], Message, messageable_type: 'Offer'  if can_manage_offer_messages?
    can [:index, :show, :create], Message, messageable_type: 'Item'  if can_manage_offer_messages?
    can [:index, :show, :create], Message, messageable_type: 'Order'  if can_manage_order_messages?
    can [:index, :show, :mark_read, :mark_all_read], Message, id: @user.subscriptions.pluck(:message_id)

    # Normal users can create non private messages on objects they own
    can :create, Message do |message|
      message.related_object&.created_by_id == @user_id && !message.is_private && @user_id != nil
    end

end
```

Note, I've also split `can_manage_messages` into `can_manage_offer_messages` and `can_manage_order_messages`. Seems a sensible way to distinguish between staff working on Offers and staff working on Orders.

Feels way too simple... am I missing a case? The specs should verify that.

bharat619:
@steveyken , updated the specs and ability method as suggested. seems I had overcomplicated that a bit

app/models/ability.rb
```diff
@@ -206,22 +207,15 @@ def item_abilities
...
-      can [:index, :show], Message, @user.messages do |message|
-        message.related_object&.created_by_id == @user_id && !message.is_private
-      end
-      # A donor / charity user cannot create a private message
-      can :create, Message do |message|
-        message.related_object&.created_by_id == @user_id && !message.is_private
-      end
+    can %i[index show create], Message, messageable_type: 'Offer' if can_manage_offer_messages?
+    can %i[index show create], Message, messageable_type: 'Item' if can_manage_offer_messages?
+    can %i[index show create], Message, messageable_type: 'Order' if can_manage_order_messages?
```
steveyken, line 212: @swatijadhav as mentioned in scrum, this might be a good time to add `can_manage_package_messages` permissions in your related PR.
cc @bharat619

app/models/ability.rb
```diff
@@ -206,22 +207,15 @@ def item_abilities
...
+    can %i[index show mark_read mark_all_read], Message, id: @user.subscriptions.pluck(:message_id)
+
+    # Normal users can create non private messages on objects they own
+    can :create, Message do |message|
+      message.related_object&.created_by_id == @user_id && !message.is_private
```
steveyken, line 217: I'd feel safer if we included `&& @user_id != nil` just to cover off that edge case...

spec/models/abilities/message_abilities_spec.rb
```diff
@@ -108,6 +121,19 @@
       end
     end
 
+    context 'when charity user recieves a message from a order admin' do
+      let(:order_administrator) { create(:user, :order_administrator) }
+      let(:message) { create :message, is_private: is_private, messageable: order, sender: order_administrator }
+      let!(:subscription) { create :subscription, subscribable: order, message: message }
```
steveyken, line 127: I think `subscriber` needs a user as well?

spec/models/abilities/message_abilities_spec.rb
```diff
@@ -172,15 +198,27 @@
...
-      is_expected.to be_able_to(:mark_read, message)
+      it 'is not allowed to mark_read subscriptions not belonging to user' do
+        is_expected.to_not be_able_to(:mark_read, message2)
+      end
     end
 
-    it 'is not allowed to mark_read subscriptions not belonging to user' do
-      is_expected.to_not be_able_to(:mark_read, message2)
+    context 'mark_all_read' do
+      it 'is allowed to mrk_all_read all subscriptions belonging to user' do
```
steveyken, line 212:
```suggestion
      it 'is allowed to mark_all_read all subscriptions belonging to user' do
```

PR #1034, opened 2020-06-30
GCW-3169 Allow messaging on Packages.

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3169

### What does this PR do?

Allow messaging on Packages.

### DIFF
--- a/app/controllers/api/v1/messages_controller.rb
+++ b/app/controllers/api/v1/messages_controller.rb
@@ -37,6 +37,7 @@ module Api
       def index
         @messages = apply_scope(@messages, params[:scope]) if params[:scope].present?
         @messages = apply_filters(@messages, params)
+        @messages = @messages.with_state_for_user(current_user, params[:state].split(",")) if params[:state].present?
         paginate_and_render(@messages, serializer)
       end
 
@@ -74,15 +75,18 @@ module Api
       param :is_private, ["true", "false"], desc: "Message Type e.g. [public, private]"
       param :state, String, desc: "Message state (unread|read) to filter on"
       def notifications
-        @messages = apply_scope(@messages, params[:scope]) if params[:scope].present?
+        @messages = apply_scope(@messages, params[:messageable_type]) if params[:messageable_type].present?
         @messages = apply_filters(@messages, params)
+        @messages = @messages.joins(:subscriptions).where(subscriptions: {user_id: current_user.id})
+        @messages = @messages.where(subscriptions: {state: params[:state]}) if params[:state].present?
 
         notification_ids = @messages
           .select("max(@messages.id) AS message_id")
           .group("messageable_type, messageable_id, is_private")
+          .page(page).per(per_page)
         @messages = @messages.where("messages.id IN (?)", notification_ids).order("messages.id DESC")
 
-        paginate_and_render(@messages, notification_serializer)
+        render json: message_response(@messages, notification_serializer)
       end
 
       private
@@ -94,16 +98,11 @@ module Api
         %i[ids offer_id order_id item_id package_id].map do |f|
           messages = messages.send("filter_by_#{f}", options[f]) if options[f].present?
         end
-
-        if options[:state].present? && %w[unread read].include?(options[:state])
-          messages = messages.with_state_for_user(current_user, options[:state].split(','))
-        end
-
         messages
       end
 
       def apply_scope(records, scope)
-        scope = [scope].flatten.compact.uniq
+        scope = scope.split(',').flatten.compact.uniq
         scope = (scope & ALLOWED_SCOPES).map(&:camelize)
         records.where("messages.messageable_type IN (?)", scope)
       end
--- a/app/controllers/api/v1/messages_controller.rb
+++ b/app/controllers/api/v1/messages_controller.rb
@@ -37,7 +37,7 @@ module Api
       def index
         @messages = apply_scope(@messages, params[:scope]) if params[:scope].present?
         @messages = apply_filters(@messages, params)
-        paginate_and_render(@messages)
+        paginate_and_render(@messages, serializer)
       end
 
       api :GET, "/v1/messages/1", "Get a message"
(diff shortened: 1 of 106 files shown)


app/models/concerns/push_updates_for_message.rb
```diff
@@ -92,7 +92,7 @@ def state_for_user(user_id)
   def app_name_for_user(user_id)
     obj = self.related_object
     created_by_id = obj.try(:created_by_id) || obj.try(:offer).try(:created_by_id)
-    if object_class == 'Order'
+    if ['Order', 'Package'].include?(object_class)
       (created_by_id == user_id) ? BROWSE_APP : STOCK_APP
     else # Offer/Item
       (created_by_id == user_id) ? DONOR_APP : ADMIN_APP
```
steveyken, line 98: I'm not sure this logic is going to work for Package here. We basically want to alert the users on the stock app only

```suggestion
    if ['Order'].include?(object_class)
      (created_by_id == user_id) ? BROWSE_APP : STOCK_APP
    elsif ['Offer', 'Item'].include?(object_class)
      (created_by_id == user_id) ? DONOR_APP : ADMIN_APP
    elsif ['Package'].include?(object_class)
      STOCK_APP
```

app/models/concerns/push_updates_for_subscription.rb
```diff
@@ -21,8 +21,8 @@ def send_new_message_notification
   def app_name
     related_object = message.related_object
     klass_name = related_object.class.name
-    created_by_id = related_object.created_by_id
-    if klass_name == 'Order'
+    created_by_id = related_object.try(:created_by_id)
+    if ['Order', 'Package'].include?(klass_name)
```
steveyken, line 25: same as per comment for `push_updates_for_message`. we just want to send package updates to the Stock app

app/models/concerns/push_updates_for_message.rb
swatijadhav: ah right, my bad. Thanks for pointing out this @steveyken

app/controllers/api/v1/messages_controller.rb
```diff
@@ -91,14 +91,31 @@ def apply_filters(messages, options)
...
+            .select("max(messages.id) AS message_id")
+            .group("messageable_type, messageable_id, is_private")
+          messages = messages.where("messages.id IN (?)", notification_ids).order("messages.id DESC")
+        end
+
         messages
       end
 
       def apply_scope(records, scope)
-        return records unless ALLOWED_SCOPES.include? scope
```
steveyken, line 83: A small trick that might be helpful here is to convert scope to an array regardless of it's type, and then you don't need to code separate `if is_a?(Array)` cases.

e.g.
```
scope = [scope].flatten.compact.uniq
> [nil] becomes []
> 1 becomes [1]
> [1,2] becomes [1, 2]
```

app/serializers/api/v1/message_serializer.rb
```diff
@@ -81,5 +81,28 @@ def state__sql
          END"
       end
     end
+
+    def unread_count
```
steveyken, line 85: I'm not sure if this is the right place to put `unread` count... This serializer is for individual messages and groups of messages but not for aggregate data like count of unread messages. This would also explain why `unread_count__sql` wasn't right.

One way we have handled this in the past is to attach use `meta` payloads that go alongside the serializers.

```
{
  meta: { 
      unread_messages_count: 3
  },
  data: {
    ...
  }
}
```

app/models/concerns/message_subscriptions.rb
```diff
@@ -46,13 +46,16 @@ def subscribe_users_to_message
...
-    elsif  ['Offer', 'Item'].include?(klass)
-      ['Reviewer', 'Supervisor']
-    else
-      []
-    end
+    roles = case klass
+      when 'Order'
+        ['Order fulfilment', 'Order administrator']
+      when 'Package'
+        ["Order administrator"]
```
steveyken, line 53: should include order fulfillment also

app/serializers/api/v1/stockit_item_serializer.rb
```diff
@@ -12,6 +12,7 @@ class StockitItemSerializer < ApplicationSerializer
     has_many :package_actions, serializer: PackageActionsSerializer, root: :item_actions
     has_one :storage_type, serializer: StorageTypeSerializer
     has_one :package_set, serializer: PackageSetSerializer::StockFormat
+    has_many :messages, serializer: MessageSerializer, polymorphic: true
```
steveyken, line 15: this can really slow down the serializer as it is used in package search. do we need it here? @swatijadhav

app/serializers/api/v1/message_serializer.rb
swatijadhav: Added separate serializer for message notifications.

app/controllers/api/v1/messages_controller.rb
```diff
@@ -68,24 +68,47 @@ def mark_all_read
...
+        messages = messages.unscoped.where(is_private: bool_param(:is_private, false)) if options[:is_private].present?
+        messages = messages.where(messageable_id: options[:messageable_id]) if options[:messageable_id].present?
+
+        %i[ids offer_id order_id item_id package_id].map do |f|
+          messages = messages.send("filter_by_#{f}", options[f]) if options[f].present?
+        end
+
+        if options[:state].present? && %w[unread read].include?(options[:state])
+          messages = messages.with_state_for_user(current_user, options[:state].split(','))
         end
```
steveyken, line 100:
```suggestion
  # scope to current user subscribed messages
  messages = messages.joins(:subscriptions).where(subscriptions: { user_id: current_user.id} )
  # state = 'unread', state='read', state = nil (means all states)
  messages = messages.where(subscriptions: { state: options[:state] } ) if options[:state]
```

app/models/concerns/message_subscriptions.rb
steveyken: Once #1044 is merged, we can use 'Stock administrator` or better still convert this to find users with specific messages permissions. Same as in https://github.com/crossroads/api.goodcity/pull/1044/files?file-filters%5B%5D=.feature&file-filters%5B%5D=.rb#diff-6063ab65baa7267cd865a29013a91297R108

steveyken:
```
    if ['Offer', 'Item'].include?(object_class)
      message_permissions = ['can_manage_offer_messages']
    elsif object_class == 'Order'
      message_permissions = ['can_manage_order_messages']
    elsif object_class == 'Package'
      message_permissions = ['can_manage_package_messages']
    else
      message_permissions = []
    end
    User.joins(roles: [:permissions]).where(permissions: { name: message_permissions } ).distinct.pluck(:id)
```

app/serializers/api/v1/stockit_item_serializer.rb
steveyken: With work on `notifications` this can now be removed.

app/models/concerns/push_updates_for_message.rb
```diff
@@ -42,7 +42,7 @@ def update_client_store
 
   # All reviewers/supervisors/order_fulfillers
   def notify_deletion_to_subscribers
-    if object_class == "Order"
+    if ["Order", "Package"].include?(object_class)
       channels = [Channel::ORDER_FULFILMENT_CHANNEL]
     else # Offer/Item
       channels = [Channel::REVIEWER_CHANNEL, Channel::SUPERVISOR_CHANNEL]
```
steveyken, line 48: Can use Inventory channel for Packages now (stock app)

```suggestion
    case object_class
    when "Order"
      channels = [Channel::ORDER_FULFILMENT_CHANNEL]
    when "Package"
      channels = [Channel::INVENTORY_CHANNEL]
    else # Offer / Item
      channels = [Channel::REVIEWER_CHANNEL, Channel::SUPERVISOR_CHANNEL]
```

Stock app listens on both Inventory Channel and OrderFulfilment channel so this change shouldn't make any actual difference but it's good to have it using the right channel for future work.

PR #1035, opened 2020-07-02
Stocktake implementation

patrixr:
# Stocktakes

This PR adds the following models/controllers:

- Stocktake
- StocktakeRevision

A stocktake has multiple revisions, each revision represents the quantity of a package.
When a stocktake gets committed, we iterate over each revision and create `gain/loss` actions in order to adjust the quantity.

### DIFF
--- a/app/models/concerns/operations/stock_operations.rb
+++ b/app/models/concerns/operations/stock_operations.rb
@@ -59,7 +59,7 @@ module StockOperations
     # @param [Integer] quantity that is added
     # @param [Location|Id] to the location to add the quantity (or its id)
     #
-    def register_gain(package, quantity:, location: nil, action: "gain", description: nil)
+    def register_gain(package, quantity:, location: nil, action: "gain", description: nil, source: nil)
       package.inventory_lock do
         raise Goodcity::NotInventorizedError unless PackagesInventory.inventorized?(package)
         PackagesInventory.public_send("append_#{action}", {
@@ -67,6 +67,7 @@ module StockOperations
           quantity: quantity.abs,
           location: Utils.to_model(location, Location),
           description: description,
+          source: source
         })
         package.reload
       end
@@ -79,7 +80,7 @@ module StockOperations
     # @param [Integer] quantity the quantity that was lost
     # @param [Location|Id] from_location the location to negate the quantity from(or its id)
     #
-    def register_loss(package, quantity:, location: nil, action: 'loss', description: nil)
+    def register_loss(package, quantity:, location: nil, action: 'loss', description: nil, source: nil)
       package.inventory_lock do
         available_count = PackagesInventory::Computer.available_quantity_of(package)
 
@@ -97,7 +98,8 @@ module StockOperations
           package: package,
           quantity: quantity.abs * -1,
           location: Utils.to_model(location, Location),
-          description: description
+          description: description,
+          source: source
         })
         package.reload
       end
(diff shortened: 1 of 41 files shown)


app/models/stocktake_revision.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  after_save :unset_dirty
+
+  # ---------------------
+  # Computed properties
+  # ---------------------
+
+  watch [PackagesInventory] do |packages_inventory|
+    #
+    # The stocktake count is invalidated is someone performs a quantity action on the package
```
steveyken, line 16:
```suggestion
    # The stocktake count is invalidated if someone performs a quantity action on the package
```

spec/models/concerns/stocktake_processor_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+      expect(errors.length).to eq(0)
+      packages_inventory = PackagesInventory.last(2).select { |pi| pi.package_id == package_2.id }.first
+
+      expect(packages_inventory.action).to eq('loss')
+      expect(packages_inventory.quantity).to eq(-2)
+      expect(packages_inventory.source_id).to eq(stocktake.id)
+      expect(packages_inventory.description).to eq(stocktake.name)
+    end
+
+    it 'doesnt record anything it the quantity is already correct' do
```
steveyken, line 56:
```suggestion
    it 'doesnt record anything if the quantity is already correct' do
```

app/models/concerns/operations/stock_operations.rb
```diff
@@ -112,14 +114,15 @@ def register_loss(package, quantity:, location: nil, action: 'loss', description
     # @param [String] action: the inventory action (optional)
     # @param [String] description: notes to detail the action
     #
-    def register_quantity_change(package, quantity:, location:, action:, description: nil)
+    def register_quantity_change(package, quantity:, location:, action:, description: nil, source: nil)
```
bharat619, line 117: we can make the params list to a hash

lib/goodcity/errors.rb
```diff
@@ -3,7 +3,22 @@ module Goodcity
...
+      super(message);
+      @status = status;
+    end
+
+    def type
+      self.class.name.split(':').last
+    end
+
+    def as_json
+      { error: self.message, type: self.type, status: self.status }
```
bharat619, line 19: self can be removed

app/models/concerns/stocktake_processor.rb
```diff
@@ -0,0 +1,76 @@
...
+    #
+    # @return [boolean]
+    #
+    def process_stocktake(stocktake)
+      errors = []
+
+      raise Goodcity::InvalidStateError.new(I18n.t('stocktakes.invalid_state')) unless stocktake.open?
+      raise Goodcity::InvalidStateError.new(I18n.t('stocktakes.dirty_revisions')) if stocktake.revisions.where(dirty: true).count.positive?
+
+      PackagesInventory.secured_transaction do
```
bharat619, line 22: can be moved to model or a different method, just to avoid a big method

patrixr: most of the processing has already been moved to `apply_package_revision`. This is just a loop that calls that method, the function looks big because I add a lot of spacing between lines

app/models/concerns/operations/stock_operations.rb
patrixr: If you mean doing `def register_quantity_change(package, opts)` then I'd rather not. Because that would make all of them optional. Ruby makes those that don't have a default value are mandatory

PR #1036, opened 2020-07-06
GCW-3189 Fix Member's of set

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3189

### What does this PR do?
Fixes User being able to change member of a set or a box/pallet to an invalid type

### Impacted Areas
Package detail screen

### DIFF
--- a/app/models/package.rb
+++ b/app/models/package.rb
@@ -364,8 +364,6 @@ class Package < ActiveRecord::Base
     end
   end
 
-
-
   def singleton_package?
     received_quantity == 1
   end
@@ -402,7 +400,7 @@ class Package < ActiveRecord::Base
   end
 
   def validate_set_member
-    return unless package_set_id.present?
+    return if package_set_id.blank?
 
     valid_package_ids = [package_set.package_type_id, package_set.package_type.child_package_types.map(&:id)].flatten
     errors.add(:errors, I18n.t('package.invalid_package_type')) unless valid_package_ids.include?(package_type_id)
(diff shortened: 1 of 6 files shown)


app/models/package.rb
```diff
@@ -398,6 +401,12 @@ def validate_set_id
     errors.add(:errors, I18n.t('package_sets.no_box_in_set')) if package_set_id.present? && storage_type&.aggregate?
   end
 
+  def validate_set_member
+    if package_set.presence && package_type != package_set.package_type
+      errors.add(:errors, "Invalid package set")
+    end
```
patrixr, line 407: this condition `package_type != package_set.package_type` is not enough.

A package can have the same type as its set, OR a children type. That condition doesn't take into account the children types

spec/controllers/api/v1/packages_controller_spec.rb
```diff
@@ -1172,6 +1172,24 @@ def test_packages_location_changes(package)
...
+        expect(response.status).to eq(422)
+      end
+
+      it 'should be able to change member of a set ot a box/pallet to valid type' do
+        set = create_list(:package_set, 3, :with_packages).last
+        package_params[:package_set_id] = package_set.id
+        put :update, format: :json, id: uninventorized_package.id, package: package_params
+        expect(response.status).to eq(200)
+      end
+    end
```
patrixr, line 1192: We want to test:

- Adding a package of wrong type to a set (what this code does)
- Changing the type of a package which is already in a set, to a type which is invalid (tests are missing)

PR #1037, opened 2020-07-06
GCW-3186-Adds quantity changes in requested_packages

shreyas098:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3186

### What does this PR do?

FEATURE: Adds quantity changes in requested_packages

### DIFF
--- a/app/controllers/api/v1/requested_packages_controller.rb
+++ b/app/controllers/api/v1/requested_packages_controller.rb
@@ -42,7 +42,7 @@ module Api
         if @requested_package.valid? && @requested_package.save
           render json: @requested_package, serializer: serializer
         else
-          render_error(@requested_package.errors.full_messages.join(". "))
+          render_error(@requested_package.errors.full_messages.join("."))
         end
       end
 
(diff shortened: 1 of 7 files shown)


app/models/ability.rb
```diff
@@ -309,10 +309,11 @@ def package_abilities
              search_stockit_items remove_from_set designate register_quantity_change
              mark_missing move print_inventory_label stockit_item_details
              split_package add_remove_item contained_packages parent_containers
-             fetch_added_quantity], Package
+             fetch_added_quantity requested_packages], Package
```
bharat619, line 312: - remove the requested_packages from both places
- add it in the requested_packages_abilities as
```ruby
def requested_packages_abilities
    can %i[create destroy index checkout update], RequestedPackage, user_id: @user_id
  end
```

shreyas098: 👍

app/controllers/api/v1/requested_packages_controller.rb
```diff
@@ -32,12 +33,19 @@ def index
 
       api :POST, '/v1/requested_packages', "Create a cart item"
       def create
-        if params[:quantity].blank?
-          @requested_package.quantity = PackagesInventory::Computer.available_quantity_of(@requested_package.package)
-        end
```
patrixr, line 37: Can you add a test case to make sure the api handles missing params correctly ? : )

shreyas098: sure

PR #1038, opened 2020-07-06
GCW-3185- Comprehensive Packages history

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3185
### What does this PR do?

FEATURE: On history tab in Item it shows comprehensive history of Actions and version changes applied on Item.

Please review this PR.

### DIFF
--- a/app/controllers/api/v1/packages_inventories_controller.rb
+++ b/app/controllers/api/v1/packages_inventories_controller.rb
@@ -1,0 +1,36 @@
+module Api
+  module V1
+    class PackagesInventoriesController < Api::V1::ApiController
+      load_and_authorize_resource :packages_inventory, parent: false
+
+      resource_description do
+        formats ['json']
+        error 401, "Unauthorized"
+        error 404, "Not Found"
+        error 422, "Validation Error"
+        error 500, "Internal Server Error"
+      end
+
+      api :GET, '/v1/packages_inventories', "List all packages_inventories"
+      def index
+        packages_inventories = @packages_inventories.for_package(params["package_id"])
+        @packages_inventories = packages_inventories.page(page).per(per_page).order('id')
+        render json: { meta: { total_pages: @packages_inventories.total_pages, packages_inventories_count: packages_inventories.size } }.merge(serialized_packages_inventories)
+      end
+
+      private
+
+      def serializer
+        Api::V1::PackageActionsSerializer
+      end
+
+      def serialized_packages_inventories
+        ActiveModel::ArraySerializer.new(
+          @packages_inventories,
+          each_serializer: serializer,
+          root: "item_actions"
+        ).as_json
+      end
+    end
+  end
+end
--- a/app/controllers/api/v1/packages_inventories_controller.rb
+++ b/app/controllers/api/v1/packages_inventories_controller.rb
@@ -13,9 +13,15 @@ module Api
 
       api :GET, '/v1/packages_inventories', "List all packages_inventories"
       def index
-        packages_inventories = @packages_inventories.for_package(params["package_id"])
-        @packages_inventories = packages_inventories.page(page).per(per_page).order('id')
-        render json: { meta: { total_pages: @packages_inventories.total_pages, packages_inventories_count: packages_inventories.size } }.merge(serialized_packages_inventories)
+        raise Goodcity::MissingParamError.new("package_id") unless params["package_id"]
+        packages_inventories = @packages_inventories.for_package(params["package_id"]).page(page).per(per_page).order('id')
+        meta = {
+          total_pages: packages_inventories.total_pages,
+          total_count: packages_inventories.size
+        }
+        render json: { meta: meta }.merge(
+            serialized_packages_inventories(packages_inventories)
+        )
       end
 
       private
@@ -24,9 +30,9 @@ module Api
         Api::V1::PackageActionsSerializer
       end
 
-      def serialized_packages_inventories
+      def serialized_packages_inventories(packages_inventories)
         ActiveModel::ArraySerializer.new(
-          @packages_inventories,
+          packages_inventories,
           each_serializer: serializer,
           root: "item_actions"
         ).as_json
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -312,6 +312,7 @@ class Ability
              fetch_added_quantity], Package
       can %i[show create update destroy], PackageSet
       can %i[index], Restriction
+      can %i[index], PackagesInventory
     end
     can [:show], Package,  orders_packages: { order: { created_by_id: @user_id }}
     can [:show], Package,  requested_packages: { user_id: @user_id }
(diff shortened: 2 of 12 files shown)


app/controllers/api/v1/packages_inventories_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+      def index
+        packages_inventories = @packages_inventories.for_package(params["package_id"])
+        @packages_inventories = packages_inventories.page(page).per(per_page).order('id')
+        render json: { meta: { total_pages: @packages_inventories.total_pages, packages_inventories_count: packages_inventories.size } }.merge(serialized_packages_inventories)
+      end
+
+      private
+
+      def serializer
+        Api::V1::PackageActionsSerializer
```
patrixr, line 24: Should we perhaps rename this to `PackagesInventorySerializer` ? cc @steveyken

app/models/packages_inventory.rb
```diff
@@ -41,6 +42,7 @@ module Actions
   belongs_to :source, polymorphic: true, touch: true
 
   after_create { PackagesInventory.emit(self.action, self) }
+  scope :for_package, ->(package_id) { where(action: PackagesInventory::ALL_INVENTORY_ACTIONS, package_id: package_id)}
```
patrixr, line 45:
```suggestion
  scope :for_package, ->(package) { where(package_id: Utils.to_id(package)) }
```

- If you specify all actions, it's not needed
- used `package` instead of `package_id` to support both an ID or a Model as argument

spec/controllers/api/v1/packages_inventories_controller_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    roles_and_permissions: { 'Reviewer' => ['can_manage_packages', 'can_manage_orders']} )}
+  let!(:package) { create(:package) }
+  let!(:package1) { create(:package) }
+  let!(:packages_inventory) { create(:packages_inventory, :gain, package_id: package.id) }
+  let!(:packages_inventory2) { create(:packages_inventory, :gain, package_id: package.id) }
+  let(:parsed_body) { JSON.parse(response.body) }
+
+  describe "GET packages_inventories" do
+
+    describe 'returns packages inventories for given package_id' do
```
patrixr, line 14: can you add a spec where you don't specify a package_id. We probably want an error message, doing an `index` on the inventory is risky so let's properly lock it down to a single package

app/controllers/api/v1/packages_inventories_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+        error 422, "Validation Error"
+        error 500, "Internal Server Error"
+      end
+
+      api :GET, '/v1/packages_inventories', "List all packages_inventories"
+      def index
+        packages_inventories = @packages_inventories.for_package(params["package_id"])
+        @packages_inventories = packages_inventories.page(page).per(per_page).order('id')
+        render json: { meta: { total_pages: @packages_inventories.total_pages, packages_inventories_count: packages_inventories.size } }.merge(serialized_packages_inventories)
+      end
```
patrixr, line 19: Potential improvements

```suggestion
      def index
        packages_inventories = @packages_inventories
            .for_package(params["package_id"])
            .page(page).per(per_page)
            .order('id')
 
         meta = {
             total_pages: packages_inventories.total_pages,
             total_count: packages_inventories.size # renamed to 'total_count', important
         }

        render json: { meta: meta }.merge(serialized_packages_inventories(packages_inventories))
      end
```

Changes:

- Immutability: I would advise against modifying the class property, keep a FP mindset 
- total_count: you want the meta properties to be generic (we already use total_count in other controllers). Using `packages_inventories_count` is dangerous, imagine a front-end component trying to read data, it can be used for multiple endpoints but it can't guess the property name every time.

app/models/ability.rb
```diff
@@ -312,6 +312,7 @@ def package_abilities
              fetch_added_quantity], Package
       can %i[show create update destroy], PackageSet
       can %i[index], Restriction
+      can %i[index], PackagesInventory
```
bharat619, line 315: Just making a note on this:

The `can_manage_packages` , is available for the roles Reviewer, Supervisor, Order Fulfilment and Order Administrator, which I feel is wrong

- Would it be better to have `can_manage_packages` permissions only for the Order Fulfilment and Order Administrator roles?
OR
- Would it be better to have a separate permission for `PackagesInventory`?

cc: @steveyken

patrixr: My 2 cents:

I agree that the Reviewer shouldn't have this access. Supervisor is all powerful, so no issues there.

Regarding a permission for the inventory:
- It _would_ be better, but I think it'll be closely tied to `can_manage_package`, they'll probably go hand-in-hand all the time. Given that it's read-only for now, I'm fine keeping it like this
- The inventory really works as an internal mechanism, I'm still a bit weary about exposing it. I'm afraid a permission would encourage people to add more functionality on the controller.

Just food for thought, it's a good question

app/controllers/api/v1/packages_inventories_controller.rb
steveyken: yes, good to keep naming consistent

app/models/ability.rb
steveyken: Good issue to raise. I'd like to rename Reviewer to "Offer Reviewer" and Supervisor to "Offer Administrator". These roles only need the ability to manage packages directly related to Offers (via items). We can scope CanCanCan rules to do that.

The Stock Fulfilment, Stock Administrator, Order Fulfillment, Order Administrator roles can have full can_manage_packages permissions and given that PackagesInventories is tightly bound to Packages I think it is fine for that to cover access to both tables so what is done above is fine.

app/serializers/api/v1/stockit_item_serializer.rb
```diff
@@ -12,6 +12,7 @@ class StockitItemSerializer < ApplicationSerializer
     has_many :package_actions, serializer: PackageActionsSerializer, root: :item_actions
     has_one :storage_type, serializer: StorageTypeSerializer
     has_one :package_set, serializer: PackageSetSerializer::StockFormat
+    has_many :versions, serializer: VersionSerializer
```
steveyken, line 15: @abulsayyad123 my concern here is that this will make the serializer quite slow. Have you profiled this at all?

steveyken: StockitItemSerializer is used on PackagesController#create, search_stockit_items, remove_from_set, contained_packages, parent_containers, update, split_package, move, designate, register_quantity_change

These are all places we don't need version history right? Particularly `search_stockit_items`

cc @patrixr

steveyken: Suggestion from @patrixr : Create an endpoint `/api/v1/packages/1/versions` and have it return all versions for the package

PR #1040, opened 2020-07-09
GCW-3215 Handle case insensitivity of email

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3215

### What does this PR do?
Handles the case insensitivity while creating users

### Impacted Areas
User login, User creation

### DIFF
--- a/app/models/concerns/user_search.rb
+++ b/app/models/concerns/user_search.rb
@@ -7,7 +7,7 @@ module UserSearch
   included do
     # user: first_name, last_name, email, mobile
     scope :search, ->(options = {}) {
-      search_text = options[:search_text] || ''
+      search_text = options[:search_text].downcase || ''
       role_name = options[:role_name].presence
       if search_text.present?
         search_query = SEARCH_ATTRIBUTES.map { |f| "#{f} ILIKE :search_text" }.join(" OR ")
--- a/app/models/concerns/user_search.rb
+++ b/app/models/concerns/user_search.rb
@@ -7,7 +7,7 @@ module UserSearch
   included do
     # user: first_name, last_name, email, mobile
     scope :search, ->(options = {}) {
-      search_text = options[:search_text].downcase || ''
+      search_text = options[:search_text] || ''
       role_name = options[:role_name].presence
       if search_text.present?
         search_query = SEARCH_ATTRIBUTES.map { |f| "#{f} ILIKE :search_text" }.join(" OR ")
--- a/lib/classes/resync_duplicate_users.rb
+++ b/lib/classes/resync_duplicate_users.rb
@@ -3,18 +3,13 @@
 class ResyncDuplicateUsers
   def self.apply
     # 1. Fetch the users with duplicate email
-    sql = "SELECT id, email FROM users
-          WHERE email IS NOT NULL AND
-                email != '' AND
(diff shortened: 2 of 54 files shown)


lib/classes/resync_duplicate_users.rb
```diff
@@ -0,0 +1,53 @@
...
+
+class ResyncDuplicateUsers
+  def self.apply
+    # 1. Fetch the users with duplicate email
+    sql = "SELECT id, email FROM users
+          WHERE email IS NOT NULL AND
+                email != '' AND
+                UPPER(email) IN (SELECT UPPER(email) FROM users GROUP BY UPPER(email) HAVING COUNT(*) > 1)
+          ORDER BY email, id"
+    duplicate_users = User.find_by_sql(sql)
```
steveyken, line 11:
```suggestion
    duplicate_users = User.where("email IS NOT NULL AND email != '' AND UPPER(email) IN (SELECT UPPER(email) FROM users GROUP BY UPPER(email) HAVING COUNT(*) > 1)")
```

lib/classes/resync_duplicate_users.rb
```diff
@@ -0,0 +1,53 @@
...
+    # 2. Create a hash of user email as key and the ids as values
+    # Do this only for charity user
+    user_hash = {}
+    duplicate_users.map do |user|
+      _user = User.find(user.id)
```
steveyken, line 16: no need for this line if you use above SQL

spec/controllers/api/v1/authentication_controller_spec.rb
```diff
@@ -41,6 +41,25 @@
...
     end
 
+    context 'email' do
+      before do
+        set_browse_app_header
+      end
+      context 'when email is case-sensitive duplicate' do
+        it 'does not create a new user' do
+          user = create(:user)
+          expect { post :signup, format: 'json', user_auth: { mobile: '', email: user.email, first_name: '', last_name: '', address_attributes: { district_id: '', address_type: '' } } }.not_to change{ User.count }
```
steveyken, line 51: Can you also add a spec to check the validation returned is 'email is already taken'. That will confirm this spec is passing for the right reasons.

bharat619: this action has logic like 
- If user exists, ignore data and just send_verification_pin
- Otherwise, create new user and send pin

so it wont be throwing the error, instead it will just login if user already exists

app/models/concerns/user_search.rb
```diff
@@ -7,7 +7,7 @@ module UserSearch
   included do
     # user: first_name, last_name, email, mobile
     scope :search, ->(options = {}) {
-      search_text = options[:search_text] || ''
+      search_text = options[:search_text].downcase || ''
```
steveyken, line 10: Reading this again, it's not actually needed as we are doing an `ILIKE` search in SQL which is a case-insensitive search.

spec/controllers/api/v1/authentication_controller_spec.rb
steveyken: ok, thanks

steveyken:
having rebased locally, the upstream commits contain a lot of master code and doesn't seem GH is able to rebase the branch. Will close this commit and open another cleaner PR

PR #1043, opened 2020-07-13
GCW-3182-Issues with the Charity/Orders page

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3182

### What does this PR do?

- Added separate endpoint for fetching organisations orders with pagination
- Preload `orders` in `organisation` model for adding `ordersCount` field in `organisation` serializer.
- Added test cases for the same.

Please review this PR

### DIFF
--- a/app/controllers/api/v1/gc_organisations_controller.rb
+++ b/app/controllers/api/v1/gc_organisations_controller.rb
@@ -30,7 +30,8 @@ module Api::V1
 
     api :GET, '/v1/organisations/:id/organisation_orders', "List all orders associated with organisation"
     def organisation_orders
-      orders = @organisation.orders.page(page).per(per_page).order('id')
+      organisation_orders = @organisation.orders
+      orders = organisation_orders.page(page).per(per_page).order('id')
       meta = {
         total_pages: orders.total_pages,
         total_count: orders.size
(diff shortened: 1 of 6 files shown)


spec/controllers/api/v1/gc_organisations_controller_spec.rb
```diff
@@ -88,4 +88,30 @@
       expect(parsed_body).to eq(serialized_gc_organisation)
     end
   end
+
+  describe "GET orders for GC organisation" do
```
patrixr, line 92: Could you add some tests for permissions ? to make sure that only entitled users are allowed to call the endpoint

app/controllers/api/v1/gc_organisations_controller.rb
```diff
@@ -38,11 +51,23 @@ def organisation_name_serializer
       Api::V1::OrganisationNamesSerializer
     end
 
+    def order_serializer
+      Api::V1::OrderShallowSerializer
+    end
+
+    def serialized_orders(orders)
+      ActiveModel::ArraySerializer.new(
```
bharat619, line 59: 👍

app/models/organisation.rb
```diff
@@ -8,6 +8,8 @@ class Organisation < ActiveRecord::Base
   has_many :orders
   has_many :users, through: :organisations_users
 
+  scope :with_eager_load, -> { includes([:orders]) }
```
bharat619, line 11: `with_eager_load` seems very generic

can rename to `scope :with_order, -> { includes(:orders) }`

abulsayyad123: Yup.. 👍

app/controllers/api/v1/gc_organisations_controller.rb
```diff
@@ -28,6 +28,19 @@ def names
       find_record_and_render_json(organisation_name_serializer)
     end
 
+    api :GET, '/v1/organisations/:id/organisation_orders', "List all orders associated with organisation"
```
steveyken, line 31: I realise I'm a little late to the party here but normally the API url would be more like this

`/v1/organisations/1/orders`

abulsayyad123: Sure.. @steveyken i'll this change in separate PR.

PR #1044, opened 2020-07-15
GCW-3212 Adding restrictions for login

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3212

### What does this PR do?
Separates the permissions for Supervisors and Reviewers for admin app and Order Administrator and Order Fulfilment for Stock app 

### Impacted Areas
Entire Application. The main impact is that users with just Reviewer or Supervisor roles will not be able to log in to Stock app


db/permissions_roles.yml
```diff
@@ -26,9 +20,9 @@ Reviewer:
 - can_handle_gogovan_order
 - can_manage_holidays
 - can_destroy_package_with_specific_states
-- can_manage_locations
 - can_read_schedule
 - can_read_versions
+- can_manage_locations
```
steveyken, line 25: can remove `can_manage_locations` from reviewer

db/permissions_roles.yml
```diff
@@ -39,42 +33,29 @@ Reviewer:
 
 Supervisor:
 - can_manage_packages
-- can_manage_package_detail
 - can_manage_offers
-- can_manage_goodcity_requests
 - can_manage_deliveries
-- can_manage_orders
```
steveyken, line 46: For all these permissions that are removed from Supervisor and Reviewer, we need to make sure they are added to Stock fulfillment / Stock Administrator above.

features/user_login.feature
```diff
@@ -0,0 +1,36 @@
...
+    Scenario: A order_administrator can login to stock app
+    Given I am a order_administrator logging into "stock" app
+    And I enter the mobile number
+    And I enter the pin
+    Then I should be allowed to login
+
+    Scenario: A order_administrator cannot login to admin app
+    Given I am a order_administrator logging into "admin" app
+    And I enter the mobile number
+    Then I should not be allowed to login
```
steveyken, line 36: good to add cases for

- Stock fulfillment can login to Stock app
- Stock fulfillment cannot login to Admin app

etc etc

patrixr: Very happy to see cucumber tests :)

Although it does encourage the practice of having per-app code behaviours on the backend, which I think we should get rid of (cc @steveyken )

steveyken:
@bharat619 I've made some extra changes to use permissions rather than roles. can you review?

app/models/concerns/push_updates_for_message.rb
```diff
@@ -103,4 +102,19 @@ def app_name_for_user(user_id)
   def subscribed_user_ids
     @subscribed_user_ids ||= self.subscriptions.pluck(:user_id)
   end
+
+  # All admin users with permission to view messages on that object
+  def relevant_staff_user_ids
```
bharat619, line 107: 👍

features/user_login.feature
steveyken: It does, I'm not sure if there is a way to avoid that in this case. We want to prevent users from logging in to apps they don't have permission on. E.g. donor on Stock app. Can you see a better way?

steveyken:
@bharat619 I think this is ready to merge now

db/permissions_roles.yml
```diff
@@ -42,49 +32,74 @@ Reviewer:
...
+- can_manage_packages_locations
+- can_manage_items
+- can_add_package_types
+- can_manage_images
+- can_destroy_image_for_imageable_states
+- can_add_or_remove_inventory_number
+- can_manage_delivery_address
+- can_handle_gogovan_order
+- can_manage_holidays
+- can_destroy_package_with_specific_states
```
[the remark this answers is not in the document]
bharat619: @steveyken ,
`can_destroy_image_for_imageable_states`, can be remove
`can_add_package_types` we are currently adding it only through a rake task. We can remove that
`can_manage_offers`, can be removed
`can_manage_packages_locations`, can be removed

PR #1045, opened 2020-07-20
FIX=> Separate endpoint for fetching versions for Items.

abulsayyad123:
This PR 
- adds separate endpoint for fetching versions for Item.
- Remove version association in StockitItem Serializer.
- Renamed `organisation_orders` endpoint to `orders`

### DIFF
--- a/app/controllers/api/v1/gc_organisations_controller.rb
+++ b/app/controllers/api/v1/gc_organisations_controller.rb
@@ -28,8 +28,8 @@ module Api::V1
       find_record_and_render_json(organisation_name_serializer)
     end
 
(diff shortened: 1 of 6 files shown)


app/controllers/api/v1/gc_organisations_controller.rb
```diff
@@ -29,7 +29,7 @@ def names
     end
 
     api :GET, '/v1/organisations/:id/organisation_orders', "List all orders associated with organisation"
```
patrixr, line 31:
```suggestion
    api :GET, '/v1/organisations/:id/orders', "List all orders associated with organisation"
```

steveyken: agree

PR #1051, opened 2020-07-23
Ruby upgrade

bharat619:
WIP


steveyken:
We're currently on ruby 2.7.3 so this is not required.

PR #1059, opened 2020-07-30
GCW-2962 Add Inventory Number validation - Post Stockit

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-2962

### What does this PR do?
Adds validation for uniqueness of inventory number after stockit has been disabled

### Impacted Areas
All Package related operations involving Inventory number

### DIFF
--- a/app/models/package.rb
+++ b/app/models/package.rb
@@ -68,6 +68,7 @@ class Package < ActiveRecord::Base
   validates :weight, :pieces, numericality: { allow_blank: true, greater_than: 0 }
   validates :width, :height, :length, numericality: { allow_blank: true, greater_than_or_equal_to: 0 }
   validate  :validate_set_id, on: [:create, :update]
+  validates_uniqueness_of :inventory_number, unless: :stockit_enabled?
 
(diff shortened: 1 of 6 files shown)


features/support/env.rb
```diff
@@ -40,6 +41,8 @@
 
 World(FactoryBot::Syntax::Methods)
 
+STOCKIT_ENABLED = true
```
steveyken, line 44: This is already stubbed in `spec/env/support.rb` Do we still need it here? If so, better to load the other file to get all the env vars.

bharat619: yeah. I tried that. But cucumber is reading only the main env i.e. `.env` in our local and whatever the `env` is set in our circleci settings. It is not reading any other files like how the `rspec` does.

app/models/package.rb
```diff
@@ -209,6 +210,10 @@ def published?
     allow_web_publish.present?
   end
 
+  def stockit_enabled?
+    STOCKIT_ENABLED
+  end
+
```
patrixr, line 216: Don't think this has its place in the package model class

PR #1066, opened 2020-08-07
GCW-3249 avoid sending set packages in package-serializer response

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3249

### What does this PR do?
In package serializer, when package-sets are passed in response, avoid passing packages again (to fix issue of sending non-published packages.)

### DIFF
--- a/app/controllers/api/v1/packages_controller.rb
+++ b/app/controllers/api/v1/packages_controller.rb
@@ -54,6 +54,7 @@ module Api
           include_orders_packages: is_stock_app?,
           include_packages_locations: is_stock_app?,
           is_browse_app: is_browse_app?,
+          exclude_set_packages: true,
           include_package_set: bool_param(:include_package_set, false)
       end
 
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/packages_controller.rb
```diff
@@ -54,6 +54,7 @@ def index
           include_orders_packages: is_stock_app?,
           include_packages_locations: is_stock_app?,
           is_browse_app: is_browse_app?,
+          exclude_set_packages: true,
           include_package_set: bool_param(:include_package_set, false)
```
steveyken, line 58: It's slightly confusing that we have `exclude_set_packages` and `include_package_set` as params now. Can you explain how they are different and perhaps we can suggest a different name to help clarify?

patrixr: I believe `exclude_set_packages` is supposed to "exclude the other packages of the set"
whereas `include_package_set` means "include the set of the package"

I think @swatijadhav 's intention is to have the PackageSet in the response, but without the sibling packages

patrixr: Although I think setting `exclude_set_packages: true` might break existing uses of the API. We likely want it to include the siblings by default, and allow to explicitly disable that @swatijadhav

PR #1075, opened 2020-08-14
GCW-3274 expiry-date for user-roles

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3274

### What does this PR do?
 expiry-date for user-roles

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -19,8 +19,13 @@ class User < ActiveRecord::Base
   has_many :offers_with_unread_messages, class_name: 'Offer', through: :unread_subscriptions, source: :subscribable, source_type: 'Offer'
   has_many :organisations_users
   has_many :organisations, through: :organisations_users
-  has_many :user_roles, -> { where("expiry_date IS NULL OR expiry_date >= ?", Date.today) }
+
+  has_many :user_roles
   has_many :roles, through: :user_roles
+
+  has_many :active_user_roles, -> { where("expiry_date IS NULL OR expiry_date >= ?", Time.now.in_time_zone) },
+            class_name: "UserRole"
+  has_many :active_roles, class_name: "Role", through: :active_user_roles, source: "role"
 
   belongs_to :image, dependent: :destroy
   belongs_to :printer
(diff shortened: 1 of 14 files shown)


spec/models/user_spec.rb
```diff
@@ -283,6 +283,19 @@
       expect(user.user_role_names).to include("Reviewer")
       expect(user.user_role_names.count).to eq(1)
     end
+
```
patrixr, line 286: @swatijadhav Can we add some end to end tests to prove that a user with an expired role cannot  access the API ? Maybe using timecop

db/migrate/20200814124000_add_expiry_date_to_user_roles.rb
```diff
@@ -0,0 +1,5 @@
+class AddExpiryDateToUserRoles < ActiveRecord::Migration
+  def change
+    add_column :user_roles, :expiry_date, :date
```
patrixr, line 3: Probably best to use `:datetime` and compare to `Time.now` to be safe.

I can imagine a scenario where someone volunteers for one day, and we want to give permissions until the end of the work day (6pm) or something like that.

Not entirely sure it'll be needed, but better safe than sorry : )

steveyken: yes agree, let's go for `datetime` so we can specify end of work day

app/models/user.rb
```diff
@@ -19,7 +19,7 @@ class User < ActiveRecord::Base
   has_many :offers_with_unread_messages, class_name: 'Offer', through: :unread_subscriptions, source: :subscribable, source_type: 'Offer'
   has_many :organisations_users
   has_many :organisations, through: :organisations_users
-  has_many :user_roles
+  has_many :user_roles, -> { where("expiry_date IS NULL OR expiry_date >= ?", Date.today) }
```
steveyken, line 22: @swatijadhav would it be better to filter inside `ability.rb` and perhaps have a scope on `UserRole` for `active` / `expired` roles?

```
class UserRole
  scope :active, -> { where("expiry_date IS NULL OR expiry_date >= ?", Date.today) }
end
```

Then we can call

`User.user_roles.active`
`User.user_roles.expired`

features/step_definitions/user_login_steps.rb
```diff
@@ -10,6 +10,11 @@
   header 'X-GOODCITY-APP-NAME', "#{app}.goodcity"
 end
 
+Given(/^I am an old ([supervisor | reviewer | stock_fulfilment]+) logging into "([^"]*)" app/) do |role, app|
+  @user = create(:user, :"with_#{role}_role", :with_can_login_to_admin_permission, role_expiry: 5.days.ago)
+  header "X-GOODCITY-APP-NAME", "#{app}.goodcity"
+end
+
```
patrixr, line 17: This step can probably be merged with the one right above it by making "old" optional 

e.g `/^ I am an (old)? .../`

The 3 steps above are too similar to each others, ideally we want them as a single configurable step

If you want to push forward for now that's fine, but could you add a TODO comment to refactor later ? :)

features/user_login.feature
```diff
@@ -51,3 +51,13 @@ Feature: 'User logging'
     And I enter the mobile number
     And I enter the mobile number
     Then I should be allowed to login
+
```
patrixr, line 54: We should handle the case where a user is already logged in, but his role expires. In which case calling the REST api should return a `403`.

If the user has an **active** `Order fulfilment` role, but an **expired** `Order administrator`. He/she should be able to log in to stock, but should receive a `403` when calling an endpoint which is reserved to the Administrator.

We need to figure out those potential scenarios as well. @steveyken any other that come to mind ?

patrixr (changes requested):
Implementation looks good ! 👍 
I just think we need to expand the test cases a bit

features/user_login.feature
steveyken: Yes, that's a good summary. At it's core, we are checking the available permissions at the beginning of *each* request so the moment their role access expires, they loose the relevant abilities. 403 is sensible.

When a role expires, we may want to consider updating the app UI, so perhaps include the (read-only) role expiry date in the frontend models also (purely for UI reasons). We will want to think this through. It may also help in cases where the role expiry means the user no longer has access to the app entirely (e.g. can't login to Stock app), in which case, we will want to gracefully handle that. Perhaps forcing a logout and a screen saying "Goodbye and thank you for volunteering today", for example.

steveyken (approved):
This is a really nice implementation 💯

patrixr (approved):

PR #1077, opened 2020-08-19
GCW-3276- Multiple favorite printers

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3276

### What does this PR do?

This PR adds 
- Many to many relationship between printers and users. 
- User can have multiple favourite printers. 
- Migration added to remove `printer_id` from user table. Along with migration script to create `printers_users` record for every printer_id field in user table.
- Fixed failing specs and controller and serializer specs added for `printers_user`.


spec/controllers/api/v1/printers_users_controller_spec.rb
```diff
@@ -0,0 +1,76 @@
+require "rails_helper"
+
+RSpec.describe Api::V1::PrintersUsersController, type: :controller do
+  let(:supervisor) { create(:user, :supervisor, :with_can_access_printers_permission)}
```
patrixr, line 4: I think `can_access_printers` isn't enough on its own as a permission.
Because looking at this, it seems it would allow someone with e.g `Stock fulfilment` role to modify a `Supervisor`'s favourite printers.

Those "lower" roles should only be allowed to create/edit their own printers. We probably need a `can_manage_printers` 

**Important**: before doing anything, write the tests that prove this is true

cc @steveyken

db/migrate/20200819173237_remove_printer_id.rb
```diff
@@ -0,0 +1,14 @@
...
+    User.find_each do |user|
+      if user.printer_id
+        PrintersUser.create(user: user, printer_id: user.printer_id, tag: "stock")
+      end
+    end
+    remove_column :users, :printer_id, :integer
+  end
+
+  def down
+    add_column :users, :printer_id, :integer
```
patrixr, line 12: You could also restore the printer_ids

app/controllers/api/v1/printers_users_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      resource_description do
+        formats ['json']
+        error 401, "Unauthorized"
+        error 404, "Not Found"
+        error 422, "Validation Error"
+        error 500, "Internal Server Error"
+      end
+
+      def_param_group :printers_users do
+        param :printers_users, Hash, required: true do
```
patrixr, line 15:
```suggestion
        param :printers_users, Hash, required: true do
          param :printer_id, Integer, desc: "Id of printer"
          param :user_id, Integer, desc: "Id of user"
          param :tag, String, desc: "A string identifier of the printer"
        end
```

app/controllers/api/v1/printers_users_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+          param :printer_id, Integer, desc: "Id of printer"
+          param :user_id, Integer, desc: "Id of user"
+        end
```
patrixr, line 18:
```suggestion
        param :printers_users, Hash, required: true do
          param :printer_id, Integer, desc: "Id of printer"
          param :user_id, Integer, desc: "Id of user"
          param :tag, String, desc: "A string identifier of the printer"
        end
```
```

File: app/controllers/api/v1/printers_users_controller.rb [hunk 5]
```diff
@@ -0,0 +1,43 @@
...
+      end
+
+      api :POST, '/v1/printers_users', "Create an printers_user"
+      def create
```
swatijadhav, line 22:
```suggestion
      param_group :printers_users 
      def create
```

app/models/ability.rb
```diff
@@ -93,6 +93,7 @@ def define_abilities
 
   def printer_abilities
     can :index, Printer if can_access_printers?
+    can [:create, :update], PrintersUser if can_access_printers?
   end
```
patrixr, line 97:
```suggestion
  def printer_abilities
    if can_manage_printers?
      can :index, Printer
      can [:create, :update], PrintersUser
    elsif can_access_printers?
      can :index, Printer
      can [:create, :update], PrintersUser, { user_id: @user_id }
  end
```

I feel we need something like this. We can't have a non super-user modify someone else's printer

patrixr: Thoughts @steveyken

steveyken: Agreed. The boilerplate @patrixr set will work. Suggest using `can_update_my_printers?` so we know the permission is scoped to `user_id` and bring `can :index, Printer` out of the if block

```ruby
  def printer_abilities
    can :index, Printer
    if can_manage_printers?
      can [:create, :update], PrintersUser
    elsif can_update_my_printers?
       can [:create, :update], PrintersUser, { user_id: @user_id }
    end
  end
```

app/models/printers_user.rb
```diff
@@ -0,0 +1,4 @@
+class PrintersUser < ActiveRecord::Base
+  belongs_to :printer
+  belongs_to :user
+end
```
steveyken, line 4: Add a unique constraint validation so that you can't have the same user and printer listed twice for the same tag

`validates :role_id, uniqueness: { scope: [:user_id, :tag }`

See https://apidock.com/rails/ActiveRecord/Validations/ClassMethods/validates_uniqueness_of

db/migrate/20200819173237_remove_printer_id.rb
```diff
@@ -0,0 +1,14 @@
+class RemovePrinterId < ActiveRecord::Migration
+  def up
+    User.find_each do |user|
```
steveyken, line 3: `User.where('printer_id IS NOT NULL').find_each` would be more optimal

db/migrate/20200819173237_remove_printer_id.rb
```diff
@@ -0,0 +1,14 @@
...
+      if user.printer_id
+        PrintersUser.create(user: user, printer_id: user.printer_id, tag: "stock")
```
steveyken, line 5: Would it be better for this to be a rake task? Imagine in the future if we don't have a `PrintersUser` model anymore... this would fail

steveyken (changes requested):
Agree with @patrixr re changes to permissions. Otherwise, just small comments.
Thanks for working on this @abulsayyad123

lib/classes/add_printers_users.rb
```diff
@@ -0,0 +1,12 @@
+class AddPrintersUsers
+  def self.apply!
+    new.apply
+  end
```
patrixr, line 4: Don't think there's any need to instantiate anything in this case, you can put everything under `def self.apply!`

lib/classes/add_printers_users.rb
```diff
@@ -0,0 +1,12 @@
...
+
+  def apply
+    User.where('printer_id IS NOT NULL').find_each do |user|
+      PrintersUser.create(user: user, printer_id: user.printer_id, tag: "stock")
+    end
+    ActiveRecord::Base.connection.execute("Alter table USERS DROP printer_id;")
```
patrixr, line 10: This looks wrong, the drop should be done via the migration file

lib/classes/add_printers_users.rb
```diff
@@ -0,0 +1,12 @@
+class AddPrintersUsers
+  def self.apply!
+    new.apply
+  end
+
+  def apply
+    User.where('printer_id IS NOT NULL').find_each do |user|
+      PrintersUser.create(user: user, printer_id: user.printer_id, tag: "stock")
```
patrixr, line 8: Might want to use `first_or_create` to avoid duplicates

lib/classes/add_printers_users.rb
abulsayyad123: Sure.. i'll add directly code in `self.apply!`.

db/permissions_roles.yml
```diff
@@ -86,10 +88,12 @@ Stock fulfilment: &stock_fulfilment
 - can_manage_order_messages
 - can_manage_stocktake_revisions
 - can_manage_package_messages
+- can_update_my_printers
```
steveyken, line 91: Charity users don't have printers. No need for this permission here

steveyken (approved):

steveyken (approved):

app/models/printers_user.rb
steveyken: @abulsayyad123 did you add the constraint?

steveyken: Can you also validate that both printer_id and user_id exist? We don't want `nil` appearing anywhere

patrixr (approved):

bharat619 (approved):

PR #1080, opened 2020-08-27
GCW-3275 Refactor OrganisationsUser

patrixr:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3275

### What does this PR do?

- Remove the charity role ❗ 
- Add a status column to organisations_users ❗ 
- Refactor OrganisationsUserBuilder
- Refactor OrganisationsUser controlle
- Rewrite all organisations_user specs
- Add OrganisationsUser ability specs


lib/classes/organisations_user_builder.rb
```diff
@@ -11,91 +13,149 @@ class OrganisationsUserBuilder
...
-    @preferred_contact_number = params["preferred_contact_number"]
-    @app_name = app_name
-    fail_with_error(I18n.t("organisations_user_builder.organisation.blank")) unless @organisation_id
-    fail_with_error(I18n.t("organisations_user_builder.user.mobile.blank")) unless @mobile
+  # ------------------------
+  # Entry points
+  # ------------------------
+
+  def self.create(organisations_user_params)
+    OrganisationsUserBuilder.new(organisations_user_params.symbolize_keys).create!
```
bharat619, line 21:
```suggestion
    new(organisations_user_params.symbolize_keys).create!
```

lib/classes/organisations_user_builder.rb
```diff
@@ -11,91 +13,149 @@ class OrganisationsUserBuilder
...
+  end
+
+  def self.update(organisations_user_id, organisations_user_params)
+    OrganisationsUserBuilder.new(organisations_user_params.symbolize_keys).update!(organisations_user_id)
```
bharat619, line 25:
```suggestion
    new(organisations_user_params.symbolize_keys).update!(organisations_user_id)
```

lib/classes/organisations_user_builder.rb
```diff
@@ -11,91 +13,149 @@ class OrganisationsUserBuilder
...
-  def build
-    users = User.where("lower(email) = (?) OR mobile = (?)", @email&.downcase, @mobile)
-    return fail_with_error(I18n.t('organisations_user_builder.invalid.user')) if users.count > 1
+  # ------------------------
+  # Implementation
+  # ------------------------
+
+  def initialize(organisation_id: nil, user_id: nil, user_attributes: nil, position: '', preferred_contact_number: '', status: '', change_author: User.current_user)
+    @change_author            = change_author
+    @user_id                  = user_id.to_i
```
bharat619, line 34: we can avoid this instance variable as `@user` is already present

lib/classes/organisations_user_builder.rb
```diff
@@ -11,91 +13,149 @@ class OrganisationsUserBuilder
...
+  def assert_integrity!(organisations_user)
+    raise Goodcity::ReadOnlyFieldError.new(:user_id).with_status(403)          if organisations_user.user_id != @user_id
+    raise Goodcity::ReadOnlyFieldError.new(:organisation_id).with_status(403)  if organisations_user.organisation_id != @organisation_id
   end
 
-  def user_belongs_to_organisation(user)
-    user.organisation_ids.include?(@organisation_id)
-  end
+  def assert_permissions!
+    return if super_user?(@change_author)
```
bharat619, line 129: `super_user` might feel as a strong word 🤔 
can have something simple as valid_user?(@change_author) since we are mainly checking for `can_manage_organisations_users`

lib/classes/organisations_user_builder.rb
```diff
@@ -11,91 +13,149 @@ class OrganisationsUserBuilder
...
+  end
+
+  def create!
+    @status = OrganisationsUser::Status::PENDING if @status.blank?
 
-    @user = build_user(users)
-    return fail_with_error(@user.errors) unless @user.valid?
-    return fail_with_error(I18n.t("organisations_user_builder.organisation.not_found")) unless organisation
+    assert_non_existing!
+    assert_permissions!
```
bharat619, line 48: Just a thought:
Would it better to do an early return even before creating an instance for this permission check from [here](https://github.com/crossroads/api.goodcity/pull/1080/files#diff-27eae9ac3f37050f9c9f0128aff14289R21)

bharat619 (commented):
AWESOME!!!! 🚀 
Just some minor observations and comments

lib/classes/organisations_user_builder.rb
patrixr: I think `valid_user?` is equally confusing. But I think `manager?` would work well here : )

lib/classes/organisations_user_builder.rb
patrixr: It's a good point. but some checks require records to actually be loaded first, for example checking the role.
So if you want the check in the static method, that would actually move logic out of the instanciated class, and would split things in a weird way.

I originally wanted to to use an instance at all, but there were just so many different params and variables that I caved :(

steveyken (changes requested):

app/controllers/api/v1/organisations_users_controller.rb
```diff
@@ -2,7 +2,7 @@ module Api
   module V1
     class OrganisationsUsersController < Api::V1::ApiController
       authorize_resource :organisations_user, parent: false
-      load_resource :organisations_user, only: [:show]
+      load_resource :organisations_user, only: [:show, :update]
```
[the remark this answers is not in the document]
patrixr: `:create` already goes through authorization

we don't want to have `load_resource` on create because we want the builder to do everything

lib/tasks/organisations.rake
```diff
@@ -1,10 +1,34 @@
...
+    count = Goodcity::Tasks::OrganisationTasks.initialize_status_field!
+    puts("#{count} records processed")
+  end
+
+  desc 'Restore Charity Role'
+  task restore_charity_roles: :environment do
+    Goodcity::Tasks::OrganisationTasks.restore_charity_roles!
+  end
+
+  desc 'Delete Charity Role'
```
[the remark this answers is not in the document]
patrixr: I love that ! we can make it into a gem

spec/controllers/api/v1/goodcity_requests_contoller_spec.rb
```diff
@@ -75,6 +75,43 @@
...
+      let(:package_type) { create(:package_type) }
+      let(:order) { create :order, created_by: charity_user }
+      let(:other_order) { create :order, organisation: other_organisation }
+      let(:organistion_order) { create :order, organisation: organisation }
+
+      before do
+        generate_and_set_token(charity_user)
+      end
+
+      it "allows me to create a request for my own organisation", :show_in_doc do
```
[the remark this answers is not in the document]
patrixr: I guess that is no longer the case based on the permission change you mention in the ability file 👍

spec/controllers/api/v1/organisations_user_controller_spec.rb
```diff
@@ -1,163 +1,727 @@
...
-          post :create, format: :json, organisations_user: new_organisations_user_params
-        }.to change(OrganisationsUser, :count).by(0)
-         .and change(User, :count).by(0)
-        expect(response.status).to eq(422)
-        expect(subject['errors']).to eq([{'message' => I18n.t("organisations_user_builder.existing_user.present"), 'status' => 422}])
+          post :create, format: :json, organisations_user: organisations_user_payload.except(:status)
+        }.to change(OrganisationsUser, :count).by(1)
+
+        expect(OrganisationsUser.last.status).to eq('pending')
+        expect(response.status).to eq(201)
```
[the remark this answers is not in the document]
patrixr: There are already specs to prevent a user from modifying his/her status here

```ruby
        [:approved, :expired, :denied].each do |status|
          it "prevents a user from setting his/her status as '#{status}'" do
            expect {
              put :update, id: organisations_user.id, organisations_user: organisations_user_payload.merge(status: status.to_s)
            }.not_to change(OrganisationsUser, :count)

            expect(response.status).to eq(403)
            expect(parsed_body['error']).to eq('Access Denied')
          end
        end
```

PR #1082, opened 2020-09-01
GCW-3284 Add Organisation Create/Update API's

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3284

### What does this PR do?
Add's API for creating and updating organisation.

### DIFF
--- a/app/controllers/api/v1/gc_organisations_controller.rb
+++ b/app/controllers/api/v1/gc_organisations_controller.rb
@@ -12,6 +12,20 @@ module Api::V1
       error 500, "Internal Server Error"
     end
 
+    def create
+      @organisation.save
+      render json: @organisation, details: false, serializer: organisation_serializer
+    end
+
+    def update
+      @organisation.assign_attributes(organisation_params)
+      if @organisation.save
+        return render json: @organisation, details: false, serializer: organisation_serializer
+      end
+
+      render json: { errors: @organisation.errors.full_messages }, status: 422
+    end
+
     api :GET, '/v1/organisations', "List all organisations"
     def index
       find_record_and_render_json(organisation_serializer)
@@ -42,6 +56,13 @@ module Api::V1
     end
 
     private
+
+    def organisation_params
+      params.require(:organisation)
+            .permit(:name_en, :name_zh_tw, :country_id,
+                    :website, :organisation_type_id,
+                    :description_en, :description_zh_tw)
+    end
 
     def organisation_serializer
       Api::V1::OrganisationSerializer
--- a/app/controllers/api/v1/gc_organisations_controller.rb
+++ b/app/controllers/api/v1/gc_organisations_controller.rb
@@ -31,19 +31,19 @@ module Api::V1
 
     def create
       if @organisation.save
-        return render json: @organisation, skip_orders: true, serializer: organisation_serializer
+        return render json: @organisation, include_orders_count: false, serializer: organisation_serializer
       end
 
       render json: { errors: @organisation.errors.full_messages }, status: 422
     end
 
-    api :PUT, '/v1/gc_organisation/1', 'Create Organisation'
+    api :PUT, '/v1/gc_organisation/1', 'Update Organisation'
     param_group :organisation
 
     def update
       @organisation.assign_attributes(organisation_params)
       if @organisation.save
-        return render json: @organisation, skip_orders: true, serializer: organisation_serializer
+        return render json: @organisation, include_orders_count: false, serializer: organisation_serializer
       end
 
       render json: { errors: @organisation.errors.full_messages }, status: 422
@@ -56,7 +56,7 @@ module Api::V1
 
     api :GET, '/v1/organisations/1', "Details of a package"
     def show
-      record = Api::V1::OrganisationSerializer.new(@organisation, root: "gc_organisations").as_json
+      record = Api::V1::OrganisationSerializer.new(@organisation, root: "gc_organisations", include_orders_count: true).as_json
       render json: record
     end
 
@@ -104,7 +104,8 @@ module Api::V1
       ActiveModel::ArraySerializer.new(
         orders,
         each_serializer: order_serializer,
-        root: "designations"
+        root: "designations",
+        include_orders_count: true
       ).as_json
     end
 
@@ -114,7 +115,7 @@ module Api::V1
       else
         records = @organisations.with_order.search(params["searchText"]).page(params["page"]).per(params["per_page"] || DEFAULT_SEARCH_COUNT)
       end
-      data = ActiveModel::ArraySerializer.new(records, each_serializer: serializer, root: "gc_organisations").as_json
+      data = ActiveModel::ArraySerializer.new(records, each_serializer: serializer, root: "gc_organisations", include_orders_count: true).as_json
       render json: { "meta": { total_pages: records.total_pages, "search": params["searchText"] } }.merge(data)
     end
   end
--- a/app/controllers/api/v1/gc_organisations_controller.rb
+++ b/app/controllers/api/v1/gc_organisations_controller.rb
@@ -13,18 +13,37 @@ module Api::V1
       error 500, "Internal Server Error"
     end
 
+    def_param_group :organisation do
+      param :organisation, Hash, required: true do
+        param :name_en, String, desc: 'English name', allow_nil: false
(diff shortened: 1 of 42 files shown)


app/controllers/api/v1/gc_organisations_controller.rb
```diff
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 
 module Api::V1
   class GcOrganisationsController < Api::V1::ApiController
```
patrixr, line 4: Can someone remind me why we have `GcOrganisationsController` and not `OrganisationsController` ? Can we maybe rename it and redirect the routes ? cc @steveyken @bharat619 @swatijadhav @abulsayyad123 @shreyas098 

Unrelated to this PR, I just noticed

app/serializers/api/v1/organisation_serializer.rb
```diff
@@ -15,5 +16,9 @@ def orders_count
     def orders_count__sql
       "(SELECT COUNT(*) FROM orders o WHERE o.organisation_id = orders.id"
     end
+
+    def include_orders_count?
+      !@options[:skip_orders]
```
patrixr, line 21: can we rename the option to `include_orders_count` ? I think `skip_orders` will lead people to believe that we're loading full orders

app/controllers/api/v1/gc_organisations_controller.rb
```diff
@@ -12,6 +13,42 @@ class GcOrganisationsController < Api::V1::ApiController
...
+
+    def create
+      if @organisation.save
+        return render json: @organisation, skip_orders: true, serializer: organisation_serializer
+      end
+
+      render json: { errors: @organisation.errors.full_messages }, status: 422
+    end
+
+    api :PUT, '/v1/gc_organisation/1', 'Create Organisation'
```
patrixr, line 40:
```suggestion
    api :PUT, '/v1/gc_organisation/1', 'Update Organisation'
```

app/models/ability.rb
```diff
@@ -22,7 +22,9 @@ class Ability
     can_access_printers can_remove_offers_packages can_access_orders_process_checklists
     can_mention_users can_read_users can_manage_printers can_update_my_printers
     can_manage_order_messages can_manage_offer_messages can_disable_user
-    can_manage_stocktakes can_manage_stocktake_revisions can_manage_package_messages
+    can_manage_stocktakes can_manage_stocktake_revisions
+    can_manage_package_messages can_access_organisation_types
+    can_manage_organisation
```
patrixr, line 27:
```suggestion
    can_manage_organisations
```

Just for consistency, we've pluralised all of them.

app/models/ability.rb
```diff
@@ -286,7 +289,14 @@ def orders_process_checklists_abilities
...
+      can %i[index search show orders], Organisation
+    end
+    can %i[create update], Organisation if can_manage_organisation? || @api_user
+  end
+
+  def organisation_type_abilities
+    if can_access_organisation_types? || @api_user
+      can [:index], OrganisationType
     end
   end
```
patrixr, line 301: Can't we make this public ? If I'm not mistaken they already are visible in the repo's yaml files anyways

spec/rails_helper.rb
```diff
@@ -60,6 +60,7 @@
 
   # Create a system_user at beginning of specs
   config.before(:all) do
+    FactoryBot.create(:country, name_en: DEFAULT_COUNTRY)
     FactoryBot.create(:user, :system)
   end
```
patrixr, line 65: If I'm not mistaken, @steveyken is strongly against using `before(:all)` rspec hooks

app/models/organisation.rb
```diff
@@ -8,10 +8,25 @@ class Organisation < ActiveRecord::Base
   has_many :orders
   has_many :users, through: :organisations_users
 
+  validates :name_en, presence: true, uniqueness: true
+  validates :organisation_type_id, presence: true
+
+  before_validation :upcase_name
```
patrixr, line 14: If we actually do intend on making all the names uppercase, we probably want to migrate all the existing data as well

app/serializers/api/v1/organisation_serializer.rb
steveyken: Yes I actually thought that. If we want to include extra attributes/relationships, it would be better if this were generally always excluded by default and included via parameters rather than the other way around. opt-in vs opt-out

db/permissions_roles.yml
```diff
@@ -89,6 +89,8 @@ Stock fulfilment: &stock_fulfilment
 - can_manage_stocktake_revisions
 - can_manage_package_messages
 - can_read_users
+- can_access_organisation_types
```
steveyken: I don't think we need a permission for viewing organisation_types. Anyone should be allowed to view them

spec/rails_helper.rb
steveyken: lol, yes, and I'm the one that wrote the line below! I think this is a necessary evil as it seems pretty inefficient to create a default_country for each spec.

* Need to take care of destroying the country at the end of the rspec run. See `config.after(:all)`
* If rspec exits halfway through due to an invalid syntax or ctrl-c then it won't run the `config.after(:all)` hook at all. In this case, we need to add the `initialize_with` method to the countries factory so that it doesn't try to create a duplicate country with the same name the next time you run rspec

app/models/organisation.rb
steveyken: I think I'd prefer not to upper or lowercase the names in the actual DB as that affects the display version everywhere in the apps also.

When we do searching, we can set it to case-insensitive (ILIKE) and it will be able to match the various forms.

steveyken (changes requested):

app/controllers/api/v1/gc_organisations_controller.rb
bharat619: Created a new ticket https://jira.crossroads.org.hk/browse/GCW-3292

app/serializers/api/v1/organisation_serializer.rb
bharat619: This is done

app/models/organisation.rb
bharat619: done. I have changed it

db/permissions_roles.yml
bharat619: I have removed it

spec/rails_helper.rb
bharat619: done

spec/models/organisation_spec.rb
```diff
@@ -84,4 +92,31 @@
...
     end
   end
+
+  describe '#trim_name' do
+    it 'trims and converts name to upcase before save' do
+      organisation = build(:organisation, name_en: 'good city   ')
+      organisation.save
+      expect(organisation.reload.name_en).to eq('good city')
+    end
+  end
```
patrixr, line 102: I don't think we want to upcase before saving. Since organisations are coming from external/official sources we don't want to alter the data. We can just upcase during comparison cc @steveyken

patrixr: also the script that populates the data might bypass the the hooks which would end up with non-uppercase names anyways

app/controllers/api/v1/gc_organisations_controller.rb
```diff
@@ -12,14 +13,50 @@ class GcOrganisationsController < Api::V1::ApiController
...
+
+    def update
+      @organisation.assign_attributes(organisation_params)
+      if @organisation.save
+        return render json: @organisation, include_orders_count: false, serializer: organisation_serializer
+      end
+
+      render json: { errors: @organisation.errors.full_messages }, status: 422
+    end
+
```
patrixr, line 51: I still think all of this should go into `organisations_controller` and have routes redirect `gc_organisations` to `organisations`. I'm okay keeping that for another ticket/PR though. Just mentioning :)

patrixr (approved):
Looks good to me :) 

Still having doubts about the uppercase part, maybe double check with Steve before merging

spec/models/organisation_spec.rb
bharat619: sorry about that 😅 
i need to change the test description. it doesn't convert it to upcase

swatijadhav (approved):

shreyas098 (approved):

abulsayyad123 (approved):

PR #1084, opened 2020-09-03
Add missing foreign keys

patrixr:
This PR creates a migration to add all the missing foreign keys to the database.
Most of the file changes are about fixing failing specs


spec/controllers/api/v1/authentication_controller_spec.rb
```diff
@@ -8,6 +8,7 @@
   let(:pin)    { user.most_recent_token[:otp_code] }
   let(:mobile) { generate(:mobile) }
   let(:mobile1) { generate(:mobile) }
+  let(:district_id) { create(:district).id.to_s }
```
steveyken, line 11: do we need the `to_s` bit?
```suggestion
  let(:district_id) { create(:district).id }
```

steveyken (approved):
Epi! Small tweaks but otherwise good to go. We will want to discuss the potential impact on Stockit before merging to master but I think it will be fine.

spec/controllers/api/v1/authentication_controller_spec.rb
patrixr: it was just for comparison to the json output, which comes as a string

PR #1087, opened 2020-09-10
GCW 3283: Update Roles and Printers for User

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3283

### What does this PR do?
- Assign/remove roles to user
- Assign Printer to user


app/models/user_role.rb
```diff
@@ -1,10 +1,17 @@
...
   belongs_to :role
 
   validates :role_id, uniqueness: { scope: :user_id }
+  before_save :set_expiry_date_time, if: lambda { expiry_date.present? }
 
   def self.create_user_role(user_id, role_id)
     create(user_id: user_id, role_id: role_id)
   end
+
+  def set_expiry_date_time
```
steveyken, line 14: Probably best if the column is named `expires_at` to follow rails conventions.

swatijadhav: Done @steveyken 👍

app/models/concerns/manage_user_roles.rb
```diff
@@ -18,27 +18,44 @@ module ManageUserRoles
 
   included do
 
-    def manage_roles_for_user(user, role_ids)
-      return unless self.can_update_roles_for_user?(user)
+    def assign_role_for_user(user_id: , role_id: , expires_at: nil)
+      return unless self.can_update_roles_for_user?(user_id)
```
bharat619, line 22:
```suggestion
      return unless can_update_roles_for_user?(user_id)
```

app/models/concerns/manage_user_roles.rb
```diff
@@ -18,27 +18,44 @@ module ManageUserRoles
...
 
-      allowed_role_ids = Role.allowed_roles(max_role_level)
-                             .where(id: role_ids)
-                             .pluck(:id)
-      self.update_roles_for_user(user, allowed_role_ids)
+      if accessible_role?(role_id)
+        self.assign_role(user_id, role_id, expires_at)
+      end
```
bharat619, line 26:
```suggestion
      assign_role(user_id, role_id, expires_at) if accessible_role?(role_id)
```

app/models/concerns/manage_user_roles.rb
```diff
@@ -18,27 +18,44 @@ module ManageUserRoles
...
     end
 
-    def can_update_roles_for_user?(other_user)
+    def remove_role_for_user(user_role)
+      return unless self.can_update_roles_for_user?(user_role.user_id)
+
+      if accessible_role?(user_role.role_id)
+        user_role.destroy
+      end
```
bharat619, line 34:
```suggestion
      return unless can_update_roles_for_user?(user_role.user_id)
      
      user.destroy if accessible_role?(user_role.role_id)
```

app/models/role.rb
```diff
@@ -6,7 +6,7 @@ class Role < ActiveRecord::Base
   has_many :user_roles
   has_many :users, through: :user_roles
 
-  has_many :active_user_roles, -> { where("expiry_date IS NULL OR expiry_date >= ?", Time.now.in_time_zone) },
+  has_many :active_user_roles, -> { where("expires_at IS NULL OR expires_at >= ?", Time.now.in_time_zone) },
```
bharat619, line 9:
```suggestion
  has_many :active_user_roles, -> { where("expires_at IS NULL OR expires_at >= ?", Time.current) },
```

Its always preferred to use `Time.current` instead of `Time.now`

spec/controllers/api/v1/user_roles_controller_spec.rb
```diff
@@ -36,4 +44,72 @@
       expect(parsed_body['user_role']['user_id']).to eq(user_role.user_id)
     end
   end
-end
\ No newline at end of file
+
+  describe "POST user_roles" do
+    context "Reviewer" do
```
bharat619, line 49: A test for other than `Reviewer` role would also be good

bharat619 (commented):
Nice work 🚀  Noted some minor observations

app/controllers/api/v1/user_roles_controller.rb
```diff
@@ -1,8 +1,16 @@
 module Api
   module V1
-    class UserRolesController < ApplicationController
+    class UserRolesController < Api::V1::ApiController
       load_and_authorize_resource :user_role, parent: false
 
+      def_param_group :user_role do
+        param :user_role, Hash, required: true do
+          param :user_id, String
+          param :role_id, String
```
patrixr, line 9:
```suggestion
          param :user_id, String, required: true
          param :role_id, String, required: true
```

app/models/concerns/manage_user_roles.rb
```diff
@@ -18,27 +18,44 @@ module ManageUserRoles
...
-    def update_roles_for_user(user, role_ids)
-      user.roles = Role.where(id: role_ids)
+    def assign_role(user_id, role_id, expires_at)
+      @user_role = UserRole
+          .where(user_id: user_id, role_id: role_id)
+          .first_or_initialize
+      @user_role.expires_at = expires_at
+      @user_role.save
+      @user_role
     end
```
patrixr, line 59: not so sure about a concern setting instance members `@something`. I barely approve of that in controllers haha

```suggestion
    def assign_role(user_id, role_id, expires_at)
      UserRole
          .where(user_id: user_id, role_id: role_id)
          .first_or_initialize do |user_role|
             user_role.expires_at = expires_at
             user_role.save
             user_role
          end
    end
```

spec/controllers/api/v1/user_roles_controller_spec.rb
patrixr: as well as testing as a public user. To ensure the abilities are correctly set

app/models/user_role.rb
```diff
@@ -1,10 +1,17 @@
...
   validates :role_id, uniqueness: { scope: :user_id }
+  before_save :set_expires_at_time, if: lambda { expires_at.present? }
 
   def self.create_user_role(user_id, role_id)
     create(user_id: user_id, role_id: role_id)
   end
+
+  def set_expires_at_time
+    self.expires_at = self.expires_at.change(hour: 20)
+  end
```
patrixr, line 16: We allow datetime (instead of date) for flexibility, enforcing `8pm` on all expiry dates defeats that purpose. Rather the front-end should send the requested day and hour when setting the expiry.

On the first go, the front-end will probably only allow selecting dates and default to `8pm`. But in the future we may want to add more granular controls on the UI. This would bypass anything we set from there.

app/serializers/api/v1/user_serializer.rb
```diff
@@ -7,6 +7,7 @@ class UserSerializer < ApplicationSerializer
     has_one :image, serializer: ImageSerializer
     has_one :address, serializer: AddressSerializer
     has_many :user_roles, serializer: UserRoleSerializer
+    # has_many :organisations_users, serializer: OrganisationsUserSerializer
```
patrixr, line 10: I assume this was forgotten : )

spec/controllers/api/v1/user_roles_controller_spec.rb
patrixr: Actually, rather than testing Roles, it's probably better to test permissions, and separately we can test that the correct roles have the right permissions.

Scenarios:
- Authenticated with `can_manage_user_roles`
- Authenticated without `can_manage_user_roles`
- Not authenticated

That's what drives the rules

app/controllers/api/v1/printers_users_controller.rb
```diff
@@ -19,6 +19,11 @@ class PrintersUsersController < Api::V1::ApiController
 
       api :POST, '/v1/printers_users', "Create an printers_user"
       def create
+        @printers_user = PrintersUser
+          .where(user_id: params["printers_users"]["user_id"], tag: params["printers_users"]["tag"])
+          .first_or_initialize
+        @printers_user.printer_id = params["printers_users"]["printer_id"]
+
         save_and_render_object(@printers_user)
       end
```
patrixr, line 28: Unless there's a specific reason for doing it, I recommend not setting instance members if we can avoid it.

```suggestion
      api :POST, '/v1/printers_users', "Create an printers_user"
      def create
        user_id, printer_id, tag = params["printers_users"].values_at('user_id', 'printer_id', 'tag')
        save_and_render_object PrintersUser
          .where(user_id: user_id, tag: tag)
          .first_or_initialize do |printers_user|
             printers_user.printer_id = printer_id
             save_and_render_object(printers_user)
          end
      end
```

app/controllers/api/v1/user_roles_controller.rb
```diff
@@ -12,11 +20,36 @@ def show
...
+      end
+
+      api :POST, "/v1/user_roles", "Add an user_role"
+      param_group :user_role
+      def create
+        @user_role = current_user.assign_role_for_user(
+          user_id: params["user_role"]["user_id"],
+          role_id: params["user_role"]["role_id"],
+          expires_at: params["user_role"]["expires_at"]
+        )
```
patrixr, line 35: Same as above, don't think creating an instance member is needed.

Note: you can use the `slice` method on the hash to pick multiple properties at once, e.g

```suggestion
        user_role = current_user.assign_role_for_user(
           params["user_role"].slice('user_id', 'role_id', 'expires_at')
        )
```

not a big deal, just worth mentioning : )

patrixr (changes requested):
Looks good : )

Few comments here and there but pretty minor stuff. The 8pm rule is the only thing I'm a bit weary of

app/models/concerns/manage_user_roles.rb
swatijadhav: Thanks 👍 , yep I missed updating it before. 

Passing block with `first_or_initialize` will not work in our case, as we also want to update `expires_at` value if role assigned already.
`first_or_initialize` only picks first object with matching where condition and returns it (does not do anything from the block passed). If no match found then only initializes record with attributes from `where` and given block.

```rb
      user_role = UserRole
          .where(user_id: user_id, role_id: role_id)
          .first_or_initialize
      user_role.expires_at = expires_at
      user_role.save
      user_role
```

app/serializers/api/v1/user_serializer.rb
swatijadhav: Nope, this was breaking the `current_user_profile` (api/v1/auth/current_user_profile)  response with error
`PG::DuplicateAlias: ERROR: WITH query name "organisations_user_ids_by_users" specified more than once`

steveyken (approved):
Well done!

shreyas098 (approved):

bharat619:
[The serializer used here](https://github.com/crossroads/api.goodcity/pull/1087/files#diff-6a96117366c36e46f72e2f023808047eR93) also needs to `UserDetailSerializer`
Because, on search of user, we also need the OrganisationUser details

abulsayyad123 (approved):

PR #1088, opened 2020-09-11
GCW-3280-Adding Changes for add user screen

shreyas098:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3280

### What does this PR do?

FEATURE: Tweaked changes for add user screen


app/models/user.rb
```diff
@@ -104,6 +104,10 @@ def self.find_user_by_mobile_or_email(mobile, email)
     end
   end
 
+  def app_name(user, app_name)
+    user.request_from_stock = (app_name == "stock")
```
bharat619, line 108: im not sure of this needs to be done this way, as the operation is dependent on the type of app, rather than roles/permission

cc: @abulsayyad123 @patrixr @swatijadhav

app/controllers/api/v1/users_controller.rb
```diff
@@ -29,6 +29,7 @@ def index
 
       api :POST, '/v1/users', "Create user"
       def create
+        @user.app_name(@user, app_name)
```
bharat619, line 32: what is the intent over here?

app/models/user.rb
abulsayyad123: Yup... previously we had talk on this that we have to remove dependency on `app` as much as we can.

bharat619 (commented):
I think you can add even controller test cases as well covering the scenarios, if its applicable

bharat619 (approved):

swatijadhav (approved):

PR #1089, opened 2020-09-14
GCW-3281 Add Charity Position

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3281

### What does this PR do?
Add minor API support and validations for adding charity positions


app/controllers/api/v1/organisations_users_controller.rb
```diff
@@ -47,6 +48,10 @@ def show
         render json: @organisations_user, serializer: serializer
       end
 
+      def user_status
+        render json: { status: OrganisationsUser.all_status }, status: 200
+      end
+
```
patrixr, line 54: Isn't it a bit overkill to create an endpoint just for a few strings ? Maybe using a `goodcity_setting` record is more appropriate ?

You could use an initializer to create it perhaps ?

app/models/organisations_user.rb
```diff
@@ -1,19 +1,37 @@
+# frozen_string_literal: true
+
 class OrganisationsUser < ActiveRecord::Base
   module Status
-    PENDING   = 'pending'.freeze
-    APPROVED  = 'approved'.freeze
-    EXPIRED   = 'expired'.freeze
-    DENIED    = 'denied'.freeze
```
patrixr, line 6: Any reason for removing the `freeze` ?

patrixr (commented):
PR looks pretty clean, I'm not so sure about the best way to share that status list with the frontend.

Typically we would store that list in a constants file in the app, although I can't find a good argument against fetching it from the backend.

app/controllers/api/v1/organisations_users_controller.rb
patrixr: I'm just thinking out loud. Any thoughts @steveyken  ?

app/models/organisations_user.rb
bharat619: I have added a magic comment at top `# frozen_string_literal: true` 
this handles the `.freeze` of strings : )

patrixr (approved):

shreyas098 (approved):

PR #1091, opened 2020-09-15
Hasura login + API V2 Setup

patrixr:
# API v2 Setup

- Creates the `v2` namespace
- Create `Api::v2::AuthenticationController`
  * note: Moved a lot of logic into a new `AuthenticationService`
  * new pin_jwt strategy, which returns a jwt containing the otp_auth_token
- Refactor of `models/ability.rb` into
  * `controllers/ability.rb` (base)
  * `controllers/v1/ability.rb` (v1 implementation)
  * `controllers/v2/ability.rb` (v2 implementation)
- New jsonapi serializers for users, roles, user_roles

### Default behaviour

The new serializers, by **default**, include:
- The attributes
- The relationship IDs

### Custom options

Serializers support configuration objects to customize what should and should not be included. e.g

```ruby
Api::V2::UserSerializer.new(user, { include: ["roles"], fields: { user: [:first_name, :roles], role: [:level]} })
```

Explanation of the above:
- The `user: [:first_name, :roles]` part means we want to include the `first_name` and ids of the `roles`
- The `include: ["roles"]` part means that we want to include the role relationship body
- The `role: [:level]` part means we only want the `level` field of role records

#### Query param option

The base `ApiController` contains a helper method used to parse an `?include=` param, allowing to generate the config object as seen above.

e.g

```ruby
# users_controller.rb
serializer_options(:user) # => will parse the params and return an object
```

On the query side, the following syntaxes are available:

- `/users?include=*`
  * Will include every attribute of the user records
- `/users?include=first_name,last_name`
  * Will only include the `first_name` and `last_name` attributes in the response
- `/users?include=first_name,roles`
  * Will only include the `first_name` attribute and the `roles` **ID**
- `/users?include=first_name,roles.name,roles.level`
  * Will only include the `first_name` attribute of the user and the roles' `level` and `name` attributes
- `/users?include=first_name,roles.{name,level}`
  * Will only include the `first_name` attribute of the user and the roles' `level` and `name` attributes
- `/users?include=first_name,roles.*`
  * Will only include the `first_name` attribute of the user and all the attributes of the roles

## Role impersonation

On the API V2, every request is processed as **a single role!**. By setting the `"X-GOODCITY-ROLE"` header of the request, we can access the api **as** a certain role. 

If no role is specified, the most powerful role is defaulted to.

Passing "user" will allow to behave as a no-role basic user

# Hasura Login

As an authenticated user, I can now call `/api/v2/auth/hasura` to get a JWT token which can be used against the Hasura GraphQL server


app/models/role.rb
```diff
@@ -19,4 +19,8 @@ class Role < ActiveRecord::Base
   def grant(user)
     user.roles << self unless user.roles.include?(self)
   end
+
+  def snake_name
+    name.downcase.strip.gsub(/\s+/, '_')
```
steveyken, line 24: `"String Text".parameterize.underscore` is also an option from ActiveSupport

app/services/hasura_service.rb
```diff
@@ -0,0 +1,43 @@
+class HasuraService
```
steveyken, line 1: Would a better name be `HasuraJwtService`?
Thinking in case we have other future needs for Hasura type things... hopefully not!

patrixr: I was thinking those other things can also go into that service, basically everything related to Hasura goes here 🤔

lib/warden/strategies/pin_strategy.rb
```diff
@@ -4,23 +4,37 @@ module Warden
...
+        request_params[:pin_for]&.to_sym || :mobile
+      end
+
+      def authenticate!        
+        auth_token = lookup_auth_token
         return success!(auth_token.user) if valid_app_store_credentials?(auth_token)
         user = auth_token.try(:user)
-        has_valid_otp_code?(auth_token) && valid_user(user) ? success!(user) : fail
+        return fail unless has_valid_otp_code?(auth_token) && valid_user(user)
+        user.set_verified_flag(pin_method) if pin_method.present?
```
steveyken, line 23: don't we want to get the `pin_method` from the JWT token?

patrixr: @steveyken That's the V1 implementation, look at `pin_jwt_strategy.rb`

patrixr: I inherited this one, and made some changes

abulsayyad123 (approved):

app/models/role.rb
```diff
@@ -6,8 +6,7 @@ class Role < ActiveRecord::Base
   has_many :user_roles
   has_many :users, through: :user_roles
 
-  has_many :active_user_roles, -> { where("expiry_date IS NULL OR expiry_date >= ?", Time.now.in_time_zone) },
-            class_name: "UserRole"
+  has_many :active_user_roles, -> { where("expiry_date IS NULL OR expiry_date >= ?", Time.now.in_time_zone) }, class_name: "UserRole"
```
bharat619, line 9:
```suggestion
  has_many :active_user_roles, -> { where("expiry_date IS NULL OR expiry_date >= ?", Time.current) }, class_name: "UserRole"
```

The difference of `.current` and `.now` is `.now` use the server's timezone, while .current use what the Rails environment is set to. If it's not set, then .current will be same as .now

app/models/token.rb
```diff
@@ -10,39 +10,81 @@ class Token
 
   validate :token_validation
 
+  attr_reader :jwt_config
+
+  GC_METADATA_NAMESPACE = "https://goodcity.hk/jwt/metadata"
+
+  module Types
+    API = "api".freeze      # An 'api' token is used as a mean of resolving the current user when connecting to the REST API
```
bharat619, line 18: You can remove the `.freeze` from all string literals
and add magik comment at top level
`#frozen_string_literal: true`

app/controllers/ability.rb
```diff
@@ -0,0 +1,52 @@
+class Ability
```
bharat619, line 1: Just a curious question
Why are we moving abilities from model to a controller?

bharat619 (commented):
Nice. Few minor observation and one question

steveyken (approved):

app/models/role.rb
patrixr: I haven't changed anything on this line

app/controllers/ability.rb
patrixr: Because we need an ability file _per api version_

Models are not versioned, so it wouldn't work

app/models/token.rb
patrixr: Having code functionalities in comments is a ticking time bomb imo... no clue why this was added in ruby 🤔

PR #1093, opened 2020-09-16
GCW-3284 Add Registration Uniqueness Validation

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3284


patrixr (approved):

steveyken (approved):
Thanks! Does this allow nil / blank by default? That would be helpful.
Does it treat upper and lower case as a clash?

bharat619:
This does allow nil/blank value.
And its case sensitive and treats uppercase and lowercase as different. 
Should it be made case insensitive?

PR #1109, opened 2020-10-06
GCW 3312 Improve Box Pallets

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3312

### What does this PR do?
Adds additional counter cache columns to specify the count of package in a box or pallet


app/models/concerns/inventory_computer.rb
```diff
@@ -5,6 +5,8 @@ module InventoryComputer
   # Adds a sub module containing all the needed SQL computations
   #
   class Computer
+    PACK_UNPACK = [PackagesInventory::Actions::PACK, PackagesInventory::Actions::UNPACK].freeze
+    DISPATCH_UNDISPATCH = [PackagesInventory::Actions::DISPATCH, PackagesInventory::Actions::UNDISPATCH].freeze
```
patrixr, line 9:
```suggestion
    PACK_UNPACK_ACTIONS         = [PackagesInventory::Actions::PACK, PackagesInventory::Actions::UNPACK].freeze
    DISPATCH_UNDISPATCH_ACTIONS = [PackagesInventory::Actions::DISPATCH, PackagesInventory::Actions::UNDISPATCH].freeze
```

app/models/concerns/inventory_computer.rb
```diff
@@ -75,13 +76,25 @@ def total_quantity
...
+      def on_hand_packed_quantities_for(package)
+        res = historical_quantity.joins("INNER JOIN packages on packages.id = packages_inventories.source_id and packages_inventories.source_type = 'Package'")
+        res = res.select('packages_inventories.*')
+        res = res.where(packages_inventories: { package_id: package.id,
+                                                action: PACK_UNPACK })
+        res = res.where.not(packages: { on_hand_quantity: 0 })
+                 .group('packages.storage_type_id').raw.sum(:quantity)
+        res = storage_type.group_by(res, :name)
+        res = box_pallet_on_hand_qty_hash(res)
+        res
```
patrixr, line 88: no point in using `historical_quantity` if you do a `raw` and `sum` after. You can just use a normal query

app/models/concerns/inventory_computer.rb
```diff
@@ -116,6 +145,10 @@ def of(model)
         where("#{model.class.name.underscore}_id = (?)", Utils.to_id(model))
       end
 
+      def group_by(data, attr)
+        data.map { |id, value| { @model.find(id).try(attr).to_s => value } }
+      end
```
patrixr, line 150: don't think this has any place in this class, it's a util

app/models/concerns/inventory_computer.rb
```diff
@@ -90,6 +103,18 @@ def update_package_quantities!(package)
           package.save!
         end
       end
+
+      private
+
+      def box_pallet_on_hand_qty_hash(value)
```
patrixr, line 109: what's `value` here ?

spec/controllers/api/v1/packages_controller_spec.rb
```diff
@@ -633,6 +633,20 @@ def test_packages_location_changes(package)
         expect(parsed_body["package"]["id"]).to eq(package.id)
         expect(parsed_body["package"]["storage_type_id"]).to eq(package.storage_type_id)
       end
+
+      it 'has 0 on_hand_boxed_quantity' do
```
patrixr, line 637:
```suggestion
      it 'has 0 on_hand_boxed_quantity by default' do
```

spec/controllers/api/v1/packages_controller_spec.rb
```diff
@@ -633,6 +633,20 @@ def test_packages_location_changes(package)
...
+        package_params[:storage_type] = 'Box'
+        package_params[:received_quantity] = 1
+        post :create, params: { package: package_params }
+        expect(parsed_body['package']['on_hand_boxed_quantity']).to eq(0)
+      end
+
+      it 'has 0 on_hand_palletized_quantity' do
```
patrixr, line 644:
```suggestion
      it 'has 0 on_hand_palletized_quantity by default' do
```

spec/controllers/api/v1/packages_controller_spec.rb
```diff
@@ -1596,41 +1611,207 @@ def unpack(qty, pkg, out_of:)
...
         }
+
         @params7 = {
           id: box.id,
           item_id: package2.id,
           location_id: location.id,
           task: "pack",
           quantity: package2.on_hand_quantity + 20,
         }
+
```
patrixr, line 1622: I think we should be using `let` over instance variables for tests.

Also I think it's better to keep a single param sample that you can override in your tests. Because we don't know where each param is used for.

Or if you prefer having multiple params, maybe they should be named explicitly, e.g `let(:add_package_to_pallet_params) { ... }`

note: I know this pre-existed your changes, just pointing it out :)

spec/controllers/api/v1/packages_controller_spec.rb
```diff
@@ -1596,41 +1611,207 @@ def unpack(qty, pkg, out_of:)
...
+          before do
+            Package::Operations.pack_or_unpack(container: pallet, package: package1, quantity: 3, location_id: location.id, user_id: user.id, task: 'pack')
+
+            Package::Operations.pack_or_unpack(container: box, package: package1, quantity: 5, location_id: location.id, user_id: user.id, task: 'pack')
+          end
+
+          context 'on removing few items from box' do
+            before do
+              @params1[:task] = 'unpack'
+              @params1[:quantity] = 2
```
patrixr, line 1742: Kind of the same comment as above, you can use a `let` within the context to define what the variables look like

app/models/concerns/inventory_computer.rb
```diff
@@ -17,6 +19,10 @@ def historical_quantity
         SumAsOf.new(PackagesInventory)
       end
 
+      def storage_type
+        SumAsOf.new(StorageType)
```
patrixr, line 23: `StorageType` cannot be summed, so this doesn't make sense 🤔

app/models/concerns/inventory_computer.rb
patrixr: FYI: SumAsOf is a numeric, meaning it's like a number and supports all those operations. But it's computed on the fly when you call it.

Allowing you to do things like

```ruby
package = Package.first
package_quantity = SumAsOf.new(PackagesInventory).of(package)
puts(package_quantity) # => 10
puts(package_quantity + 1) # => 11
puts(package_quantity.as_of(1.month.ago)) #=> 5
puts(package_quantity.as_of_now) #=> 10
```

patrixr: You can look at the base class `ComputableNumeric` in `computable_numeric.rb`

bharat619: yup thanks. I moved it out of the class. I had second thoughts of keeping in that class, because the object was of StorageType, which didnt make much sense as a computation.
It Should be clear now

spec/controllers/api/v1/packages_controller_spec.rb
bharat619: done. refactored it : )

app/models/concerns/inventory_computer.rb
bharat619: yeah. corrected it as mentioned in above comment

app/models/concerns/inventory_computer.rb
bharat619: changed it to a better name now. have a look

patrixr (approved):

abulsayyad123 (approved):

PR #1112, opened 2020-10-12
Fix: unlink beneficiary from order before deletion

patrixr:
### Changes

- orders.beneficiary_id nulled before beneficiary is deleted
- restored some commented tests
- api v2 role specification: empty role defaults to all role behaviour
- added a default error message for all foreign key related issue
- prevent frontend from setting `stockit_organisation_id` as it caused broken foreign keys

### DIFF
--- a/spec/controllers/api/v1/api_controller_spec.rb
+++ b/spec/controllers/api/v1/api_controller_spec.rb
@@ -11,12 +11,31 @@ RSpec.describe Api::V1::ApiController, type: :controller do
       end
     end
 
-    # TODO: Fix tests for 4XX status
-    # it do
-    #   get :index
-    #   expect(response.status).to eq(404)
-    #   expect(subject).to eql( {"error"=>"Oh noes !"} )
-    # end
+    it do
+      get :index
+      expect(response.status).to eq(404)
+      expect(subject).to eql( {"error"=>"Oh noes !"} )
+    end
+  end
+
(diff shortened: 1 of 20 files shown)


spec/controllers/api/v1/api_controller_spec.rb
```diff
@@ -11,12 +11,31 @@ def index
...
+    controller do
+      def index
+        raise PG::ForeignKeyViolation
+      end
+    end
+
+    it do
+      get :index, format: 'json'
+      expect(response.status).to eq(409)
+      expect(subject['error']).to eq('A broken entity relationship has occurred')
```
bharat619, line 31: can read from locales

spec/controllers/api/v1/appointment_slot_presets_controller_spec.rb
```diff
@@ -43,16 +42,15 @@
 
   describe "POST /appointment_slot_presets" do
     context 'When not logged in' do
-      # TODO: Fix tests for 4XX status
-      # it "denies creation of a default appointment slot" do
-      #   post :create, appointment_slot_preset: payload
-      #   expect(response.status).to eq(401)
-      # end
+      it "denies creation of a default appointment slot" do
```
bharat619, line 45: 🔥

bharat619 (approved):
Tests are working. Thanks not to Warden 😬

shreyas098 (approved):

PR #1114, opened 2020-10-13
GCW-3236-Added changes for international order creation

shreyas098:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3236

### What does this PR do?

FEATURE: Changes for international orders


app/models/order.rb
```diff
@@ -128,6 +131,15 @@ def delete_orders_packages
     end
   end
 
+  def international_orders?
+    detail_type == "Shipment" || detail_type == "CarryOut"
```
bharat619, line 135:
```suggestion
    [Type::SHIPMENT, Type::CARRY_OUT].includes?(detail_type)
```

and ORDER_TYPE can be a module in the order model something like
```
   module Type
      SHIPMENT = 'Shipment'
      CARRY_OUT = 'CarryOut'
   end
```

bharat619: So u can make use of those type anywhere in the application just by doing
`Order::Type::SHIPMENT` or `Order::Type::CARRY_OUT`

bharat619 (commented):
A minor improvement suggested on the constants

app/controllers/api/v1/orders_controller.rb
```diff
@@ -73,6 +73,16 @@ def transition
         render json: serializer.new(@order, include_allowed_actions: true)
       end
 
+      def fetch_shipment_or_carryout_code
+        record = Order.get_subsequent_international_code(params['detail_type'])
+
+        if record <= 99_999
```
bharat619, line 79: can move to either constants.rb, or in the module

app/controllers/api/v1/orders_controller.rb
```diff
@@ -73,6 +73,16 @@ def transition
...
       end
 
+      def fetch_shipment_or_carryout_code
+        record = Order.get_subsequent_international_code(params['detail_type'])
+
+        if record <= Order::Type::MAX_INTERNATIONAL_ORDER_CODE
+          render json: { code: record }
+        else
+          render json: { errors: "Code Limit Exhausted" }, status: 404
+        end
```
bharat619, line 83: No need to do this check here. You can rewrite the method somewhat like this

```suggestion
        code = Order.get_next_code_for(params['detail_type'])

        render json: { code: record }
```
render error in case there is an exception or if code is nil/invalid

app/models/order.rb
```diff
@@ -1,4 +1,10 @@
 class Order < ApplicationRecord
+  module Type
+    SHIPMENT = 'Shipment'.freeze
+    CARRY_OUT = 'CarryOut'.freeze
+    MAX_INTERNATIONAL_ORDER_CODE = 99_999
```
bharat619, line 5: I dont think this constant is needed at all
There is no need to maintain a limit to order code
This will be limiting the number of shipment/carryout orders that can be created in the long run

app/models/order.rb
```diff
@@ -128,6 +136,20 @@ def delete_orders_packages
...
+
+  def self.get_subsequent_international_code(params)
+    record = Order.where(detail_type: params).order("id desc").pluck(:code)
+    unless record.empty?
+      susequent_value = record.map{|s| s.gsub(/\D/, "")}
+      susequent_value.map!(&:to_i).max + 1
+    else
+      1000
+    end
+  end
```
bharat619, line 151:
```suggestion
# missing_code will give the next code
# ex: if 0001 and 0004 exists, missing_code will give 2, which will be converted to 0002
# If there is no missing_code, then next_code will give 0005
def self.get_next_code_for(type)
    code = missing_code_for(type) || next_code(type) + 1
    code.to_s.rjust(5, '0')
  end

  def self.missing_code_for(type)
    query = <<-query
      select s.i as first_missing_code from generate_series(1, :max) s(i) WHERE s.i not in (
          select CAST(substring(orders.code, 2) as INTEGER) as code from orders where orders.detail_type = :type and substring(orders.code, 2) ~ '^\\d+$'
      ) limit 1
    query

    code = connection.exec_query(sanitize_sql_array([query, type: type, max: Order.where(detail_type: type).count])).first['first_missing_code']
    code
  end

  def self.next_code(type)
    query = <<-query
    SELECT max(CAST(substring(orders.code, 2) as INTEGER)) as code from orders where orders.detail_type = :type and substring(orders.code, 2) ~ '^\d+$'
    query

    code = connection.exec_query(sanitize_sql_array([query, type: type])).first["code"]
    code
  end
```

bharat619: You can refactor the above
It has repetitive queries, you can extract them to separate method if possible

Also you can move to a separate concern or lib : )

bharat619 (changes requested):

app/models/order.rb
patrixr: we probably want to consolidate the generation of shipment codes with the one of goodcity orders

There's an existing `generate_gc_code` method. We should not merge this PR having 2 separate entry points for generating codes.

We can probably end with a single entry point, maybe like `generate_code(detail_type)`

shreyas098: There is a toggle of Shipment and Carryout in the shipment form so thought of doing it with different entry, will do it with generate_code(detail_type)

steveyken (changes requested):
A few comments below and I think you are talking with @patrixr Would be good to take a slightly different approach. Great first take though

app/controllers/api/v1/ability.rb
```diff
@@ -176,51 +176,51 @@ def item_abilities
...
-    
+
       def order_abilities
-        can :create, Order
+        can %i[create fetch_shipment_or_carryout_code], Order
         can %i[index show transition], Order, created_by_id: @user_id
         can %i[update destroy], Order, created_by_id: @user_id, state: %w[draft submitted processing awaiting_dispatch]
         if can_manage_orders? || @api_user
-          can %i[create index show update transition summary], Order
+          can %i[create index show update transition summary fetch_shipment_or_carryout_code], Order
```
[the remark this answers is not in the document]
shreyas098: sure

app/controllers/api/v1/orders_controller.rb
```diff
@@ -73,6 +73,12 @@ def transition
         render json: serializer.new(@order, include_allowed_actions: true)
       end
 
+      # GET /orders/next_code?detail_type=
+      def next_code
+        record = Order.generate_gc_code(params['detail_type'])
```
patrixr, line 78: do validate the `detail_type` param before running this

app/models/order.rb
```diff
@@ -65,6 +68,8 @@ class Order < ApplicationRecord
 
   NON_PROCESSED_STATES = ["processing", "submitted", "draft"].freeze
 
+  INTERNATIONAL_ORDERS = ["Shipment", "CarryOut"].freeze
```
patrixr, line 71: We can remove the "international orders" term completely. We don't really want that to be a thing. All orders should be processed via the same flow. You can unify the 2 validations into 1, maybe with something like this:

```ruby
  validate :verify_code_format

  def verify_code_format
    prefix = order_code_prefix(self.detail_type)
    errors.add(:base, 'Invalid code format') unless /\A#{prefix}\d{4,5}([A-Z]{1})?\z/.match(self.code)
  end
```

app/models/order.rb
```diff
@@ -485,7 +511,7 @@ def email_properties
   private
 
   def assign_code
-    self.code = Order.generate_gc_code if goodcity_order?
+    self.code = Order.generate_gc_code(detail_type) if goodcity_order?
```
patrixr, line 514: You can rename that method into `generate_code`, since it no longer deals exclusively with gc orders

spec/controllers/api/v1/orders_controller_spec.rb
```diff
@@ -631,6 +631,34 @@ def create_order_with_transport(state)
     end
   end
 
+  describe 'Order Next Code' do
```
patrixr, line 634: You're testing "Shipments" and "CarryOuts" but not "GoodCity" orders.
All types should get the same treatment : )

patrixr: Also, note that in the "CarryOut" tests, you're testing an (potentially) erroneous case but not a normal expected case. And you're doing the opposite in "Shipment"

Make sure you test for all potential scenarios, for all types : )

spec/controllers/api/v1/orders_controller_spec.rb
```diff
@@ -631,6 +631,42 @@ def create_order_with_transport(state)
...
```
bharat619, line 634:
```suggestion
  describe 'GET next_code' do
```

app/controllers/api/v1/orders_controller.rb
```diff
@@ -73,6 +73,12 @@ def transition
         render json: serializer.new(@order, include_allowed_actions: true)
       end
 
+      # GET /orders/next_code?detail_type="Shipment"
+      def next_code
+        record = Order.generate_code(params['detail_type'])
+        render json: { code: record }
+      end
```
steveyken, line 80: 👍

app/controllers/api/v1/orders_controller.rb
```diff
@@ -23,7 +23,7 @@ class OrdersController < Api::V1::ApiController
           param :stockit_id, String, desc: "stockit designation record id", allow_nil: true
           param :beneficiary_id, String, allow_nil: true
           param :address_id, String
-          param :booking_type_id, String, desc: 'Booking type.(Online order or appointment)'
+          param :booking_type_id, String, desc: 'Booking type.(Online order or appointment)', allow_nil: true
```
steveyken, line 26: Is there a reason we need `allow_nil`?

app/models/order.rb
```diff
@@ -429,13 +431,35 @@ def self.orders_count_per_state(orders)
...
+    prefix = order_code_prefix(detail_type)
+    splitter = prefix.length.to_i+1
+    reg = %r/^\d+$/
+
+    code = next_code(detail_type, splitter, reg) + 1
+    code = code.to_s.rjust(5, '0')
+    "#{prefix}#{code}"
+  end
+
+  def self.next_code(type, splitter, reg)
```
steveyken, line 451:
```suggestion
  def self.next_code(detail_type, splitter, reg)
```

Avoid usage of the keyword `type`

app/models/order.rb
```diff
@@ -429,13 +431,35 @@ def self.orders_count_per_state(orders)
     orders.each { |key, value| orders[key] = value.count }
   end
 
-  def self.generate_gc_code
-    record = where(detail_type: "GoodCity").order("id desc").first
-    "GC-" + gc_code(record).to_s.rjust(5, "0")
+  def self.order_code_prefix(detail_type)
```
steveyken, line 434: This is good. Can you put it in a concern? This code has a single responsibility and would fit neatly inside a concern where all relevant methods can be kept together.

steveyken (commented):
This is much better. One main change and a query. Good job!

app/models/order.rb
```diff
@@ -429,21 +446,12 @@ def self.orders_count_per_state(orders)
...
-
-  def goodcity_order?
-    detail_type == "GoodCity"
+  def valid_order?
+    [DetailType::SHIPMENT, DetailType::GOODCITY, DetailType::CARRYOUT].include? detail_type
   end
 
-  def draft_goodcity_order?
-    state == "draft" && goodcity_order?
+  def draft_valid_order?
```
swatijadhav, line 453: `valid_draft_order?` looks more readable.

app/controllers/api/v1/orders_controller.rb
bharat619: I think this is because `booking_type_id` is not required for international orders

cc: @shreyas098

bharat619 (approved):

app/controllers/api/v1/orders_controller.rb
shreyas098: Yes

bharat619 (approved):

patrixr (approved):
I think it's good to merge, just make sure to include those 2 test scenarios 

```ruby

        it "fails to create a code for invalid detail types" do
          get :next_code, params: { detail_type: detail_type }
          expect(response.status).to eq(422)
          expect(parsed_body['error']).to match(/Invalid or missing detail_type/)
        end
        it "fails to create a code for missing detail types" do
          get :next_code, params: {}
          expect(response.status).to eq(422)
          expect(parsed_body['error']).to match(/Invalid or missing detail_type/)
        end
```

app/models/concerns/order_code_generator.rb
```diff
@@ -0,0 +1,67 @@
...
+              orders.detail_type = :detail_type and SUBSTRING(orders.code, :delimiter) ~ :matcher
+          ) limit 1
+      QUERY
+      code = exec_query(query, detail_type: detail_type, matcher: matcher.source,
+                                 max: Order.where(detail_type: detail_type).count, delimiter: delimiter)
+      code.first.present? ? code.first['first_missing_code'] : nil
+    end
+
+    def exec_query(query, params)
+      klass.connection.exec_query(klass.sanitize_sql_array([query, params]))
```
steveyken, line 64: Full marks for sanitizing the query. Calling the methods directly on `ActiveRecord::Base` means you don't need to pass in or maintain an instance variable for `klass`

```suggestion
      sanitized_query = ActiveRecord::Base.send(:sanitize_sql_array, [query, params])
      ActiveRecord::Base.connection.exec_query(sanitized_query)
```

app/models/order.rb
```diff
@@ -13,6 +16,12 @@ class Order < ApplicationRecord
     ]
   end
 
+  module DetailType
+    SHIPMENT = 'Shipment'
```
steveyken, line 20: This is starting to feel like it should be a taxonomy table now? cc @patrixr  for thoughts

app/models/token.rb
```diff
@@ -107,7 +107,7 @@ def token_validation
       return errors.add(:base, I18n.t('token.invalid')) unless ALL_TYPES.include?(access_type)
       if exp_time < cur_time
         errors.add(:base, I18n.t('token.expired'))
-      elsif !(iat_time < cur_time && iat_time < exp_time)
+      elsif !(iat_time <= cur_time && iat_time < exp_time)
```
steveyken, line 110: why this change?

spec/models/order_spec.rb
```diff
@@ -724,11 +724,128 @@
...
     let(:order) { create(:order, :with_state_draft, :with_orders_packages) }
 
-    it "Assigns GC Code" do
-      expect(order.code).to include("GC-")
+    context 'Goodcity orders' do
+      it "Assigns GC Code" do
+        expect(order.code).to include('GC-')
+      end
+    end
+
```
steveyken, line 735: This is good testing but the spec file is super long. Can you extract these out into `spec/models/concerns/order_code_generator_spec.rb`?

app/models/order.rb
```diff
@@ -429,21 +446,12 @@ def self.orders_count_per_state(orders)
...
   end
 
-  def self.generate_gc_code
-    record = where(detail_type: "GoodCity").order("id desc").first
-    "GC-" + gc_code(record).to_s.rjust(5, "0")
-  end
-
-  def self.gc_code(record)
-    record ? record.code.gsub(/\D/, "").to_i + 1 : 1
+  def valid_order?
```
steveyken, line 449: Rename to

```suggestion
  def valid_detail_type?
```

app/models/order.rb
```diff
@@ -497,4 +507,24 @@ def order
   def offer
     nil
   end
+
+  def validate_shipment_date
+    is_valid = shipment_date >= Date.current
```
steveyken, line 512: This will mean we can never update shipment that have been sent in the past. I think it may be best to remove this validation entirely or just leave it on 'create' and not 'update'.

steveyken (changes requested):
Great work @shreyas098 - definitely going in the right direction. a few tweaks below then we'll be good to go

app/models/order.rb
shreyas098: Sure will just leave it on create

PR #1128, opened 2020-10-19
GCW-3301- Selecting default printer while doing `print_label` command

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3301

### What does this PR do?

This PR accepts `tag` params from app and print label on favourite printer as per tag params.

### DIFF
--- a/app/controllers/api/v1/packages_controller.rb
+++ b/app/controllers/api/v1/packages_controller.rb
@@ -212,7 +212,7 @@ module Api
       end
 
       def print_inventory_label
-        printer = PrintersUser.where(user_id: User.current_user.id, tag: params[:tag]).first.try(:printer)
+        printer = PrintersUser.where(user_id: current_user.id, tag: params[:tag]).first.try(:printer)
         return render json: { errors: I18n.t("package.printer_not_found") }, status: 400 unless printer
         opts = {
           print_count: print_count,
--- a/app/controllers/api/v1/packages_controller.rb
+++ b/app/controllers/api/v1/packages_controller.rb
@@ -212,7 +212,12 @@ module Api
       end
 
       def print_inventory_label
-        PrintLabelJob.perform_later(@package.id, User.current_user.id, "inventory_label", print_count)
+        opts = {
+          print_count: print_count,
+          label: "inventory_label",
+          tag: params[:tag]
+        }
+        PrintLabelJob.perform_later(@package.id, User.current_user.id, opts)
         render json: {}, status: 204
       end
 
--- a/app/controllers/api/v1/packages_controller.rb
+++ b/app/controllers/api/v1/packages_controller.rb
@@ -212,12 +212,13 @@ module Api
       end
 
       def print_inventory_label
+        printer = PrintersUser.where(user_id: User.current_user.id, tag: params[:tag]).first.try(:printer)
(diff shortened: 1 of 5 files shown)


swatijadhav (approved):

shreyas098 (approved):

app/jobs/print_label_job.rb
```diff
@@ -1,9 +1,9 @@
...
-  def perform(package_id, current_user_id, label_type, print_count)
+  def perform(package_id, current_user_id, opts={})
     package = Package.find(package_id)
-    options = { inventory_number: package.inventory_number, print_count: print_count }
-    label = "Labels::#{label_type.classify}".safe_constantize.new(options)
-    printer = User.find_by_id(current_user_id).try(:printers).first
+    options = { inventory_number: package.inventory_number, print_count: opts[:print_count] }
+    label = "Labels::#{opts[:label_type].classify}".safe_constantize.new(options)
+    printer = PrintersUser.where(user_id: current_user_id, tag: opts[:tag]).first.try(:printer)
     PrintLabel.new(printer, label).print
```
patrixr, line 7: this is at risk of breaking since printer can be nil

app/controllers/api/v1/packages_controller.rb
```diff
@@ -212,7 +212,12 @@ def package_valuation
       end
 
       def print_inventory_label
-        PrintLabelJob.perform_later(@package.id, User.current_user.id, "inventory_label", print_count)
+        opts = {
```
patrixr, line 215: It's probably better to lookup the printer at this step. That way if the printer isn't set for this tag, you can return an error.

abulsayyad123: ohh.. thanks for pointing it out, will add check for this.

app/controllers/api/v1/packages_controller.rb
```diff
@@ -212,7 +212,13 @@ def package_valuation
       end
 
       def print_inventory_label
-        PrintLabelJob.perform_later(@package.id, User.current_user.id, "inventory_label", print_count)
+        printer = PrintersUser.where(user_id: User.current_user.id, tag: params[:tag]).first.try(:printer)
```
bharat619, line 215:
```suggestion
        printer = PrintersUser.where(user_id: current_user.id, tag: params[:tag]).first.try(:printer)
```

app/controllers/api/v1/packages_controller.rb
```diff
@@ -212,7 +212,13 @@ def package_valuation
...
+        return render json: { errors: I18n.t("package.printer_not_found") }, status: 400 unless printer
+        opts = {
+          print_count: print_count,
+          label: "inventory_label"
+        }
+        PrintLabelJob.perform_later(@package.id, printer.id, opts)
```
bharat619, line 221: There is already an existing `print_inventory_label` method in the `packages_controller.rb`, probably you can enhance that to take in arguments

or even better move it to a model method
```
# from controller
printer.print_inventory_label(opts)

# in model 
def print_inventory_label
 PrintLabelJob.perform_later(slef.id, opts[:user_id], "inventory_label", opts[:print_count])
end
```

bharat619 (approved):

patrixr (approved):

PR #1130, opened 2020-10-22
GCW-3376 Fix User Charity Search

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3376

### What does this PR do?
Fix the search of charity users - Instead of searching on the basis of role name, performs search on the approved or pending organisation status

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -171,6 +171,14 @@ class User < ApplicationRecord
       .order("orders.id DESC").limit(5)
   end
 
+  def self.search_by_role(name)
+    if name == 'Charity'
+      joins(:organisations_users).where(organisations_users: { status: [OrganisationsUser::ACTIVE_STATUS] })
+    else
+      with_roles(name)
+    end
+  end
+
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -171,6 +171,14 @@ def self.recent_orders_created_for(user_id)
       .order("orders.id DESC").limit(5)
   end
 
+  def self.search_by_role(name)
```
patrixr, line 174: I don't think it's good to keep 'Charity' as a valid role input. Rather we should filter on users with organisations with a query param which indicates so.

But since we want to fix this asap, this will work fine. Can refactor later :)

patrixr (approved):

swatijadhav (approved):

app/models/user.rb
steveyken: agreed, have been working with @bharat619 on a simliar solution. We'll release this as a hotfix and then a follow-up ticket to use `organisations_users` state filter properly

PR #1141, opened 2020-10-29
GCW-3381 Support Package Description Translation

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3381

### What does this PR do?
The description field will now have two states (on English and one Chinese) in create, edit and view modes.


spec/controllers/api/v1/packages_controller_spec.rb
```diff
@@ -649,6 +649,30 @@ def test_packages_location_changes(package)
...
+
+      it 'creates package with notes in en and zh_tw languages' do
+        package_params[:notes] = en_note
+        package_params[:notes_zh_tw] = zh_note
+        post :create, params: { package: package_params }
+        expect(parsed_body['package']['notes']).to eq(en_note)
+        expect(parsed_body['package']['notes_zh_tw']).to eq(zh_note)
+      end
+
+      context 'if english note is empty' do
```
patrixr, line 664: we probably want to test the flip side of things as well

```ruby
      context 'if only the zh_tw note is empty' do
          it 'creates the package successfully' do
          end
      end
```

Not a big deal, but as a habit we want to always test the different permutations

patrixr (approved):

app/models/package.rb
```diff
@@ -60,6 +60,7 @@ class Package < ApplicationRecord
   end
 
   validates :package_type_id, presence: true
+  validates :notes, presence: true
```
steveyken, line 63: Migration note: there are 413 packages where `notes = NULL` and 84 packages `where notes = ''` so we will need to fill these in before migrating. Easiest option is to copy across the code description. Can be done in SQL.

Can you add a postgresql constraint as well?

steveyken (approved):
Approved, added a migration note

app/models/package.rb
bharat619: Sure. I have added it

spec/controllers/api/v1/packages_controller_spec.rb
bharat619: sure

steveyken (approved):

PR #1149, opened 2020-11-04
GCW-3285- transition offer from Received to Receiving

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3285

### What does this PR do?

This PR adds "resume_receiving" action to transition offer from Received to Receiving state.

### DIFF
--- a/app/controllers/api/v1/offers_controller.rb
+++ b/app/controllers/api/v1/offers_controller.rb
@@ -146,6 +146,12 @@ module Api
         render json: @offer, serializer: offer_serializer
       end
 
+      api :PUT, '/v1/offers/1/resume_receiving', "Change Offer to Receiving state"
+      def resume_receiving
+        @offer.update_attributes({ state_event: 'start_receiving' })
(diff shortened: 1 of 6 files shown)


bharat619 (approved):

app/controllers/api/v1/offers_controller.rb
```diff
@@ -146,6 +146,12 @@ def reopen_offer
         render json: @offer, serializer: offer_serializer
       end
 
+      api :PUT, '/v1/offers/1/resume_receiving', "Change Offer to Receiving state"
+      def resume_receiving
+        @offer.update_attributes({ state_event: 'start_receiving' })
```
patrixr, line 151: you can probably do `@offer.start_receiving`

Also, on which initial states can you apply this transition ? I assume you can't do that from any states.
If that's the case, make sure there are tests to prove you can't do this for any offer in any state

patrixr (approved):
Looks good, please check the one comment I left though

patrixr (approved):

PR #1150, opened 2020-11-05
GCW-3373 Removed stockit-id reference, stockit sync jobs, stockit-request based code changes

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3373

### What does this PR do?
Removed 
- stockit_id column from all tables
- stockit sync jobs,
- stockit-request based code changes
- STOCKIT_ENABLED check
- removed health-check code


app/controllers/api/v1/packages_controller.rb
```diff
@@ -1,7 +1,6 @@
 module Api
```
patrixr, line 1: Great work !

**Out of scope for this PR:**

I do think this controller needs some extra refactoring

The problem is mostly due to the fact that both _updates_ and _creates_ use the same entry point. Which creates really convoluted code full of potential issues.
It also checks if an inventory number is present, and creates the PackagesInventory's "inventory" row when it detects that the inventory number has been set. We should probably have an `/packages/123/inventory` endpoint to make the whole thing more explicit (just thinking out loud)

Any volunteers ? 🤣 

@swatijadhav @abulsayyad123 @bharat619 @shreyas098 @steveyken

app/controllers/api/v1/stockit_activities_controller.rb
```diff
@@ -14,7 +14,6 @@ class StockitActivitiesController < Api::V1::ApiController
       def_param_group :stockit_activity do
```
patrixr, line 14: not sure what a "stockit_activity" is, but can't we remove the whole thing ?

patrixr (approved):
Loving this ❤️ !

My thoughts on following stockit actions:

- Refactor the packages_controller (wrote another comment on this above ☝️ )
- Remove stockit tables we don't need (stockit_activities, stockit_contacts, stockit_organisations, ...)
- Make PackagesLocation a readonly table (writes should now only come through the inventory)

I'm in full support of anyone who wants to pickup work mentioned above :)

app/controllers/api/v1/packages_controller.rb
swatijadhav: Yep, let's do it. Good to add a ticket for this and pick in the next sprint.

steveyken (approved):
Epic work. Approved :)

app/models/item.rb
```diff
@@ -28,13 +28,6 @@ class Item < ApplicationRecord
   scope :accepted, -> { where("state = 'accepted'") }
   scope :donor_items, ->(donor_id) { joins(:offer).where(offers: { created_by_id: donor_id }) }
 
-  # Workaround to set initial state fror the state_machine
```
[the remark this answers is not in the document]
swatijadhav: That was old commented out code, so removed it. I think that code was used before rails 5.

PR #1166, opened 2020-12-01
GCW-3388 Valid Stock App Mentions

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3388

### What does this PR do?
Allow valid users to be mentioned in chats


shreyas098 (approved):

app/models/concerns/mentionable.rb
```diff
@@ -46,4 +46,22 @@ def parse_id_from_mention_text
...
     ids.map { |id| id.match(/\d+/).to_s }
   end
+
+  included do
+    def self.mentionable_users(opts = {})
+      users = active.exclude_user(User.current_user.id)
+                    .with_roles(mentionable_roles(opts[:roles])).distinct.to_a
+      users << add_order_creator(users, opts[:order_id]) if opts[:order_id]
+      users.compact
+    end
```
patrixr, line 56: It's usually preferable to use ruby options instead of a hash, that way it's much clearer
```ruby
    def self.mentionable_users(roles:, order_id: nil)
      users = active.exclude_user(User.current_user.id)
                    .with_roles(mentionable_roles(roles)).distinct.to_a
      users << add_order_creator(users, order_id) if order_id
      users.compact
    end
```

Not a big deal, but now we know what the parameters are and which ones are optional by looking at the function signature

patrixr (approved):

app/models/concerns/mentionable.rb
```diff
@@ -46,4 +46,22 @@ def parse_id_from_mention_text
...
+
+    def self.add_order_creator(users, order_id)
```
steveyken, line 58: still doesn't feel right to have this in a concern... it's too 'order' specific

PR #1169, opened 2020-12-03
GCW-3416- Show recently used `package_type` on `package_type` selection overlay

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3416

### What does this PR do?

This PR adds `user_favourites` model to display recently used package_types.

### DIFF
--- a/app/controllers/api/v1/ability.rb
+++ b/app/controllers/api/v1/ability.rb
@@ -302,6 +302,7 @@ module Api
           can %i[show create update destroy], PackageSet
           can %i[index], Restriction
           can %i[index], PackagesInventory
+          can %i[index], UserFavourite
         end
         can [:show], Package,  orders_packages: { order: { created_by_id: @user_id }}
         can [:show], Package,  requested_packages: { user_id: @user_id }
--- a/app/controllers/api/v1/ability.rb
+++ b/app/controllers/api/v1/ability.rb
@@ -302,7 +302,7 @@ module Api
           can %i[show create update destroy], PackageSet
           can %i[index], Restriction
           can %i[index], PackagesInventory
-          can %i[index], UserFavourite
+          can %i[index], UserFavourite, user_id: @user_id
         end
         can [:show], Package,  orders_packages: { order: { created_by_id: @user_id }}
         can [:show], Package,  requested_packages: { user_id: @user_id }
--- a/app/controllers/api/v1/user_favourites_controller.rb
+++ b/app/controllers/api/v1/user_favourites_controller.rb
@@ -13,7 +13,8 @@ module Api
 
       api :GET, '/v1/user_favourites', "Fetches user_favourites"
       def index
-        render json: @user_favourites, each_serializer: serializer
+        user_favourites = @user_favourites.where(user_id: current_user.id)
+        render json: user_favourites, each_serializer: serializer
       end
 
       private
--- a/app/controllers/api/v1/user_favourites_controller.rb
+++ b/app/controllers/api/v1/user_favourites_controller.rb
@@ -1,0 +1,21 @@
+module Api
+  module V1
+    class UserFavouritesController < Api::V1::ApiController
+      load_and_authorize_resource :user_favourite, parent: false
+
+      resource_description do
+        formats ['json']
+        error 401, "Unauthorized"
+        error 404, "Not Found"
+        error 422, "Validation Error"
+        error 500, "Internal Server Error"
+      end
+
+      api :GET, '/v1/user_favourites', "Fetches user_favourites"
+      def index
+        user_favourites = @user_favourites.where(favourite_type: params[:types])
+        render json: user_favourites
+      end
+    end
+  end
+end
--- a/app/controllers/api/v1/user_favourites_controller.rb
+++ b/app/controllers/api/v1/user_favourites_controller.rb
@@ -13,8 +13,13 @@ module Api
 
       api :GET, '/v1/user_favourites', "Fetches user_favourites"
       def index
-        user_favourites = @user_favourites.where(favourite_type: params[:types])
-        render json: user_favourites
+        render json: @user_favourites, each_serializer: serializer
+      end
+
+      private
+
+      def serializer
+        Api::V1::UserFavouriteSerializer
       end
     end
   end
--- a/app/controllers/api/v1/user_favourites_controller.rb
+++ b/app/controllers/api/v1/user_favourites_controller.rb
@@ -13,8 +13,7 @@ module Api
 
       api :GET, '/v1/user_favourites', "Fetches user_favourites"
       def index
-        user_favourites = @user_favourites.where(user_id: current_user.id)
-        render json: user_favourites, each_serializer: serializer
+        render json: @user_favourites, each_serializer: serializer
       end
 
       private
--- a/app/models/concerns/auto_favourite.rb
+++ b/app/models/concerns/auto_favourite.rb
@@ -14,6 +14,7 @@ module AutoFavourite
   end
   
   class_methods do
+    @@auto_favourite_relations = false
     def auto_favourite(enabled = true)
       @@auto_favourite_enabled = enabled
     end
--- a/app/models/concerns/auto_favourite.rb
+++ b/app/models/concerns/auto_favourite.rb
@@ -4,21 +4,17 @@ module AutoFavourite
   included do
     before_save   :mark_recent_usage
     after_destroy :remove_from_favourites
-    
+
     scope :recently_used, -> (user_id) {
-      joins(%{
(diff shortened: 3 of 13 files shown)


app/models/concerns/auto_favourite.rb
```diff
@@ -0,0 +1,46 @@
...
+  
+  included do
+    after_create  :mark_recent_usage
+    after_update  :mark_recent_usage
+    after_destroy :remove_from_favourites
+    
+    scope :recently_used, -> (user_id) {
+      joins(%{
+        LEFT JOIN user_favourites ON user_favourites.favourite_type = '#{self.class.name}' AND user_favourites.user_id = #{user_id}
+      }).order('user_favourites.updated_at').limit(20)
```
patrixr, line 12: don't think `limit` has its place in a scope. That's the responsibility of whoever/wherever calls the scope

app/models/concerns/auto_favourite.rb
```diff
@@ -0,0 +1,46 @@
...
+  def mark_recent_usage    
+    UserFavourite.add_user_favourite(self) if @@auto_favourite_enabled
+    
+    relations = @@auto_favourite_relations || []
+    
+    relations.each do |name|
+      rel = self.try(name)
+      if rel.present?
+        foreign_key = self._reflections[name].foreign_key
+        UserFavourite.add_user_favourite(rel) #if self["#{foreign_key}_changed?"].present?
```
patrixr, line 38: you need that commented condition, otherwise all the relations will be marked every time. Or is that something we want ??

patrixr: @steveyken If I modify, say the notes of a package, do we consider the package's type as "recently used" ? Or only when the package's type is changed.

spec/controllers/api/v1/user_favourites_controller_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'rails_helper'
+
+RSpec.describe Api::V1::UserFavouritesController, type: :controller do
+  let!(:user) { create(:user, :with_token, :with_can_manage_packages_permission) }
+  let(:user1) { create(:user) }
+  let!(:user_favourite) { create(:user_favourite, user: user) }
+  let!(:user_favourite1) { create(:user_favourite) }
+  let(:parsed_body) { JSON.parse(response.body) }
+
+  describe "GET user_favourit" do
```
patrixr, line 10: typo

app/controllers/api/v1/ability.rb
```diff
@@ -302,6 +302,7 @@ def package_abilities
           can %i[show create update destroy], PackageSet
           can %i[index], Restriction
           can %i[index], PackagesInventory
+          can %i[index], UserFavourite
```
patrixr, line 305: That's not right, a user should only be able to see his/her favourites

```suggestion
can [:index], UserFavourite,   user_id: @user_id
```

app/controllers/api/v1/user_favourites_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+        formats ['json']
+        error 401, "Unauthorized"
+        error 404, "Not Found"
+        error 422, "Validation Error"
+        error 500, "Internal Server Error"
+      end
+
+      api :GET, '/v1/user_favourites', "Fetches user_favourites"
+      def index
+        user_favourites = @user_favourites.where(user_id: current_user.id)
```
patrixr, line 16: You can let cancan handle the user filtering part (look at comment I wrote in ability.rb)

patrixr (changes requested):

app/controllers/api/v1/user_favourites_controller.rb
abulsayyad123: sure.. ill add this change.

app/models/package.rb
```diff
@@ -88,6 +89,9 @@ class Package < ApplicationRecord
 
   attr_accessor :skip_set_relation_update, :request_from_admin, :detail_attributes
 
+  auto_favourite                              # the package will be auto-favourited
+  auto_favourite_relations ['package_type']   # the package_type will be auto-favourited
```
patrixr, line 93:
```suggestion
  auto_favourite_relations ['package_type']   # the package_type will be auto-favourited when changed
```

app/models/concerns/auto_favourite.rb
```diff
@@ -0,0 +1,45 @@
...
+  def mark_recent_usage    
+    UserFavourite.add_user_favourite(self) if @@auto_favourite_enabled
+    
+    relations = @@auto_favourite_relations || []
+    
+    relations.each do |name|
+      rel = self.try(name)
+      if rel.present?
+        foreign_key = self._reflections[name].foreign_key
+        UserFavourite.add_user_favourite(rel) if self.new_record? || self.send("#{foreign_key}_changed?")
```
patrixr, line 37: `self.try("#{foreign_key}_changed?".to_sym)` is probably safer

patrixr (approved):

app/models/package.rb
```diff
@@ -88,6 +89,9 @@ class Package < ApplicationRecord
 
   attr_accessor :skip_set_relation_update, :request_from_admin, :detail_attributes
 
+  auto_favourite                              # the package will be auto-favourited
+  auto_favourite_relations ['package_type']   # the package_type will be auto-favourited when changed
```
steveyken, line 93: We could collapse the API into this:

```suggestion
    auto_favourite(relations: ['package_type'], enabled: true)
```

app/models/user_favourite.rb
```diff
@@ -0,0 +1,53 @@
...
+    def add_user_favourite(record, persistent: false, user: User.current_user)
+      return unless user.present?
+      
+      UserFavourite.where(
+        user: user,
+        favourite: record,
+        persistent: persistent
+      ).first_or_initialize do |fav|
+        fav.save  if fav.new_record?
+        fav.touch if fav.persisted?
```
steveyken, line 42: This may fire 2 updates in some cases, so touching the updated_at attribute manually and then saving, should fix this.

```suggestion
        fav.updated_at = Time.now
        fav.save
```

db/migrate/20201126092838_create_user_favourites.rb
```diff
@@ -0,0 +1,15 @@
...
+      t.string    :favourite_type, index: true
+      t.integer   :favourite_id
+      t.integer   :user_id
+      t.boolean   :persistent, default: false 
+
+      t.timestamps
+    end
+
+    add_index :user_favourites, [:favourite_type, :favourite_id], unique: true
+    add_foreign_key :user_favourites, :users, column: :user_id
```
steveyken, line 13: Can also add an index for `updated_at` as we are using that to order records.

```
add_index :updated_at
```

db/migrate/20201126092838_create_user_favourites.rb
```diff
@@ -0,0 +1,15 @@
...
+    create_table :user_favourites do |t|
+      t.string    :favourite_type, index: true
+      t.integer   :favourite_id
+      t.integer   :user_id
+      t.boolean   :persistent, default: false 
+
+      t.timestamps
+    end
+
+    add_index :user_favourites, [:favourite_type, :favourite_id], unique: true
```
steveyken, line 12: It's good practice to have the reverse index on polymorphic associations as postgresql takes SQL parameter ordering rather seriously. We want best chance of an index match if writing `WHERE record_id=123 and record_type='XZY'`

```suggestion
    add_index :user_favourites, [:favourite_type, :favourite_id], unique: true
    add_index :user_favourites, [:favourite_id, :favourite_type], unique: true
```

steveyken (commented):
Overall, a very good feature. Some minor changes/improvement suggestions

db/migrate/20201126092838_create_user_favourites.rb
abulsayyad123: This is something I learned. Thanks.

spec/controllers/api/v1/user_favourites_controller_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  let!(:user) { create(:user, :with_token, :with_can_manage_packages_permission) }
+  let(:user1) { create(:user) }
+  let!(:user_favourite) { create(:user_favourite, user: user) }
+  let!(:user_favourite1) { create(:user_favourite) }
+  let(:parsed_body) { JSON.parse(response.body) }
+
+  describe "GET user_favourite" do
+    before(:each) do
+      generate_and_set_token(user) 
+      current_user = user
```
bharat619, line 13: I think this is not needed

swatijadhav (approved):

bharat619 (approved):

PR #1173, opened 2020-12-08
Drop sidekiq retries down to 3

steveyken:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3435

### What does this PR do?

FEATURE: For push notification jobs, these jobs don't need to be retried 25 times. If they fail, we should try 3 times and then give up.

### Impacted Areas

Push notifications sent to iOS and Android devices.


steveyken:
I'm going back to the drawing board on this... `sidekiq_options` requires sidekiq 6 which requires rails 6... we are on rails 5.2 at the moment

bharat619 (approved):

PR #1174, opened 2020-12-09
GCW-3442 Shareables

patrixr:
# Shareables

New: 
- Shareable model + migration + controller + serializer + specs

## Sharing a record

Create a shareable row

```ruby
Shareable.create(resource: offer)
```

An expiry date can be included

```ruby
Shareable.create(resource: offer, expires_at: 10.days.from_now)
```

### public_uid

Each shareable resource will have a `public_uid` which can be used to reference them

### allow_listing

Even though a resource has been shared, whether it can be exposed on the website depends on this `allow_listing` boolean

### Fetching shared resources directly

`GET /api/v2/shared/offers` 
`GET /api/v2/shared/offers/:public_uid`

### Creating/updating shareable rows via the API

`GET /api/v2/shareables`
`GET /api/v2/shareables/:id`
`POST /api/v2/shareables`
`PUT /api/v2/shareables/:id`
`DELETE /api/v2/shareables/:id`

#### Added model scopes

- `:publicly_shared` Will only select records which are published (and not expired)
- `:publicly_listed` Will only select records which are published AND can be listed

#### `public_context`

This concern gives each model a "public context" which can be configured, and natively has a default scope preventing non-shared records from showing up. It's basically an _altered_ version of the model

Example

```ruby
class Offer < ApplicationRecord
  include ShareSupport

  has_many :items

  public_context do
    # Add any specific relationship/scope/method here that only applies to the public context
    # e.g under the public context, the order will only ever have items which are also public
    has_many :items, -> { publicly_shared }
  end
end

PublicOffer = Offer.public_context
puts Offer.count            # => 10
puts PublicOffer.count # => 2
```

##### Why this exists

Allowing to have a "public" version of a model, allows us to pipe those models in a serializer without worrying about exposing any relationships it shouldn't


app/models/offer.rb
```diff
@@ -36,6 +36,13 @@ class Offer < ApplicationRecord
   has_many :packages, through: :offers_packages
   has_many :messages, as: :messageable, dependent: :destroy
 
+  #
```
patrixr, line 39: @steveyken 😉

app/controllers/api/v1/offers_controller.rb
```diff
@@ -80,7 +80,10 @@ def search
 
       api :GET, '/v1/offers/1', "List an offer"
       def show
-        render json: offer_serializer.new(@offer, exclude_messages: params["exclude_messages"] == "true").as_json
+        render json: offer_serializer.new(@offer, {
+          include_organisations_users: params["include_organisations_users"] == "true",
```
steveyken, line 84: We need to ensure this param can only be set to true for staff users. A small helper method could accomplish this

app/serializers/api/v1/user_serializer.rb
```diff
@@ -7,7 +7,7 @@ class UserSerializer < ApplicationSerializer
     has_one :image, serializer: ImageSerializer
     has_one :address, serializer: AddressSerializer
     has_many :user_roles, serializer: UserRoleSerializer
-    # has_many :organisations_users, serializer: OrganisationsUserSerializer
+    has_many :organisations_users, serializer: OrganisationsUserSerializer
```
steveyken, line 10: I've made a comment in `OffersController` to ensure this serializer is only included if user is staff

steveyken (commented):
Awesome work! Just one main comment in offer controller.

PR #1179, opened 2020-12-14
GCW-3427 Adding Process Actions

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3427

### What does this PR do?
Add source to Process actions

### DIFF
--- a/app/controllers/api/v1/packages_controller.rb
+++ b/app/controllers/api/v1/packages_controller.rb
@@ -280,6 +280,7 @@ module Api
           action: params[:action_name],
           source: source,
           description: params[:description])
+
         send_stock_item_response
       end
 
--- a/app/controllers/api/v1/packages_controller.rb
+++ b/app/controllers/api/v1/packages_controller.rb
@@ -278,6 +278,7 @@ module Api
           quantity: params[:quantity].to_i,
           location: params[:from],
           action: params[:action_name],
+          processing_destination_lookup_id: params[:processing_destination_lookup_id],
           description: params[:description])
 
         send_stock_item_response
--- a/app/controllers/api/v1/packages_controller.rb
+++ b/app/controllers/api/v1/packages_controller.rb
@@ -278,9 +278,8 @@ module Api
           quantity: params[:quantity].to_i,
           location: params[:from],
           action: params[:action_name],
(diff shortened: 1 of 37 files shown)


shreyas098 (approved):

app/models/processing_destinations_lookup.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class ProcessingDestinationsLookup < ApplicationRecord
```
patrixr, line 3: I feel the name `ProcessingDestinationsLookup` is unnecessarily long. `ProcessingDestination` is more than enough

patrixr: also it would kinda mean that the db table is a table of lookups
whereas it _is_ a lookup table, of processing destinations

app/models/concerns/operations/stock_operations.rb
```diff
@@ -114,7 +114,7 @@ def register_loss(package, quantity:, location: nil, action: 'loss', description
     # @param [String] action: the inventory action (optional)
     # @param [String] description: notes to detail the action
     #
-    def register_quantity_change(package, quantity:, location:, action:, description: nil, source: nil)
+    def register_quantity_change(package, quantity:, location:, action:, description: nil, source: nil, processing_destination_lookup_id: nil)
```
patrixr, line 117: That parameter doesn't have its place here. Please remove that. 
The processing destination _is_ a source, which is already in the params.

The _caller_ of this method should pass in the source

app/controllers/api/v1/packages_controller.rb
```diff
@@ -278,6 +278,7 @@ def register_quantity_change
           quantity: params[:quantity].to_i,
           location: params[:from],
           action: params[:action_name],
+          processing_destination_lookup_id: params[:processing_destination_lookup_id],
```
patrixr, line 281: Set the source correctly _here_.

patrixr (changes requested):
Small changes, nothing major

patrixr (approved):

steveyken (commented):
Good job

PR #1181, opened 2020-12-16
GCW-3388 Add restrictions for mentioning users

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3388
NOTE: Specify ticket link

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -76,8 +76,9 @@ module Api
       def mentionable_users
         return render json: { users: [] } if params['roles'].nil?
 
-        @users = User.mentionable_users({ roles: params[:roles],
-                                          order_id: params[:order_id] })
+        @users = User.mentionable_users( roles: params[:roles],
+                                         order_id: params[:order_id],
+                                         is_private: params[:is_private])
 
         render json: @users, each_serializer: Api::V1::UserMentionsSerializer
       end
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -73,12 +73,17 @@ module Api
         render json: Order.counts_for(params[:id])
       end
 
+      api :GET, '/v1/users/mentionable_users', 'Get mentionable users based on messageable context and roles'
+      param :roles, String, desc: 'String of roles that needs to be mentioned'
+      param :messageable_type, String, desc: 'Type of messageable. Offer, Item, Package etc.', allow_nil: true
+      param :messageable_id, String, desc: 'Id of any messageable type for Offer, Item, Package etc.', allow_nil: true
       def mentionable_users
         return render json: { users: [] } if params['roles'].nil?
 
-        @users = User.mentionable_users( roles: params[:roles],
-                                         order_id: params[:order_id],
-                                         is_private: params[:is_private])
+        @users = User.mentionable_users(roles: params[:roles],
+                                        messageable_type: params[:messageable_type],
+                                        messageable_id: params[:messageable_id],
+                                        is_private: params[:is_private])
 
         render json: @users, each_serializer: Api::V1::UserMentionsSerializer
       end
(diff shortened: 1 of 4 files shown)


app/models/concerns/mentionable.rb
```diff
@@ -48,20 +48,27 @@ def parse_id_from_decorated_ids(ids)
...
 
   included do
-    def self.mentionable_users(opts = {})
-      users = active.exclude_user(User.current_user.id)
-                    .with_roles(mentionable_roles(opts[:roles])).distinct.to_a
-      users << add_order_creator(users, opts[:order_id]) if opts[:order_id]
+    def self.mentionable_users(roles:, order_id:, is_private: false)
+      users = User.active.exclude_user(User.current_user.id)
+                  .with_roles(mentionable_roles(roles)).distinct.to_a
+      users << add_order_creator(users, order_id) unless bool_cast.call(is_private)
```
patrixr, line 54: Looks good, but I would make it generic to any resource type, not just order. 
For sure in the coming months we'll talk about mentionable users on offers, etc

app/models/concerns/mentionable.rb
```diff
@@ -48,20 +48,27 @@ def parse_id_from_decorated_ids(ids)
...
       user = Order.find_by(id: order_id)&.created_by
       user unless users.include? user
     end
 
     def self.mentionable_roles(roles)
       (MENTIONABLE_ROLES & roles.split(',').map(&:strip).uniq)
     end
+
+    def self.bool_cast
+      ->(x) { ActiveModel::Type::Boolean.new.cast(x) }
```
patrixr, line 71: why the proc ?

```suggestion
      ActiveModel::Type::Boolean.new.cast(x)
```

app/controllers/api/v1/users_controller.rb
```diff
@@ -76,8 +76,9 @@ def orders_count
       def mentionable_users
         return render json: { users: [] } if params['roles'].nil?
 
-        @users = User.mentionable_users({ roles: params[:roles],
-                                          order_id: params[:order_id] })
+        @users = User.mentionable_users( roles: params[:roles],
+                                         order_id: params[:order_id],
```
patrixr, line 80: Same as the comment below, messages are not order-specific, so if possible this shouldn't either

app/models/concerns/mentionable.rb
```diff
@@ -48,20 +48,27 @@ def parse_id_from_decorated_ids(ids)
...
-      users << add_order_creator(users, opts[:order_id]) if opts[:order_id]
+    def self.mentionable_users(roles:, order_id:, is_private: false)
+      users = User.active.exclude_user(User.current_user.id)
+                  .with_roles(mentionable_roles(roles)).distinct.to_a
+      users << add_order_creator(users, order_id) unless bool_cast.call(is_private)
       users.compact
     end
 
     def self.add_order_creator(users, order_id)
+      valid_roles = User.current_user.roles.pluck(:name) & ['Order administrator', 'Order fulfilment']
```
steveyken, line 59: `add_order_creator` is very specific to `order`. Can you use `mentionable_roles` or a different approach?

patrixr (approved):

app/models/concerns/mentionable.rb
```diff
@@ -48,27 +55,37 @@ def parse_id_from_decorated_ids(ids)
   end
 
   included do
-    def self.mentionable_users(roles:, order_id:, is_private: false)
+    def self.mentionable_users(roles:, messageable_id:, messageable_type:, is_private: false)
+      messageable = construct_messageable(messageable_type, messageable_id) unless [messageable_id, messageable_type].all?(&:nil?)
       users = User.active.exclude_user(User.current_user.id)
                   .with_roles(mentionable_roles(roles)).distinct.to_a
```
steveyken, line 61: This should probably include only users who have roles that are not expired.

steveyken (approved):
Thanks for the changes, Much better. One final thing I spotted.

PR #1182, opened 2020-12-21
Correct the indexes on user_favourites table

bharat619:
As in the title

### DIFF
--- a/app/models/user_favourite.rb
+++ b/app/models/user_favourite.rb
@@ -2,10 +2,10 @@ class UserFavourite < ApplicationRecord
   include PushUpdatesMinimal
 
   LIMIT_PER_TABLE = 20
-  
+
   belongs_to :favourite, polymorphic: true
   belongs_to :user
-  
+
   after_create :apply_limit
 
   # Live update rules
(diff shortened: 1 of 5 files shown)


app/models/user_favourite.rb
```diff
@@ -14,35 +14,36 @@ class UserFavourite < ApplicationRecord
...
     # Mark a record as a favourite
     #
     def add_user_favourite(record, persistent: false, user: User.current_user)
       return unless user.present?
-      
+
       UserFavourite.where(
         user: user,
         favourite: record,
         persistent: persistent
```
patrixr, line 39: gotta remove this line here

patrixr (changes requested):
Can you add a test for this as well ?

- create a persistent favourite
- create a non-persistent favourite for the same record and user
- should not try to create a new record

app/models/user_favourite.rb
patrixr:
```suggestion
```

app/models/user_favourite.rb
```diff
@@ -14,35 +14,36 @@ class UserFavourite < ApplicationRecord
...
       ).first_or_initialize do |fav|
+        fav.persistent = fav.persistent ? true : persistent
```
swatijadhav, line 41:
```suggestion
        fav.persistent = fav.persistent || persistent
```

patrixr (approved):

PR #1185, opened 2020-12-22
GCW-3441 Set offers-packages for inventorized packages

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3441

### What does this PR do?
Set offers-packages for inventorized packages

### DIFF
--- a/lib/tasks/packages.rake
+++ b/lib/tasks/packages.rake
@@ -97,4 +97,17 @@ namespace :goodcity do
   task :compute_quantities, [:param] => [:environment] do |_, args|
     Goodcity::PackageSetup.compute_quantities(rehearse: args[:param].eql?('rehearse'))
   end
+
+  # rails goodcity:set_offer_for_packages
+  desc "update offers for packages"
+  task set_offer_for_packages: :environment do
+    Item.find_in_batches(batch_size: 100).each do |items|
+      items.each do |item|
+        packages = item.packages.inventorized.received
+        packages.each do |package|
+          package.assign_offer
+        end
+      end
+    end
+  end
 end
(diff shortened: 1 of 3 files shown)


lib/tasks/packages.rake
```diff
@@ -97,4 +97,17 @@ namespace :goodcity do
   task :compute_quantities, [:param] => [:environment] do |_, args|
     Goodcity::PackageSetup.compute_quantities(rehearse: args[:param].eql?('rehearse'))
   end
+
+  # rails goodcity:set_offer_for_packages
+  desc "update offers for packages"
+  task set_offer_for_packages: :environment do
+    Item.find_in_batches(batch_size: 100).each do |items|
```
steveyken, line 104: I think it would be faster to start with a much smaller list as this could get potentially very large. Something like this? (untested)

```Package.inventorize.received.joins(:item).includes(:item, :offers_package).where('items.offer_id IS NOT NULL').find_each do |package|
  packages.assign_offer unless packages.offers_packages.map(&:offer_id).include(package.item.offer_id)
```

steveyken (approved):

lib/tasks/packages.rake
swatijadhav: Updated. Thanks @steveyken 👍🏻

steveyken (approved):

steveyken:
Given the rake task is not going to be run regularly and is a background job, I'm fine if we don't optimise any further. Potentially, we could have used `includes(...)` to load the `offer_package` association but no need

bharat619 (approved):

PR #1193, opened 2021-01-06
GCW-3465 ADD: notes fields to shareables

patrixr:
## Changes

- ADD `notes` and `notes_zh_tw` to the Shareables
- Refactor 'PublicAttributes' serializer concern into `Formats`

## Serializer formats concern

On serializers V2, you can now add the `Formats` concern to create different versions of the serializers

Example

```ruby
  class OfferSerializer < Api::V2::GoodcitySerializer
    include Api::V2::Concerns::Formats

    # Normal attributes
    attributes :id, :notes

    format :internal do
      # These attributes will only be included if the 'internal' format is specified during render
      attributes :review_completed_at, :received_at, :delivered_by, :closed_by_id, :cancelled_at
      attribute :received_by_id

      format :delivery do
        # These attributes will only be included if the 'internal' AND 'delivery' formats are specified during render
        attribute :gogovan_transport_id
      end
    end

    format :conditions do
      # conditions still work
      attribute(:foo, if: Proc.new { true }) { 'foo' }
      attribute(:bar, if: Proc.new { false }) { 'bar' }
    end
  end
  
  
# How to set formats

OfferSerializer.new(offer, { params: { format: :internal } })
OfferSerializer.new(offer, { params: { format: [:internal] } })
OfferSerializer.new(offer, { params: { format: [:internal, :delivery] } })
```


steveyken (approved):
Formats... Stunning. Mind blown!

app/serializers/api/v2/concerns/formats.rb
```diff
@@ -0,0 +1,50 @@
...
+        def scoped_formats
+          @@scoped_formats ||= []
+        end
+
+        alias_method :jsonapi_attributes, :attributes
+
+        #
+        # Override the attribute method to support contexts
+        #
+        def attributes(*attributes_list, &block)
```
bharat619, line 17: 🔥

bharat619 (approved):

PR #1195, opened 2021-01-12
GCW-3407 Gogox Gem Integration and Transport APIs

swatijadhav:
### Ticket Link: 

NOTE: Specify ticket link

### What does this PR do?

FEATURE: Specify feature name

BUG: Specify bug 

NOTE: Give a description of what this PR does. If it fixes any bug, specify cause of an issue and approach/solution added to fix that issue.

### Impacted Areas

### Screenshots

### Mockup Link

Note: Specify mockup link

### Linked PR's:


config/routes.rb
```diff
@@ -215,6 +215,13 @@
       resources :printers_users, only: [:create, :update]
       resources :processing_destinations, only: :index
 
+      get "transports/providers", to: "transports#providers"
```
steveyken, line 218: suggestions:

```
get "transports/providers", to: "transports#providers"
post "transports/quote", to: "transports#quote
post "transports/book", to: "transports#book"
get "transports/:id", to: "transports#show"
post "transports/:id/cancel", to: "transports#cancel"
post "transports/update_hook", to: "transports#update_hook" 
```

app/services/transport_service.rb
```diff
@@ -0,0 +1,124 @@
...
+        status: response["status"],
+        order_uuid: booking_id,
+        metadata: response
+      })
+    end
+  end
+
+  private
+
+  def storeOrderDetails(response)
```
steveyken, line 47:
```suggestion
  def store_order_details(response)
```

app/models/transport_provider_order.rb
```diff
@@ -0,0 +1,2 @@
+class TransportProviderOrder < ApplicationRecord
```
steveyken, line 1: Can we shorten this to `TransportOrder`?

steveyken (commented):
Great work. Good to see this taking shape and getting ready to integrate the new GGX booking

app/services/gogox.rb
```diff
@@ -0,0 +1,129 @@
...
+
+  attr_accessor :params, :time, :vehicle, :district_id
+
+  VEHICLE_TYPES = ["van", "mudou", "mudou9"]
+
+  def initialize(options = {})
+    @params        = options
+    @time         = parse_pickup_time(options[:scheduled_at])
+    @vehicle      = options[:vehicle_type]
+    @district_id  = options[:district_id]
```
bharat619, line 11: Is the `@district_id` instance variable used in this class somehow?

app/services/gogox.rb
```diff
@@ -0,0 +1,129 @@
+class Gogox
```
bharat619, line 1: Just thing which can be done, no compulsoion

```ruby
class Gogox < BaseTransportation
   # it inherits a base class
end
```

```ruby
class BaseTransportation
   # Common code goes in here which would be applicable to almost all the types, for ex:

  def parse_pickup_time(time = nil)
    return time if time.present?

    # next available date within next 5 days
    next_available_date = DateSet.new(5, 1).available_dates.first
    (next_available_date.beginning_of_day + 12.hours)
  end

  def parse_time
    # logic follows
  end

   def book
     # raise error if child class doesnt provide implementation for it
     raise NotImplementedError
   end
end
```

app/services/gogox.rb
```diff
@@ -0,0 +1,129 @@
...
+    (next_available_date.beginning_of_day + 12.hours)
+  end
+
+  def parse_time
+    @time = DateTime.parse(@time.to_s)
+    @time = @time.zone == "HKT" ? @time : @time.in_time_zone("Asia/Hong_Kong")
+    @time.to_i
+  end
+
+  class ValueError < StandardError; end
```
bharat619, line 128: Instead of just value error defining here, create a error in errors.rb
something like TransportationServiceError

app/services/transport_service.rb
```diff
@@ -0,0 +1,110 @@
+class TransportService
+
+  attr_accessor :provider_name, :params, :provider, :booking_id, :user, :district_id
+
+  def initialize(options={})
+    @params = options
+    @provider_name = options && options[:provider]
```
bharat619, line 7:
```suggestion
    @provider_name = options.dig(:provider)
```

app/services/transport_service.rb
```diff
@@ -0,0 +1,110 @@
...
+    @provider ||= Object::const_get(provider_name)
+    @booking_id = options && options[:booking_id]
```
bharat619, line 9:
```suggestion
    @booking_id = options.dig(:booking_id)
```

app/services/gogox.rb
swatijadhav: yes, not used, will remove it.

app/services/gogox.rb
swatijadhav: @bharat619 you are suggesting to use template design pattern here, which seems valid in this case.
But I have not added one more abstraction layer here since we are not yet sure about how the other providers' API is, do they work in the same manner or there are different.

Once we have at least one more provider API available, we can introduce above suggestion.

app/controllers/api/v1/transports_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+      param :provider, String, desc: "Provider selected for transport"
+      param :vehicle_type, String, desc: "Transport vehicle-type"
+      param :offer_id, String, desc: "Id of the offer"
+      param :schedule_at, String, desc: "Scheduled time for delivery"
+      param :district_id, String, desc: "Id of the district"
+      param :pickup_contact_name, String, desc: "Contact Person Name"
+      param :pickup_contact_phone, String, desc: "Contact Person Mobile"
+      param :pickup_street_address, String, desc: "Pickup Address"
+      def book
+        order_info = TransportService.new(transport_params.to_h).book
```
steveyken, line 34: Can you filter by `offer_id` to ensure user can only book for their offers (or admin can book on their behalf)?

app/controllers/api/v1/transports_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+        render json: TransportProvider.all.cached_json
+      end
+
+      api :POST, '/v1/transports/quote', "Get provider quote"
+      param :provider, String, desc: "Provider selected for transport"
+      param :vehicle_type, String, desc: "Transport vehicle-type"
+      param :offer_id, String, desc: "Id of the offer"
+      param :schedule_at, String, desc: "Scheduled time for delivery"
+      param :district_id, String, desc: "Id of the district"
+      def quote
```
steveyken, line 19: Can you filter by `offer_id` to ensure user can only quote for their offers (or admin can book on their behalf)?

app/controllers/api/v1/transports_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+
+      api :GET, '/v1/transports/providers', "List all GoodCity Tranports Options."
+      def providers
+        render json: TransportProvider.all.cached_json
+      end
+
+      api :POST, '/v1/transports/quote', "Get provider quote"
+      param :provider, String, desc: "Provider selected for transport"
+      param :vehicle_type, String, desc: "Transport vehicle-type"
+      param :offer_id, String, desc: "Id of the offer"
```
steveyken, line 16: What if it is a booking for order delivery rather than offer delivery?

app/controllers/api/v1/transports_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+      end
+
+      api :POST, '/v1/transports/:order_uuid/cancel', "Cancel GoodCity Tranport order."
+      def cancel
+        order_info = TransportService.new({booking_id: params[:order_uuid], provider: transport_provider}).cancel
+        render json: order_info
+      end
+
+      api :POST, '/v1/transports/update_hook', "Webhook to update transport status"
+      def update_hook
```
steveyken, line 51: When implemented, we should use our own authorization check here. E.g. webhook can be `/api/v1/transports/update_hook?shared_key=ds23f934dfs&param1=df&param2=23`

Then check `shared_key` is as expected.

db/migrate/20210111115250_create_transport_orders.rb
```diff
@@ -0,0 +1,14 @@
+class CreateTransportOrders < ActiveRecord::Migration[5.2]
+  def change
+    create_table :transport_orders do |t|
+      t.integer :transport_provider_id
+      t.string :order_uuid
+      t.string :status
+      t.datetime :scheduled_at
+      t.jsonb :metadata
+      t.integer :offer_id
```
steveyken, line 9: I think we need this to work for `offers` and `orders`. Polymorphic?

steveyken (changes requested):
Can you remove `ngrok` binary?

Great work, almost complete

steveyken (changes requested):
There's a few issues with un-sanitized user input which need to be addressed. Otherwise, looking good.

PR #1199, opened 2021-01-15
GCW 3475 Adding carryouts and other to offer filter

shreyas098:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3475

### What does this PR do?

FEATURE: STOCK: user can filter orders of type "carry out" or "other"

### DIFF
--- a/spec/controllers/api/v1/orders_controller_spec.rb
+++ b/spec/controllers/api/v1/orders_controller_spec.rb
@@ -236,11 +236,12 @@ RSpec.describe Api::V1::OrdersController, type: :controller do
           expect(parsed_body["meta"]["search"]).to eql("iphone")
         end
 
-        ["appointment", "online_orders", "shipment", "other"].each do |type|
+        ["appointment", "online_orders", "shipment", "carry_out", "other"].each do |type|
           it "can return a single order of type #{type}" do
             create :appointment, :with_state_submitted, description: "IPhone 100s"
             create :online_order, :awaiting_dispatch, description: "IPhone 100s", order_transport: ggv_transport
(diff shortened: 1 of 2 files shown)


spec/controllers/api/v1/orders_controller_spec.rb
```diff
@@ -251,6 +252,14 @@ def create_order_with_transport(state, opts = {})
           end
         end
 
+        it "does not return valid order type when type is other" do
+          create :online_order, :awaiting_dispatch, description: "Testing", order_transport: ggv_transport
+          get :index, params: { type: "other" }
+          expect(response.status).to eq(200)
+          expect(parsed_body["designations"].count).to eq(0)
```
bharat619, line 259: instead of count / total pages,
it would be good to test if all the response object has detail_type only as 'other'

````
order_types = response['designations'].map { |res| res['detail_type'] }
expect(order_types.uniq).to match_array(['other'])
```

spec/controllers/api/v1/orders_controller_spec.rb
```diff
@@ -251,6 +252,14 @@ def create_order_with_transport(state, opts = {})
           end
         end
 
+        it "does not return valid order type when type is other" do
```
bharat619, line 255:
```suggestion
        context "when order_type is other" do
          it 'response will not have other order_type' do
             <your test case>
          end
        end
```

bharat619 (changes requested):
Changes requested in specs.
Also it would be good to have spec if order_type is carryouts as well

bharat619 (approved):

abulsayyad123 (approved):

PR #1208, opened 2021-01-28
GCW-3489: Remove watcher toggle

patrixr:
Stocktake revision population code now uses a single query, that allows:

- the removal of callback toggle on Watcher, that was causing the bug
- improved performance as it's all done in a single query now

Note: I would advise against using `thread_mattr_accessor` for the time being as it clearly hasn't been working as intented

### DIFF
--- a/app/models/stocktake.rb
+++ b/app/models/stocktake.rb
@@ -66,7 +66,7 @@ class Stocktake < ApplicationRecord
 
     ActiveRecord::Base.connection.execute <<-SQL
       INSERT INTO stocktake_revisions (package_id, stocktake_id, quantity, dirty, state, created_by_id, created_at, updated_at)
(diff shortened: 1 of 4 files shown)


abulsayyad123 (approved):

bharat619 (approved):

app/models/stocktake.rb
```diff
@@ -58,29 +58,26 @@ class Stocktake < ApplicationRecord
...
+    attrs = {
+      created_by_id: User.current_user&.id || User.system_user.id,
+      state: "'pending'",
+      dirty: true,
+      stocktake_id: id,
+      quantity: 0,
+    }
+
+    ActiveRecord::Base.connection.execute <<-SQL
+      INSERT INTO stocktake_revisions (package_id, created_at, updated_at, #{attrs.keys.join(',')})
```
steveyken, line 70: Whilst this code is currently safe, can further protect with `sanitize_sql` to protect from future accidental inclusions of user unsanitized input.

steveyken (approved):

PR #1209, opened 2021-01-28
GCW-3491 Set preferred language for current_user

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3491

### What does this PR do?
Adds default language

### DIFF
--- a/app/controllers/api/v1/authentication_controller.rb
+++ b/app/controllers/api/v1/authentication_controller.rb
@@ -157,6 +157,14 @@ module Api
 
       def current_user_profile
         authorize!(:current_user_profile, User)
+
+        # If the preferred_langugage is not set for the user, then
+        # Set the current_user.preferred_langugage to the locale which is used by the user in browser
+        # NOTE: This is done in this request because
+        #  For setting the preferred_langugage in other endpoints of authentication_controller,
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/authentication_controller.rb
```diff
@@ -157,6 +157,14 @@ def verify
 
       def current_user_profile
         authorize!(:current_user_profile, User)
+
+        # If the preferred_langugage is not set for the user, then
+        # Set the current_user.preferred_langugage to the locale which is used by the user in browser
+        # NOTE: This is done in this request because
+        #  For setting the preferred_langugage in other endpoints of authentication_controller,
+        #  the user needs to be logged out
+        current_user.update(preferred_language: I18n.locale) unless current_user.preferred_language.present?
```
steveyken, line 166: If you use the conditional, then the language won't change when the user changes their language in the browser. Note: `update` will only issue an SQL write operation if the attribute changes so we can rely on this to avoid 'writing' to the database on every request.

```suggestion
        current_user.update(preferred_language: I18n.locale)
```

steveyken (approved):

abulsayyad123 (approved):

PR #1210, opened 2021-02-04
GCW-2785 avoid creating empty offers by admin-staff

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-2785

### What does this PR do?
avoid creating empty offers by admin-staff

### DIFF
--- a/app/controllers/api/v1/offers_controller.rb
+++ b/app/controllers/api/v1/offers_controller.rb
@@ -40,6 +40,9 @@ module Api
       param_group :offer
       def create
         @offer.created_by_id = current_user.id unless offer_params.has_key?(:created_by_id)
+        if @offer.created_by_id.blank? && @offer.under_review?
+          @offer = current_user.reviewed_offers.empty_offer_by_admin || @offer
+        end
         save_and_render_object(@offer)
       end
 
--- a/spec/controllers/api/v1/offers_controller_spec.rb
+++ b/spec/controllers/api/v1/offers_controller_spec.rb
@@ -270,6 +270,17 @@ RSpec.describe Api::V1::OffersController, type: :controller do
           expect(created_offer.created_by).to eq(nil)
         end
 
+        it "returns previously added empty offer" do
+          admin_offer_params = { created_by_id: nil,
+            state: "under_review", reviewed_by_id: supervisor.id.to_s  }
+          old_offer = create :offer, admin_offer_params
+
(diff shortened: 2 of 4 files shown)


spec/controllers/api/v1/offers_controller_spec.rb
```diff
@@ -270,6 +270,17 @@
           expect(created_offer.created_by).to eq(nil)
         end
 
+        it "rseturns previously added empty offer" do
```
shreyas098, line 273:
```suggestion
        it "returns previously added empty offer" do
```

shreyas098 (approved):

app/controllers/api/v1/offers_controller.rb
```diff
@@ -40,6 +40,9 @@ class OffersController < Api::V1::ApiController
       param_group :offer
       def create
         @offer.created_by_id = current_user.id unless offer_params.has_key?(:created_by_id)
+        if @offer.created_by_id.blank? && @offer.under_review?
```
bharat619, line 43: This fix works.

Putting out thoughts related to the existing implementation:

We first create an empty offer, and then as we add details, the offer is updated.  (Which I feel is wrong)
Ideally it should not even create an offer record if everything is empty. Only after an item is added or any other details are added, only then it should create it.

Should we change our current implementation over here? 

FYI, everything above can be thought of a separate ticket.

cc: @steveyken @patrixr @abulsayyad123 @shreyas098 @swatijadhav

steveyken: Good point to raise. Is there a reason that we allow admins to create blank offers? I know we do for donors to smooth their experience but it is necessary for admins also?

patrixr: I think we definitely want to avoid something like this. Generally any controller code that is written `if (some very unique state)` is dangerous.

I think the potential fixes are:
- Create a non-empty offer from the start (as @bharat619 suggests)
- Prompt the admin to select his/her existing offers when he/she tries to create another offer

More generally:
An admin **is allowed** to create an empty offer, so the behaviour of `POST /offers` should not change.
We can't have an endpoint that _pretends_ to create something but actually returns something else.
That would mean the server is making an _assumption_ on what the user is trying to do, and that's a poor api design choice (and a ticking time bomb)

Do make sure to get @steveyken 's input on this. Matt can also help to figure out a nice UI flow to avoid that problem.

PR #1211, opened 2021-02-09
GCW-3492 Rails Upgrade 6.0.1

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3492

### What does this PR do?
Rails API Version Upgrade


app/models/image.rb
```diff
@@ -6,7 +6,10 @@ class Image < ApplicationRecord
   include ShareSupport
 
   has_one :user, inverse_of: :image
-  belongs_to :imageable, polymorphic: true, touch: true
+  belongs_to :item
```
steveyken, line 9: @patrixr is this correct?

config/application.rb
```diff
@@ -1,11 +1,15 @@
...
+require "rails"
 # Pick the frameworks you want:
 require "active_model/railtie"
+require "active_job/railtie"
 require "active_record/railtie"
+require "active_storage/engine"
 require "action_controller/railtie"
 require "action_mailer/railtie"
-# require "action_view/railtie"
+require "action_view/railtie"
```
steveyken, line 11: Do we need action_view for Rails API projects?

config/environments/production.rb
```diff
@@ -14,17 +14,30 @@
   config.consider_all_requests_local       = false
   config.action_controller.perform_caching = true
 
-  # Enable Rack::Cache to put a simple HTTP cache in front of your application
-  # Add `rack-cache` to your Gemfile before enabling this.
-  # For large-scale production use, consider using a caching reverse proxy like nginx, varnish or squid.
-  # config.action_dispatch.rack_cache = true
+   # Ensures that a master key has been made available in either ENV["RAILS_MASTER_KEY"]
+  # or in config/master.key. This key is used to decrypt credentials (and other encrypted files).
+  # config.require_master_key = true
```
steveyken, line 19: should we be using this?

.rvmrc
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
```
steveyken, line 1: not sure if we need this checked in. rvm installations may vary per developer

steveyken (commented):
Thanks @bharat619 excellent work and clearly worked very hard on this. Just a few small comments

app/models/image.rb
patrixr: that looks wrong, that would remove the polymorphic relationship altogether

bharat619: @steveyken @patrixr , I havent removed that line
I have just moved it below

The belongs_to side of the relationship needed to be defined before the polymorphic association was defined. Else Rails didnt recognise the relationship and was throwing error

config/environments/production.rb
bharat619: We can adopt this next in a separate PR, get rid of secrets.yml file and make use of config/credentials.yml.enc
This master key will be the one to encrypt and decrypt the credentials.yml.enc

.rvmrc
bharat619: oops. was playing around with gemset. I have removed it

config/application.rb
bharat619: no. removed it

app/models/image.rb
steveyken: got it, thanks

config/environments/production.rb
steveyken: ok yes, separate PR another time. thanks

steveyken (approved):

app/models/user.rb
```diff
@@ -101,7 +101,7 @@ class User < ApplicationRecord
   scope :system, -> { where(roles: {name: "System"}).joins(:active_roles) }
 
   scope :staff, -> { where(roles: {name: ["Supervisor", "Reviewer"]}).joins(:active_roles) }
-  scope :except_stockit_user, -> { where.not(first_name: "Stockit", last_name: "User") }
```
patrixr, line 104: is that old version no longer working ?

app/models/user_favourite.rb
```diff
@@ -19,7 +19,7 @@ def apply_limit
     #
     # Delete non-persistent favorites > 20
     #
-    UserFavourite
+    UserFavourite.default_scoped
```
patrixr, line 22: @bharat619 not sure why we would need that

app/models/user.rb
bharat619: yeah.
multiple `not` conditions needs to be separated henceforth
for example `where.not(first_name: 'Anakin', last_name: 'Skywalker')` should be 
`where.not(first_name: 'Anakin').where.not(last_name: 'Skywalker')`

patrixr (approved):

app/models/user_favourite.rb
bharat619: @patrixr  This is to avoid the warning
```
Class level methods will no longer inherit scoping from `new` in Rails 6.1. To continue using the scoped relation, 
pass it into the block directly. To instead access the full set of models, as Rails 6.1 will, use `UserFavourite.default_scoped`
```

PR #1221, opened 2021-03-01
GCW-3220 validate packages value_hk_dollar

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3220

### What does this PR do?
validate packages value_hk_dollar


app/models/package.rb
```diff
@@ -60,6 +60,8 @@ class Package < ApplicationRecord
   end
 
   validates :package_type_id, presence: true
+  validates :value_hk_dollar, presence: true, numericality: { greater_than: 0 },
```
bharat619, line 63: model spec for testing positive, 0, negative and null value would be great

db/migrate/20210226072406_add_not_null_constraint_to_packages.rb
```diff
@@ -0,0 +1,5 @@
+class AddNotNullConstraintToPackages < ActiveRecord::Migration[5.2]
+  def change
+    change_column_null :packages, :value_hk_dollar, false
```
steveyken, line 3: I hadn't thought about the box/pallet case as they have no valuation. What should we do about that?

app/models/package.rb
```diff
@@ -60,6 +60,7 @@ class Package < ApplicationRecord
   end
 
   validates :package_type_id, presence: true
+  validates :value_hk_dollar, presence: true, if: -> { !new_record? && !box_or_pallet? }
```
steveyken, line 63: why do we need to have a condition on `new_record`?

steveyken (commented):
Thanks Swati, just had a couple of questions as I'm not sure what we should do in this case.

app/models/package.rb
swatijadhav: Here we already have a `before_create` callback to set `value_hk_dollar` when its blank.
https://github.com/crossroads/api.goodcity/blob/master/app/models/concerns/valuation_calculator.rb#L8-L12 

Hence skipping this validation for new-records.

steveyken: thanks, actually `before_validation` makes a lot more sense here... if we are computing a value, we really should check it meets our validation requirements.

db/migrate/20210226072406_add_not_null_constraint_to_packages.rb
swatijadhav: For box and pallet, valuation calculation sets value as 0.0 hence it will be handled.

steveyken (approved):

bharat619 (approved):

PR #1222, opened 2021-03-04
GCW-3443 rake task to clear and merge locations

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3443

### What does this PR do?
rake task to clear and merge locations

### DIFF
--- a/lib/classes/manage_location.rb
+++ b/lib/classes/manage_location.rb
@@ -1,10 +1,9 @@
 class ManageLocation
-
   def initialize(location_id)
     @location = Location.find_by(id: location_id)
   end
 
-  def is_empty_location?
+  def empty_location?
     @location.packages_locations.count.zero? &&
       @location.package_types.count.zero? &&
       PackagesInventory.where(location_id: @location.id).count.zero? &&
@@ -13,10 +12,9 @@ class ManageLocation
   end
 
   def self.merge_location(source_location, destination_location)
-    ['Printer', 'Stocktake', 'PackagesInventory', 'PackageType', 'PackagesLocation', 'Package'].each do |klass|
+    %w[Printer Stocktake PackagesInventory PackageType PackagesLocation Package].each do |klass|
       records = Object::const_get(klass).where(location_id: source_location.id)
       records.update_all(location_id: destination_location.id)
     end
   end
-
 end
--- a/lib/classes/manage_location.rb
+++ b/lib/classes/manage_location.rb
@@ -1,0 +1,22 @@
+class ManageLocation
+
+  def initialize(location_id)
+    @location = Location.find_by(id: location_id)
+  end
+
+  def is_empty_location?
+    @location.packages_locations.count.zero? &&
+      @location.package_types.count.zero? &&
+      PackagesInventory.where(location_id: @location.id).count.zero? &&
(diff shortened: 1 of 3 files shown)


lib/tasks/locations.rake
```diff
@@ -0,0 +1,53 @@
+namespace :goodcity do
+
+  # rails goodcity:remove_unused_locations
+  desc 'Delete locations which does not have any associated data'
+  task remove_unused_locations: :environment do
+    Location.find_in_batches(batch_size: 100).each do |locations|
```
bharat619, line 6: should we move the task to a new / existing lib file
and support the operations with specs?

lib/tasks/locations.rake
```diff
@@ -0,0 +1,53 @@
...
+          puts "Deleted location: #{location.id} #{location.building}-#{location.area}"
+        end
+      end
+    end
+  end
+
+  # rails goodcity:merge_locations
+  desc 'Move data from one location to another and delete location'
+  task merge_locations: :environment do
+    source_location_id = ENV['SOURCE_LOCATION_ID']
```
bharat619, line 19: same as above

lib/tasks/locations.rake
patrixr: I think we almost always want to enrich the capabilities of goodcity with tested operations, and call them from rakes

abulsayyad123 (approved):

lib/classes/manage_location.rb
```diff
@@ -0,0 +1,20 @@
+class ManageLocation
+  def initialize(location_id)
+    @location = Location.find_by(id: location_id)
+  end
+
+  def empty_location?
+    @location.packages_locations.count.zero? &&
+      @location.package_types.count.zero? &&
+      PackagesInventory.where(location_id: @location.id).count.zero? &&
```
patrixr, line 9: There will likely be cases where the we have packages_inventories records, but the quantity is 0. Those wouldn't be detected with this, right ?

patrixr (approved):
Looks pretty safe, although I dropped in a question : )

lib/classes/manage_location.rb
```diff
@@ -0,0 +1,20 @@
...
+      Printer.where(location_id: @location.id).count.zero? &&
+      Stocktake.where(location_id: @location.id).count.zero?
+  end
+
+  def self.merge_location(source_location, destination_location)
+    %w[Printer Stocktake PackagesInventory PackageType PackagesLocation Package].each do |klass|
```
patrixr, line 15: Do packages have a `location_id` ? If so we should probably remove that column

swatijadhav: Yeah it is there, maybe an old column as it is not used anywhere. But just to be on the safer side updating that value too.

shreyas098 (approved):

lib/classes/manage_location.rb
swatijadhav: yes @patrixr , those records will not be detected.

lib/classes/manage_location.rb
patrixr: yup sounds good, I'll create a ticket for removing this column separately

steveyken (commented):
Thank you. Good work

PR #1233, opened 2021-03-22
GC-3533-Validation should be added to ensure packages are present before creating online orders

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3533

### What does this PR do?

This PR adds validation for online orders to have orders_package before submitting. Also, fixed test cases as per changes.

### DIFF
--- a/spec/controllers/api/v1/requested_packages_controller_spec.rb
+++ b/spec/controllers/api/v1/requested_packages_controller_spec.rb
@@ -193,81 +193,107 @@ RSpec.describe Api::V1::RequestedPackagesController, type: :controller do
       end
     end
 
+    def create_requested_packges_for(user, order)
+      packages = order.orders_packages.map(&:package)
+      packages.map do |p|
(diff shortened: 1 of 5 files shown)


spec/controllers/api/v1/requested_packages_controller_spec.rb
```diff
@@ -193,81 +193,107 @@
       end
     end
 
+    def create_requested_packges_for(user, order)
+      packages = order.orders_packages.map(&:package)
+      packages.map do |p|
+        create(:requested_package, :with_available_package, package: p, user: user)
+      end
+    end
+
```
patrixr, line 202: that's kinda weird, if an order has `orders_packages` then there shouldn't be any `requested_packages`
When an order is submitted (cart_checkout), all the requested packages are added to the order, and then the requested packages are emptied

shreyas098 (approved):

bharat619 (approved):

saunmanoj888 (approved):

PR #1236, opened 2021-03-30
GCW-3534 Merge Users

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3534

### What does this PR do?
Feature to merge users.


abulsayyad123 (approved):

bharat619 (approved):

lib/goodcity/user_utils.rb
```diff
@@ -0,0 +1,141 @@
+module Goodcity
+  class UserUtils
+    def self.merge_user!(master_user_id, other_user_id)
+      master_user = User.find_by(id: master_user_id)
+      other_user = User.find_by(id: other_user_id)
+
+      if master_user && other_user
```
steveyken, line 7: I would wrap this whole part in a transaction. We either want it to work completely, or not at all. We don't want it to fail halfway through and not rollback.

```
User.transaction do 
  <code to reassign user>
end
```

steveyken (approved):
This is really comprehensive and very well thought through. Good set of rspecs too. Thank you!

One comment about using a transaction to ensure either all the assignment changes work or get rolled back, but otherwise, this is good to go.

PR #1256, opened 2021-05-05
GCW-3569 Removed messages from Offer and Item serializer

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3569

### What does this PR do?
- Remove messages from Offer and Item serializer
- Updated shared offer messages serializer response

#### Note:
This change does not hold the backward compatibility, hence Donors who are using old version of Donor app, will not see any messages in the app

Reverting this change: https://github.com/crossroads/api.goodcity/pull/1238


patrixr:
@swatijadhav This is modifying a rather delicate part of the server, once it goes on staging, can we get some proper testing across all apps to make sure it doesn't affect messaging (Tim & Tim can help, you can have a chat with them).

Also please get the confirmation from Matt & Steve that we are okay with a non backwards compatible change (unless it has already been confirmed)

@steveyken  Can we allow a non backward compatible change on the donor app ? Also I'm really hoping we can freeze API v1 soon these kinds of changes on serializers/controllers is dangerous because of all the existing backend spaghetti code.

Review comment: Ideally I would really want to avoid any changes on the serializers/controllers on Api v1, I deem them dangerous at this point. They're too far down the wrong path to fix now, so I don't have any strong suggestion regarding the code apart from making sure it's safe.

I'll approve the PR hopefully @steveyken can have a look before merging

app/controllers/api/v1/messages_controller.rb
```diff
@@ -139,12 +139,18 @@ def notification_serializer
...
+        ActiveModel::ArraySerializer.new(
+          records,
           each_serializer: serializer,
+          include_organisations_users: (staff? && params["include_organisations_users"] == "true"),
           root: "messages"
         ).as_json
       end
 
+      def staff?
+        current_user.has_permission?("can_manage_offers")
```
steveyken, line 151: What about stock app? Can we use `can_manage_offer_messages`, `can_manage_package_messages` and `can_manage_order_messages`

app/controllers/api/v1/offers_controller.rb
```diff
@@ -80,10 +80,7 @@ def search
 
       api :GET, '/v1/offers/1', "List an offer"
       def show
-        render json: offer_serializer.new(@offer, {
-          include_organisations_users: (staff? && params["include_organisations_users"] == "true"),
-          exclude_messages: params["exclude_messages"] == "true"
-        }).as_json
+        render json: @offer, serializer: offer_serializer
```
steveyken, line 83: 💯 so much simpler :)

app/controllers/api/v1/offers_controller.rb
```diff
@@ -184,10 +177,10 @@ def filter_created_by(offers)
       def offer_summary_response(records)
         records = params[:companies]== 'true' ? records : records.with_summary_eager_load
         serializer = params[:companies] == 'true' ? offer_company_serializer : summary_serializer
-        ActiveModel::ArraySerializer.new(records,
+        ActiveModel::ArraySerializer.new(
+          records,
           each_serializer: serializer,
-          root: "offers",
-          include_messages: params[:include_messages] == "true"
```
steveyken, line 190: excited to see this

app/models/item.rb
```diff
@@ -11,9 +11,7 @@ class Item < ApplicationRecord
   belongs_to :rejection_reason
   belongs_to :donor_condition
   has_many   :images, as: :imageable, dependent: :destroy
-  has_many :messages, -> {
-    where(is_private: false) if User.current_user.try(:donor?)
```
steveyken, line 15: so glad this is going away!

app/serializers/api/v1/offer_serializer.rb
```diff
@@ -10,8 +10,6 @@ class OfferSerializer < ApplicationSerializer
       :company_id, :start_receiving_at, :cancellation_reason_id, :cancel_reason
 
     has_many :items, serializer: ItemSerializer
-    # default scope will exclude { is_private = true } messages which is what we want
-    has_many :messages, serializer: MessageSerializer, polymorphic: true
```
steveyken, line 14: excellent!

steveyken (commented):
This is really great so far. The main problem is that it will break existing apps so we can't merge it for a while until the old donor apps are updated. We'll have to hold it here for the time-being but I think the overall direction is excellent

app/controllers/api/v1/messages_controller.rb
swatijadhav: This was specific to shared offer messages shown in admin app and code-change was introduced to load shared-offer message-thread to load charity-details. 
Hence admin-app users with permission `"can_manage_offers"` can see the chairty-details too of messgae-sender.

PR #1275, opened 2021-06-21
GCW-3663-Add offer response table

shreyas098:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3663

### What does this PR do?

FEATURE: Adds offer response table

### DIFF
--- a/app/controllers/api/v1/ability.rb
+++ b/app/controllers/api/v1/ability.rb
@@ -103,10 +103,10 @@ module Api
       def offer_response_abilities
         can :create, OfferResponse do |offer_response|
           offer_response.user_id == @user_id &&
-          Shareable.non_expired.find_by(
-            resource_type: 'Offer',
-            resource_id: offer_response.offer_id
-          ).present?
+            Shareable.non_expired.find_by(
+              resource_type: 'Offer',
+              resource_id: offer_response.offer_id
+            ).present?
         end
 
         can %i[index show], OfferResponse
@@ -217,7 +217,7 @@ module Api
           next false if (
             (message.is_private) || # e.g donor trying to talk in the staff channel
             (message.recipient_id && message.recipient_id != @user_id) || # e.g donor is trying to contact another donor
-            (message.messageable_type== "OfferResponse" && !Shareable.shared_resource?(message.messageable.offer)) #e.g A charity user trying to discuss with the offer that is not shared
+            (message.messageable_type === "OfferResponse" && !Shareable.shared_resource?(message.messageable.offer)) # e.g A charity user trying to discuss with the offer that is not shared
           )
 
           #
--- a/app/controllers/api/v1/ability.rb
+++ b/app/controllers/api/v1/ability.rb
@@ -54,6 +54,7 @@ module Api
         printer_abilities
         offers_package_abilities
         canned_response_abilities
+        offer_response_abilities
         processing_destination_abilities
       end
 
@@ -96,7 +97,11 @@ module Api
       end
 
       def canned_response_abilities
-        can %I[index create update destroy show], CannedResponse if can_manage_canned_response?
+        can %i[index create update destroy show], CannedResponse if can_manage_canned_response?
+      end
+
+      def offer_response_abilities
+        can [:create, :index], OfferResponse
       end
 
       def computer_abilities
@@ -190,12 +195,13 @@ module Api
       def message_abilities
         can %i[index show create notifications], Message, messageable_type: 'Offer' if can_manage_offer_messages?
         can %i[index show create notifications], Message, messageable_type: 'Item' if can_manage_offer_messages?
+        can %i[index show create notifications], Message, messageable_type: 'OfferResponse'
         can %i[index show create notifications], Message, messageable_type: 'Order' if can_manage_order_messages?
         can %i[manage notifications], Message, messageable_type: 'Package' if can_manage_package_messages?
         can %i[mark_read mark_all_read], Message, id: @user.subscriptions.pluck(:message_id)
 
-        can [:show, :index, :notifications], Message, { is_private: false, recipient_id: @user_id, messageable_type: ['Item', 'Offer', 'Order'] }
-        can [:show, :index], Message, { is_private: false, sender_id: @user_id, messageable_type: ['Item', 'Offer', 'Order'] }
+        can [:show, :index, :notifications], Message, { is_private: false, recipient_id: @user_id, messageable_type: ['Item', 'Offer', 'Order', 'OfferResponse'] }
+        can [:show, :index], Message, { is_private: false, sender_id: @user_id, messageable_type: ['Item', 'Offer', 'Order', 'OfferResponse'] }
 
         can :create, Message do |message|
           next false if (
--- a/app/controllers/api/v1/ability.rb
+++ b/app/controllers/api/v1/ability.rb
@@ -109,7 +109,7 @@ module Api
             ).present?
         end
 
-        can %i[index show], OfferResponse
+        can %i[index show], OfferResponse, { user_id: @user_id }
 
         can :manage, OfferResponse if can_manage_offer_response_messages?
       end
--- a/app/controllers/api/v1/ability.rb
+++ b/app/controllers/api/v1/ability.rb
@@ -101,8 +101,17 @@ module Api
       end
 
       def offer_response_abilities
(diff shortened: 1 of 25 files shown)


app/controllers/api/v1/offer_responses_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+        param :offer_response, Hash, required: true do
+          param :user_id, String, desc: "Id of user"
+          param :offer_id, String, desc: "Id of offer"
+        end
+      end
+
+      api :POST, "/v1/offer_responses","Create a offer Response"
+      def create
+        user_id, offer_id = params["offer_response"].values_at('user_id', 'offer_id')
+        return  render json: { errors: "Offer and User Id required"}, status: 422 unless user_id.present? && offer_id.present?
```
steveyken, line 25: Model validation should pick this up already so not required

app/controllers/api/v1/offer_responses_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+
+        @offer_response=OfferResponse.new
+        @offer_response.user=User.find_by_id(user_id)
+        @offer_response.offer=Offer.find_by_id(offer_id)
```
steveyken, line 29: rails active-record should handle the user assignment, so this isn't needed

steveyken (changes requested):
I haven't run through everything but made a few comments... let's chat on Slack about this so far

app/controllers/api/v1/ability.rb
```diff
@@ -96,7 +97,11 @@ def beneficiary_abilities
       end
 
       def canned_response_abilities
-        can %I[index create update destroy show], CannedResponse if can_manage_canned_response?
+        can %i[index create update destroy show], CannedResponse if can_manage_canned_response?
+      end
+
+      def offer_response_abilities
+        can [:create, :index], OfferResponse
```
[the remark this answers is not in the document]
patrixr: @steveyken should we maybe only allow creation on offers that have been shared ? Otherwise I could start inputting random IDs and get access to them

db/migrate/20210620222418_create_offer_responses.rb
```diff
@@ -0,0 +1,13 @@
...
+  def change
+    create_table :offer_responses do |t|
+      t.integer   :user_id
+      t.integer   :offer_id
+
+      t.timestamps
+    end
+
+  add_foreign_key :offer_responses, :users, column: :user_id
+  add_foreign_key :offer_responses, :offers, column: :offer_id
```
steveyken, line 11: would be good to add 2 indexes for offer_id and user_id also

app/controllers/api/v1/ability.rb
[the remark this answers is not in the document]
steveyken: Good call... @shreyas098  will include a block to call `Shared.shareable_resource?` as for `messages_abilities` below

app/models/concerns/message_subscriptions.rb
```diff
@@ -47,7 +47,11 @@ def required_staff_permission(klass)
     if ['Offer', 'Item'].include?(klass)
       return 'can_manage_offer_messages'
     end
-    
+
+    if ['OfferResponse'].include?(klass)
+      return 'can_manage_offer_messages'
+    end
+
```
patrixr, line 54:
```suggestion
    if ['Offer', 'Item', 'OfferResponse'].include?(klass)
      return 'can_manage_offer_messages'
    end
```

app/controllers/api/v1/ability.rb
```diff
@@ -96,7 +97,12 @@ def beneficiary_abilities
       end
 
       def canned_response_abilities
-        can %I[index create update destroy show], CannedResponse if can_manage_canned_response?
+        can %i[index create update destroy show], CannedResponse if can_manage_canned_response?
+      end
+
+      def offer_response_abilities
+        can [:create, :index], OfferResponse, user_id: @user.id
```
patrixr, line 104: This permission rule would still allow me to create an `offer_response` record to an offer that isn't shared

patrixr: (do write a test for that)

patrixr (changes requested):
@shreyas098 Things looking good overall, just added a couple of comments to be looked at. I recommend you have Steve take another look as I'm afraid I might miss some things

app/controllers/api/v1/ability.rb
shreyas098: sure , working on it

app/controllers/api/v1/offer_responses_controller.rb
```diff
@@ -19,27 +19,22 @@ class  OfferResponsesController < Api::V1::ApiController
         end
       end
 
-      api :POST, "/v1/offer_responses","Create a offer Response"
+      api :POST, "/v1/offer_responses", "Create a Offer Response"
       def create
-        user_id, offer_id = params["offer_response"].values_at('user_id', 'offer_id')
-        return  render json: { errors: "Offer and User Id required"}, status: 422 unless user_id.present? && offer_id.present?
+        record = OfferResponse.find_or_create_by(offer_responses_params)
```
steveyken, line 24: `CanCanCan` already gives you `@offer_response` with the params loaded, you don't need to do a `find_or_create_by` here. Just do `if @offer_response.save`

app/models/concerns/message_subscriptions.rb
```diff
@@ -47,7 +47,11 @@ def required_staff_permission(klass)
     if ['Offer', 'Item'].include?(klass)
       return 'can_manage_offer_messages'
     end
-    
+
+    if ['OfferResponse'].include?(klass)
```
steveyken, line 51: Why not make the permission `can_manage_offer_response_messages` and then you don't need this extra `if` statement, you can rely on the convention below to find it

app/controllers/api/v1/ability.rb
steveyken: Yes, `user_id` is correct here but we need to scope `offer_id` also. It should be the list of offers that the user is has created.

```
can [:create, :index], OfferResponse, { user_id: @user_id, offer: { created_by_id: @user.id } }
```

steveyken (commented):
Looking much better but a few more things to tidy up.

app/controllers/api/v1/ability.rb
patrixr: @steveyken I think you're thinking of a different rule here. I think here `offer_id` is scoped to the shared offers

i.e

```ruby
can [:create, :index], OfferResponse, { user_id: @user_id } do |offer_response|
  Shareable.non_expired.find_by(
    resource_type: 'Offer',
    resource_id: offer_response.offer_id
  ).present?
end
```

steveyken:
Ah yes, that's right. Good to scope to shared offers only

app/controllers/api/v1/offer_responses_controller.rb
patrixr: + 1

app/controllers/api/v1/ability.rb
```diff
@@ -96,7 +97,21 @@ def beneficiary_abilities
...
+      def offer_response_abilities
+        can :create, OfferResponse do |offer_response|
+          offer_response.user_id == @user_id &&
+            Shareable.non_expired.find_by(
+              resource_type: 'Offer',
+              resource_id: offer_response.offer_id
+            ).present?
+        end
+
+        can %i[index show], OfferResponse
```
patrixr, line 112:
```suggestion
        can %i[index show], OfferResponse, { user_id: @user_id }
```
A user should only be allowed to view his/her records

patrixr: Note: make sure there's a test for it : )

db/migrate/20210620222418_create_offer_responses.rb
patrixr: @shreyas098 we probably need a unique composite index to avoid duplicates. And we need tests for duplicates generally

patrixr (changes requested):

patrixr (approved):
Looks good @shreyas098 ! Thanks for all the work you've done on this !

swatijadhav (approved):

steveyken:
This looks good now. Thanks @shreyas098

PR #1280, opened 2021-07-05
GCW-3665 Set offer shareable expires_at on offer close

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3665

### What does this PR do?
Set offer shareable expires_at on offer close

### DIFF
--- a/app/models/offer.rb
+++ b/app/models/offer.rb
@@ -192,9 +192,7 @@ class Offer < ApplicationRecord
     end
 
     after_transition on: :cancel do |offer, _transition|
-      if User.current_user == offer.created_by
-        offer.expire_shareable_resource
-      end
+      offer.expire_shareable_resource
     end
 
     after_transition on: :submit do |offer, _transition|
--- a/app/models/offer.rb
+++ b/app/models/offer.rb
@@ -265,13 +265,11 @@ class Offer < ApplicationRecord
     end
   end
 
-  def expire_shareable_resource(expires_at=Time.now)
-    return if self.shareable.blank?
-    expires_at = DateTime.strptime(expires_at, '%d/%m/%y %H:%M %z') if expires_at.is_a?(String)
-    self.shareable.update(expires_at: expires_at)
-
-    if self.packages.present?
-      self.packages.each{ |record| record.shareable.update(expires_at: expires_at) }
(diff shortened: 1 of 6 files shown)


app/models/offer.rb
```diff
@@ -191,6 +191,12 @@ def set_initial_state
       offer.inactive_at = Time.now
     end
 
+    after_transition on: :cancel do |offer, _transition|
+      if User.current_user == offer.created_by
+        offer.expire_shareable_resource
+      end
```
patrixr, line 197: why the user condition 🤔  ?

app/models/offer.rb
```diff
@@ -259,6 +265,16 @@ def offers_count_per_state(offer)
     end
   end
 
+  def expire_shareable_resource(expires_at=Time.now)
+    return if self.shareable.blank?
+    expires_at = DateTime.strptime(expires_at, '%d/%m/%y %H:%M %z') if expires_at.is_a?(String)
```
patrixr, line 270:
```suggestion
    expires_at = Time.parse(expires_at) if expires_at.is_a?(String)
```
(we don't want to assume the format, that's a ticking time bomb)

app/models/offer.rb
```diff
@@ -259,6 +265,16 @@ def offers_count_per_state(offer)
     end
   end
 
+  def expire_shareable_resource(expires_at=Time.now)
```
patrixr, line 268: This is a useful method, but I don't think it should be on the offer. Plus, we'll probably need it for other models in the future.

Can we move it to a static Shareable method ?

e.g

```ruby
Shareable.expire(offer, time)
Shareable.expire(offer.packages, time)
```

patrixr (commented):
Looks okay overall, just dropped a couple of minor comments

app/models/offer.rb
swatijadhav: When Donor cancels offer, we want to auto stop sharing that offer and hence condition added.

patrixr: My point is that I think we should cancel regardless of who cancels the offer

patrixr: Also, it is dangerous to rely on a state (current_user) which only exists in the context of a request.

patrixr (approved):

PR #1282, opened 2021-07-08
Fixed filtering of messages

swatijadhav:
On fetching messages, `messageable_type` should be present along with `messageable_id` values.

### DIFF
--- a/app/controllers/api/v1/messages_controller.rb
+++ b/app/controllers/api/v1/messages_controller.rb
@@ -104,9 +104,9 @@ module Api
       private
 
       def invalid_messageable_resource
-        render json: {
-          errors: "Please provide valid values for messageable_id and messageable_type"
-        }, status: 403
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/messages_controller.rb
```diff
@@ -37,6 +37,8 @@ class MessagesController < Api::V1::ApiController
       param :recipient_id, String, desc: "Specific user id to fetch the messages of"
       param :state, String, desc: "Message state (unread|read) to filter on"
       def index
+        return invalid_messageable_resource if params[:messageable_id].present? && params[:messageable_type].blank?
```
patrixr, line 40:
```suggestion
        raise Goodcity::InvalidParamsError.new('the.translation.key') if params[:messageable_id].present? && params[:messageable_type].blank?
```
To have a consistent error format

patrixr (approved):
Looks good, one tiny comment for the error format

PR #1300, opened 2021-09-24
Fix bug with state_value for current_user

steveyken:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3695

### What does this PR do?

BUG: Fixed bug where state_value was being set to the wrong value

TLDR; When a message is created, the after_create hook fires which sends push_updates to subscribed users. To ensure the correct 'unread/read' state is sent, we adjust the virtual attribute 'state_value' for each user we send to. However, when this is all done, we were not resetting state_value back for the current_user and hence were likely to send the wrong value back to the user who created the message.

This PR fixes the problem but it does assume that the user identified by 'sender_id' (i.e. themself) is creating the message. It wouldn't work if an admin was allowed to create a message on behalf of another user. This currently isn't allowed and would be an unlikely use case. Documented here for completeness.

### Impacted Areas

Dashboard -> Message unread count

### DIFF
--- a/app/controllers/api/v1/messages_controller.rb
+++ b/app/controllers/api/v1/messages_controller.rb
@@ -55,7 +55,13 @@ module Api
       param_group :message
       def create
         @message.sender_id = current_user.id
-        save_and_render_object(@message)
+        if @message.save
+          # after_create hooks may set state_value to the wrong value so reset to 'read' for current_user
+          @message.state_value = 'read'
+          render json: @message, serializer: serializer_for(@message), status: 201
+        else
+          render json: @message.errors, status: 422
+        end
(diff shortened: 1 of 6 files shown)


swatijadhav (approved):

steveyken:
Will ignore code climate issues for now.
Waiting for CI build to finish... possible flaky tests introduced as a result of this PR... checking to see
cc @swatijadhav

steveyken:
@swatijadhav worth re-reviewing... made a few other changes also

swatijadhav (approved):

shreyas098 (approved):

PR #1302, opened 2021-09-27
GCW-3726-Allow chat on expired offers

shreyas098:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3726

### What does this PR do?

FEATURE: Allow chat on shareable offers for users who have previously responded


steveyken (approved):
This works but I'm going to recommend we refactor this `can :create, Message` method as it allows a user to create any kind of message... we can tighten this up a bit with `CanCanCan` parameters

PR #1304, opened 2021-09-28
[Release]v 0.24.1

shreyas098:
Tickets affected : 
GCW-3714
GCW-3703
GCW-3688
GCW-3710
GCW-3711
GCW-3715
GCW-3726


app/controllers/api/v1/ability.rb
```diff
@@ -180,6 +187,7 @@ def image_abilities
             record.imageable.offer.created_by_id == @user_id
           end
           can [:show], Image, { imageable_type: "Package", imageable: { order: { created_by_id: @user_id } } }
+          can :create, Image, { imageable_type: nil }
```
steveyken, line 190: @swatijadhav Possible to change to ?

`can :create, Image, { imageable_type: 'User', imageable_id: @user_id }`

steveyken (approved):

app/controllers/api/v1/ability.rb
swatijadhav: As per our last discussion it is going to be association change and will reflect in all apps on front-end too and will need few more days to work on it.

PR #1337, opened 2026-05-05
Upgrade Ruby to 3.1.7

warp:
Updates version pins across Docker, CircleCI, and deploy scripts, and enables YAML alias loading where Psych now requires it.

## Summary by CodeRabbit

* **Chores**
  * Upgraded Ruby to 3.1.7 across CI, Docker, deploy and runtime; updated Bundler to 2.3.27 and simplified CI test setup (Postgres client install).
  * Regenerated DB schema (timestamp/index normalization).
  * Enabled YAML alias support when loading fixtures/seeds.
  * Set explicit API port and updated Sidekiq runtime/service.

* **Refactor**
  * Improved argument handling across services, token generation, builders and factories; deterministic ordering added to a paginated endpoint; removed an initializer patch.


warp:
@steveyken The only outstanding issue I know of is that state_machines-activerecord ~> 0.9.0 has a high vulnerability. I can't seem to find details, but it appears we would need to upgrade Rails to move to a patched version.

